PR #45, opened 2015-03-07
Add an option to show a diff in the commit

### LINKED ISSUE #38: Diff while commiting
When creating a commit, it's handy to see at a glance what you are actually committing, a diff displayed at the bottom is the optimal solution.

Should also be optional, as it would probably add extra load time on larger commits.

p3lim:
Still needs a update to the syntax, but it's a start.
Resolves #38.

Note; I'm new to Python.

### DIFF
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -34,12 +34,10 @@ class GsCommitCommand(WindowCommand, GitCommand):
 
     def run_async(self, repo_path=None, include_unstaged=False, amend=False):
         repo_path = repo_path or self.repo_path
-        show_diff = sublime.load_settings("GitSavvy.sublime-settings").get("show_commit_diff")
         view = self.window.new_file()
         view.settings().set("git_savvy.get_long_text_view", True)
         view.settings().set("git_savvy.commit_view.include_unstaged", include_unstaged)
         view.settings().set("git_savvy.commit_view.amend", amend)
-        view.settings().set("git_savvy.commit_view.diff", show_diff)
         view.settings().set("git_savvy.repo_path", repo_path)
         view.set_syntax_file("Packages/GitSavvy/syntax/make_commit.tmLanguage")
         view.set_name(COMMIT_TITLE)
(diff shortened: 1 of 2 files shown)


core/commands/commit.py
```diff
@@ -62,6 +65,10 @@ def run(self, edit):
         else:
             initial_text = COMMIT_HELP_TEXT
 
+        if self.view.settings().get("git_savvy.commit_view.diff"):
```
divmain:
```python
if sublime.load_settings("GitSavvy.sublime-settings").get("show_commit_diff"):
```

core/commands/commit.py
```diff
@@ -33,10 +34,12 @@ def run(self, **kwargs):
 
     def run_async(self, repo_path=None, include_unstaged=False, amend=False):
         repo_path = repo_path or self.repo_path
+        show_diff = sublime.load_settings("GitSavvy.sublime-settings").get("show_commit_diff")
```
divmain: Delete.

core/commands/commit.py
```diff
@@ -33,10 +34,12 @@ def run(self, **kwargs):
...
         view = self.window.new_file()
         view.settings().set("git_savvy.get_long_text_view", True)
         view.settings().set("git_savvy.commit_view.include_unstaged", include_unstaged)
         view.settings().set("git_savvy.commit_view.amend", amend)
+        view.settings().set("git_savvy.commit_view.diff", show_diff)
```
divmain: Delete.

divmain:
Thanks @p3lim!  Please make the requested updates, and I'll merge this in.  I plan to follow-up with #50 and #51.

alextegelid:
When I'm writing the commit message I get the "GitSavvy Make Commit" syntax which doesn't really affect anything but the @ character and the comments. Is this intentional or is something wrong? If I choose the "GitSavvy Commit" I get a better syntax highlighting.
screen

(Also, thanks for an awesome plugin!)

p3lim:
@alextegelid see #50.

alextegelid:
I see, thanks! I thought it had already been implemented

Issue #36, opened 2015-03-06
Add setting to allow signed commits instead of regular commits


TheHippo:
Add setting to allow signed and/or signed-off commits instead of "regular" commit.

divmain:
@TheHippo, do envision this would be an additional palette command `git: commit (signed)`, or a global option where the flag would be included when set?

p3lim:
Implementation suggestion; hold <kbd>Shift</kbd> when pressing <kbd>CTRL</kbd>+<kbd>Enter</kbd> when committing to also sign it.

TheHippo:
There are two options for signing commits:

* `git commit -s` or `git commit --signoff` which adds just a text snippet to the commit message and seems to be implemented by @p3lim in #46 .
* `git commit -S` or `git commit --gpg-sign` which signs the commit with your GPG key.

Both options can be combined (`git commit -Ss`). In my opinion it would be the best option to have a "normal" commit command (`git: commit`) and a configurable signed commit command (`git: commit(signed)`) where your have to choose in a config file which signing options should be used.)

The GPG sign could be more complicated, since it might require unlocking your GPG agent / key ring.

Other then turning them on and off no other options are required, since GPG key and  message could be configured via the `git` command.

.

p3lim:
> In my opinion it would be the best option to have a "normal" commit command (git: commit) and a configurable signed commit command (git: commit(signed)) where your have to choose in a config file which signing options should be used.)

What about having a separate commit command (`git: commit (signed)`) for GPG-signed, and also have my implementation in #46, so the user can mix and match between them (as you could with the porcelain).

TheHippo:
@p3lim This sounds good.

divmain:
Closing this also, since #68 seems to be blocked indefinitely.

rpbaptist:
I looked into this and this can actually be solved outside of GitSavvy by simply configuring git globally or in your project. 

`git config (--global) commit.gpgsign true`

If you have a use case where you sometimes do and sometimes do not wish to sign your commits in the same project this won't help, but that's not something I am running into.

For more details: https://stackoverflow.com/questions/10161198/is-there-a-way-to-autosign-commits-in-git-with-a-gpg-key

TheHippo:
@rpbaptist: This still doesn't solve the issue of unlocking you GPG key.

rpbaptist:
[...]

Issue #44, opened 2015-03-07
Output to panels


p3lim:
Most of the outputs in this plugin is done to the status bar.
I would like to see an option to output to panels, especially for certain actions like commit or push where you'd want to see the output, like a response from the server(s).

jbrooksuk:
:+1: for showing the push output in a panel.

This is especially useful when pushing to Dokku or Heroku for example, where the buildstep progress is long.

divmain:
In `GitSavvy.sublime-settings`:
```json
  "show_panel": ["push"]
```

p3lim:
As far as I understood it, this is how you'd create and output to a panel in ST3:
```Python
window = sublime.active_window()
output = window.get_output_panel("name_of_panel")
window.run_command("show_panel", {"panel": "output.name_of_panel"})
output.set_read_only(False)
output.run_command("append", {"characters": "Hello world!"})
output.set_read_only(True)
```
Documentation is not that good, so this might save you some time.

divmain:
Thanks @p3lim.  I'll pick this up early this upcoming week.  GitSavvy displays panel output for failed Git commands, so additional conditions is a simple add.  I'll make this opt-in, where you can configure panel output using `show_panel` as seen in my above comment.

PR #46, opened 2015-03-07
Add a keybinding to the commit view to easily sign off commits

p3lim:
Partial fix for #36, this will sign-off commits by hitting <kbd>Super</kbd>+<kbd>s</kbd>.

As with my previous PR, this will need proof-checking, I'm still quite new to Python.

### DIFF
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -17,6 +17,11 @@ COMMIT_HELP_TEXT = """
 ## key to reference an issue in a different GitHub repo.
 """.format(key="CTRL" if os.name == "nt" else "SUPER")
 
+COMMIT_SIGN_TEXT = """
+
+Signed-off-by: {name} <{email}>
+"""
+
 COMMIT_TITLE = "COMMIT"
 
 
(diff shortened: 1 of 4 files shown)


TheHippo:
Why the manual sign off message generation. Couldn't this be done by `git` itself?

p3lim:
I chose this method instead of amending -s (or --signoff) to the commit command (GsCommitViewDoCommitCommand) was to give more of a visual representation to the user that it actually did sign-off.

I suppose my original implementation suggestion from #36 could work fine, but that would require a more verbose output (like suggested in #44) so that the user could verify it.

core/commands/commit.py
```diff
@@ -92,3 +97,26 @@ def run_async(self):
...
+
+        config_name = self.git("config", "user.name").strip()
+        config_email = self.git("config", "user.email").strip()
+
+        sign_text = COMMIT_SIGN_TEXT.format(name=config_name, email=config_email)
+        view_text_list[0] += sign_text
+        view_text_list.insert(1, COMMIT_HELP_TEXT)
+
+        self.view.run_command("gs_replace_view_text", {
+            "text": "".join(view_text_list),
```
divmain: This should read:
```python
"text": COMMIT_HELP_TEXT.join(view_text_list)
```

core/commands/commit.py
```diff
@@ -92,3 +97,26 @@ def run_async(self):
...
+    def run(self, edit):
+        view_text = self.view.substr(sublime.Region(0, self.view.size()))
+        view_text_list = view_text.split(COMMIT_HELP_TEXT)
+
+        config_name = self.git("config", "user.name").strip()
+        config_email = self.git("config", "user.email").strip()
+
+        sign_text = COMMIT_SIGN_TEXT.format(name=config_name, email=config_email)
+        view_text_list[0] += sign_text
+        view_text_list.insert(1, COMMIT_HELP_TEXT)
```
divmain: You can remove this line.

divmain:
Looks good to me.  Can you also update the `COMMIT_HELP_TEXT` to indicate the availability of the shortcut?

PR #47, opened 2015-03-08
Conditional check if remote url ends with .git

iamolivinius:
github integration doesn't work with remote urls that to not end on ".git".

users like me who tend to clone repos by just copy and pasting the address line out of the browser into the terminal would be affected by this issue.


divmain:
Seems reasonable.  Thanks @iamolivinius!

PR #62, opened 2015-03-09
Avoid empty sublime.Region for removed newline. Closes #61.

### LINKED ISSUE #61: inline-diff newline change ambiguity
This code

    Line 1 content
    // comment
    Line 3 content

and this code

    Line 1 content
    
    // comment
    Line 3 content

generate the same inline diff:

    Line 1 content
    
    // comment  <--- (green text)
    Line 3 content

There's no indication of whether I've added a newline or not.

If I simply add or remove a newline, the diff acts as expected. It's only when I add text to a previously blank line that this ambiguity shows up.

Obviously this isn't a huge deal, as I know where I've added newlines and where I haven't, and these comments aren't generally things I'll be committing; it's just a little bit unexpected behavior.

ethanbustad:
Check out this fix when you have a chance. Not sure what necessitated the region-end as you had it before, but this seems to work for several different types of diffs I tested....

### DIFF
--- a/core/commands/inline_diff.py
+++ b/core/commands/inline_diff.py
@@ -264,8 +264,7 @@ class GsInlineDiffRefreshCommand(TextCommand, GitCommand):
                     region_type = line_type
                     region_start = line.begin()
                 elif region_type != line_type:
-                    # End region with final character of previous line.
-                    region_end = line.begin() - 1
+                    region_end = line.begin()
                     l = add_regions if region_type == "+" else remove_regions
                     l.append(sublime.Region(region_start, region_end))
 


divmain:
Very interesting!  Would you mind posting a screenshot of the before/after - I had trouble visualizing the differences in highlights between the two scenarios you referenced in the issue.  Thanks @ethanbustad!

ethanbustad:
Alright, haha, I think I'm confident of this fix now.
I can't find it in any documentation, but I think sublime.Region acts like list slice: the first index inclusive, the second exclusive. So by subtracting one from the hunk, it was leaving out one character.

The case I indicated in the issue: https://dl.dropboxusercontent.com/u/101485540/screenshots/newline-before-and-after.png
It can also be seen in other add+remove situations, where the newline at the end of the removed line was missing: https://dl.dropboxusercontent.com/u/101485540/screenshots/add-remove-before-and-after.png

PR #68, opened 2015-03-10
Signed commits

TheHippo:
Started to implement GPG signed commits. This is work in progress.

- [x] command for signed commits
- [x] command for signed amends
- [ ] test what happens if GPG key isn't unlocked
- [ ] test what happens if no GPG key is configured
- [x] allow signed commit from status view.
- [ ] figure out how to work around the missing `tty` problem.

If there is anything you don't like about my code, please let me know. This is the first time I'm working / hacking on a Sublime Text plugin.

### DIFF
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -9,7 +9,10 @@ from ..git_command import GitCommand
 COMMIT_HELP_TEXT = """
 
 ## To make a commit, type your commit message and press {key}-ENTER. To cancel
-## the commit, close the window. To sign off the commit press {key}-S.
+## the commit, close the window. Press {key}-SHIFT-ENTER to create a GPG signed
+## commit.
+
+## To sign off the commit message press {key}-S.
 
 ## You may also reference or close a GitHub issue with this commit.  To do so,
 ## type `#` followed by the `tab` key.  You will be shown a list of issues
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -106,11 +106,7 @@ class GsCommitViewDoCommitCommand(TextCommand, GitCommand):
 
         commit_args.extend(["-F", "-"])
 
-        commit_kwargs = {
-            "stdin": commit_message
-        }
-
-        self.git(*commit_args, **commit_kwargs)
+        self.git(*commit_args, stdin=commit_message)
 
         self.view.window().focus_view(self.view)
         self.view.window().run_command("close_file")
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -37,13 +37,14 @@ class GsCommitCommand(WindowCommand, GitCommand):
     def run(self, **kwargs):
         sublime.set_timeout_async(lambda: self.run_async(**kwargs), 0)
 
-    def run_async(self, repo_path=None, include_unstaged=False, amend=False):
+    def run_async(self, repo_path=None, include_unstaged=False, amend=False, signed=False):
         repo_path = repo_path or self.repo_path
         view = self.window.new_file()
         view.settings().set("git_savvy.get_long_text_view", True)
         view.settings().set("git_savvy.commit_view.include_unstaged", include_unstaged)
         view.settings().set("git_savvy.commit_view.amend", amend)
         view.settings().set("git_savvy.repo_path", repo_path)
+        view.settings().set("git_savvy.signed", signed)
         view.set_syntax_file("Packages/GitSavvy/syntax/make_commit.tmLanguage")
         view.set_name(COMMIT_TITLE)
         view.set_scratch(True)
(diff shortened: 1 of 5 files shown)


TheHippo:
This should fix #36.

core/commands/commit.py
```diff
@@ -95,10 +96,21 @@ def run_async(self):
...
+        if self.view.settings().get("git_savvy.signed"):
+            commit_args.append("--gpg-sign")
+
+        commit_args.extend(["-F", "-"])
+
+        commit_kwargs = {
+            "stdin": commit_message
+        }
+
+        self.git(*commit_args, **commit_kwargs)
```
divmain: Since you're not generating the keyword args dynamically, please change this line to:
```python
self.git(*commit_args, stdin=commit_message)
```

divmain:
@TheHippo, let me know when this is ready to go.  Do you have a workaround in mind for when the private key is password-locked?

divmain:
I have some thoughts for how to handle passphrase-protected private keys.  If you don't make headway with this, let me know and I'll pick up this branch and move forward with what I have in mind.

TheHippo:
Actually I'd just display a message when the key is locked. People who lock their keys although git commits probably happen very frequently are maybe also too cautious to enter them into a Sublime Plugin? What is your idea?

I also used GitSavvy for some work this afternoon and find the current situation less helpful than I thought. (I do switch a lot between repositories.) Having an additional command for signed / unsigned commits / amends on all places where they are displayed creates some UI clutter and switching back end forth is annoying because the command palette just "learned" the other command for when you type "comm" into the search input field. Since most people don't care about GPG signing their commits, so it might be the easiest solution to enable commit signing via the config file (globally and maybe per Sublime project file?) and use signing every time a commit is triggered. (And in addition maybe check if `git config user.signingkey` is defined.)

```javascript
{
    /* GPG sign commits, default is false */
    "sign_commits": true
}

**Comment by divmain** [comment]
How about this alternative: you get to the view where you enter a commit message via the usual means.  Then, instead of `Super+Enter` you press `Super+Shift+Enter` to do a GPG-signed commit.  That way there's no switching back and forth in the command palette, and UI/UX remains (largely) unchanged.  Thoughts?

**Comment by TheHippo** [comment]
This sound good too:

* <kbd>SUPER</kbd> + <kbd>ENTER</kbd> = commit
* <kbd>SUPER</kbd> + <kbd>S</kbd> = commit with signed off message
* <kbd>SUPER</kbd> + <kbd>SHIFT</kbd> + <kbd>ENTER</kbd> = signed commit
* <kbd>SUPER</kbd> + <kbd>SHIFT</kbd> + <kbd>S</kbd> = signed commit with signed off commit message

This would allow all combinations from within one view without any additional UI clutter. And keep required user input consistent.

I'd implement these changes and update my pull request.

**Comment by divmain** [comment]
@TheHippo, let's do:
- <kbd>SUPER</kbd> + <kbd>S</kbd> = add signed-off text to commit message (already implemented)
- <kbd>SUPER</kbd> + <kbd>ENTER</kbd> = commit (already implemented)
- <kbd>SUPER</kbd> + <kbd>SHIFT</kbd> + <kbd>ENTER</kbd> = signed commit

There's nothing stopping the user from doing <kbd>SUPER</kbd> + <kbd>S</kbd> and then <kbd>SUPER</kbd> + <kbd>SHIFT</kbd> + <kbd>ENTER</kbd> to first add the "Signed-off" message and then sign the commit with GPG.

**Comment by TheHippo** [comment]
Never tested signed off message because it conflicts with a Gnome shortcut. I thought it'd sign off the commit and perform the actual commit instantly, but it just add the message to the window. My fault, sorry.

About the locked keys: The env variable should be named `GPG_PASSPHRASE_FD` and point to a file descriptor which holds the password for the key. This should works, because afaik git does use *gpg* and not *libgpg*. As a bash script:

```bash
GPG_PASSPHRASE_FD=<( printf -- "secretpassword" ) git commit ...
```

BTW: I just learned since git 2.0 it is possible to configure git to [auto sign all commits](https://git.kernel.org/cgit/git/git.git/tree/Documentation/RelNotes/2.0.0.txt#n138) on its own.

TheHippo:
Implemented the signed commit command. This works as expected as long as you start sublime from your terminal, otherwise signed commit raise an error like this:

```
/usr/bin/git commit -q --amend --gpg-sign -F -` failed with following output:

gpg: cannot open tty `/dev/tty': No such device or address
error: gpg failed to sign the data
```

GPG does have an option that forces ignoring the missing TTY, but this option is not exposed through the git command line flags.

Slowly it get more and more clearer to me, why (nearly) no none has support for signed commits in their IDE / editor / etc.

divmain:
@TheHippo, it sounds like we had the same idea regarding getting the password to GPG.  I will open a separate issue with my thoughts for handling locked private keys.

core/commands/commit.py
```diff
@@ -95,15 +99,38 @@ def run_async(self):
...
-        else:
-            self.git("commit", "-q", "-F", "-", stdin=commit_message)
+            commit_args.append("--amend")
+
+        if self.signed_commit:
+            commit_args.append("--gpg-sign")
+
+        commit_args.extend(["-F", "-"])
+
+        self.git(*commit_args, stdin=commit_message)
```
divmain: I've used the following pattern elsewhere (the `git` method automatically strips falsey args).  Please use it here.

```python
self.git(
    "commit",
    "-q",
    "--amend" if self.view.settings().get("git_savvy.commit_view.amend") else None,
    "--gpg-sign" if self.signed_commit else None,
    "-F"
    "-"
    stdin=commit_message
    )
```

divmain:
Hey @TheHippo, is there additional work to do for this?  Two of the items in the checklist are not marked, which is why I haven't merged.

TheHippo:
In addition to the two open items on the checklist I also need to figure out the 

```
gpg: cannot open tty `/dev/tty': No such device or address
error: gpg failed to sign the data
```

error which occurs if Sublime isn't started from a terminal.

divmain:
This PR is pretty stale at this point.  Feel free to re-open if and when you have time.

TheHippo:
The whole `tty` problem is way to complicated to solve (you cannot force users to open sublime from console) just for making signed commits. The will probably get a lot more easier if `gpg` version 2 is widely adopted.

PR #72, opened 2015-03-11
Add an option to suppress the input when outputting to a panel

p3lim:


p3lim:
Outputting the input could be considered a dev option and could be tied to the "dev_mode" option instead.

PR #76, opened 2015-03-11
Don't lock up the panels/inputs while pushing

p3lim:
I moved `do_push` outside of the classes (since they all use something like this) and turned all calls of it (and `self.push`) into calling it asynchronously.

I was thinking of moving it even further away (maybe into the mixins) since other files might make use of it as well. And while on that train of thought I was also thinking of maybe moving every occurrence of the following block of code just the same.

```Python
self.remotes = list(self.get_remotes().keys())

if not self.remotes:
    self.window.show_quick_panel([NO_REMOTES_MESSAGE], None)
else:
    self.window.show_quick_panel(
        self.remotes,
        self.on_select_remote,
        flags=sublime.MONOSPACE_FONT
        )
```

### DIFF
--- a/core/commands/push.py
+++ b/core/commands/push.py
@@ -10,26 +10,27 @@ START_PUSH_MESSAGE = "Starting push..."
 END_PUSH_MESSAGE = "Push complete."
 PUSH_TO_BRANCH_NAME_PROMPT = "Enter remote branch name:"
 
-def do_push(self, remote, branch):
-    """
-    Perform `git push remote branch`.
-    """
-    sublime.status_message(START_PUSH_MESSAGE)
-    self.push(remote, branch)
-    sublime.status_message(END_PUSH_MESSAGE)
-    util.view.refresh_gitsavvy(self.window.active_view())
+class PushBase(GitCommand):
+    def do_push(self, remote, branch):
+        """
+        Perform `git push remote branch`.
+        """
+        sublime.status_message(START_PUSH_MESSAGE)
+        self.push(remote, branch)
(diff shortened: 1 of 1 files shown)


divmain:
This is a nice refactor.  I did some work to support status bar mini-status earlier today, which is why you have the merge conflict here.  Should be straight-forward to resolve.  Ping me when this PR is ready to merge.

p3lim:
Do you want me to also do the other things I mentioned above? (Moving `do_push` and the remote quickpanel into mixins.)

divmain:
Now that I think about it, because of the passing of `self` to a straight-up function, this might be cleaner as a base-class here in the module.

```python
class PushBase(GitCommand):

    def do_push(self, remote, branch):
        """
        Perform `git push remote branch`.
        """
        sublime.status_message(START_PUSH_MESSAGE)
        self.push(remote, branch)
        sublime.status_message(END_PUSH_MESSAGE)
        if self.window.active_view().settings().get("git_savvy.status_view"):
            self.window.active_view().run_command("gs_status_refresh")

# ...

class GsPushCommand(WindowCommand, PushBase):

# ...

        sublime.set_timeout_async(lambda: self.do_push(None, None))

class GsPushToBranchCommand(WindowCommand, PushBase):

# ...

class GsPushToBranchNameCommand(WindowCommand, PushBase):

# ...
```

p3lim:
Your code, your call :smile:

p3lim:
@divmain It's good to go.

divmain:
Excellent.

Issue #75, opened 2015-03-11
Handle all valid git branch names?


ethanbustad:
Related to https://github.com/divmain/GitSavvy/issues/52 (and https://github.com/divmain/GitSavvy/issues/60).

My current requirement is that branch names with a period display correctly. Currently branch `ee-6.2.x` is showing up as `ee-6`.

However, given the previous issues which described a similar problem, it seems worthwhile to prevent an issue a week for handling another valid branch name, until all valid branch names are supported.

PR #77, opened 2015-03-11
Sacrifice specificity for sensitivity in status view branch pattern

ethanbustad:
The idea here is not to validate a branch's name, but to parse it from the
git status output. Git will not allow a branch to be created with an
invalid name. Thus what I've tried to do is to create a simple regex that
will recognize any valid branch, while not worrying about each branch
naming rule at http://git-scm.com/docs/git-check-ref-format#_description.
Closes https://github.com/divmain/GitSavvy/issues/75.

### DIFF
--- a/core/git_mixins/active_branch.py
+++ b/core/git_mixins/active_branch.py
@@ -1,4 +1,5 @@
 import re
+import string
 
 
 class ActiveBranchMixin():
@@ -24,7 +25,7 @@ class ActiveBranchMixin():
           3) remote branch name
           4) boolean indicating whether branch is clean
           5) # commits ahead of remote
-          6) # commits behind of remote 
+          6) # commits behind of remote
         """
         stdout = self.git("status", "-b", "--porcelain").strip()
 
@@ -38,7 +39,9 @@ class ActiveBranchMixin():
         if first_line.startswith("## Initial commit on "):
             return False, True, first_line[21:], clean, None, None, None
 
-        short_status_pattern = r"## ([A-Za-z0-9\-_\/]+)(\.\.\.([A-Za-z0-9\-_\/]+)( \[((ahead (\d+))(, )?)?(behind (\d+))?\])?)?"
+        valid_punctuation = "".join(c for c in string.punctuation if c not in "~^:?*[\\")
+        branch_pattern = "[A-Za-z0-9" + re.escape(valid_punctuation) + "]+?"
+        short_status_pattern = r"## (" + branch_pattern + ")(\.\.\.(" + branch_pattern + ")( \[((ahead (\d+))(, )?)?(behind (\d+))?\])?)?$"
         status_match = re.match(short_status_pattern, first_line)
 
         if not status_match:


core/git_mixins/active_branch.py
```diff
@@ -34,7 +35,8 @@ def get_branch_status(self):
         if first_line.startswith("## Initial commit on "):
             return "Initial commit on `{}`.".format(first_line[21:])
 
-        short_status_pattern = r"## ([A-Za-z0-9\-_\/]+)(\.\.\.([A-Za-z0-9\-_\/]+)( \[((ahead (\d+))(, )?)?(behind (\d+))?\])?)?"
+        branch_pattern = "[A-Za-z0-9" + re.escape(re.sub("[~\\^\\:\\?\\*\\[\"]", "", string.punctuation)) + "]+?"
```
divmain: Why the `?` at the end?  Does this work without?

ethanbustad: Without the `?`, it greedily gobbles up remotes, too. So in the case of

    ## master...origin/master

With the `?` it yields `master` as the branch name, with `origin/master` as the remote name.
Without the `?` it yields `master...origin/master` as the branch name, no remote recognized.

core/git_mixins/active_branch.py
```diff
@@ -34,7 +35,8 @@ def get_branch_status(self):
...
+        short_status_pattern = r"## (" + branch_pattern + ")(\.\.\.(" + branch_pattern + ")( \[((ahead (\d+))(, )?)?(behind (\d+))?\])?)?$"
```
ethanbustad: Because of the reluctant quantifier in the branch pattern, we need to add the `$` at the end to make sure the whole line is matched. In the case of

    ## branch_name

With the `$` it yields `branch_name`.
Without the `$` it yields `b`. Very reluctant.

divmain:
Alright, I'm convinced :)  I mean, holy moly, that's quite a regex.  But you're right, original version is not really correct.  I just did some refactoring in this file - can you pull down latest master and put this into its new location (should be as simple as that).

While we're at it, the `branch_pattern` could be further broken up like so:
```python
valid_punctuation = "".join(c for c in string.punctuation if not c in ("~", ...))
branch_pattern = "[A-Za-z0-9" + valid_punctuation + "]+"
short_status_pattern = "## (" + branch_pattern + ")(\.\.\.(" + branch_pattern + ")( \[((ahead (\d+))(, )?)?(behind (\d+))?\])?)?$"
```

The `valid_punctuation` line has `...` in it because I had trouble seeing straight with all the escaped characters.

ethanbustad:
Haha ok. Sorry about the crazy escapes. That was the clearest I was able to write it so far -- I tried to do it with a `r"raw string"` but I couldn't get it to work.
I'll pull down latest, try to clean this up a little more if I can, test it out, and update this PR.

ethanbustad:
Just updated the PR. Let me know what you think!

divmain:
Thanks!

PR #78, opened 2015-03-11
Adding head SHA1 and abbreviated commit msg to status view

jlangston:
Switching to GitSavvy from SublimeGit one small feature I and possibly others might find useful to is having a quick reference to the HEAD current commit SHA1 and commit msg as part of the status view. I find this handy for info about the most recent commit as well as when I am cherry picking the latest commit I can easily grab the SHA1 I'll be cherry picking before switching branches to do so.

### DIFF
--- a/core/commands/status.py
+++ b/core/commands/status.py
@@ -42,6 +42,7 @@ STASHES_TEMPLATE = """
 STATUS_HEADER_TEMPLATE = """
   BRANCH:  {branch_status}
   ROOT:    {repo_root}
+  HEAD:    {current_head}
 """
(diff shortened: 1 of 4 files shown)


divmain:
Seems reasonable.

jlangston:
Thanks for merging this. 

Great work on this plugin! It's really well done.

divmain:
@jlangston :+1, and thanks for the PR!  I made a small modification after the fact to how it is styled.

screen shot 2015-03-11 at 2 55 07 pm

So meta!

Issue #10, opened 2015-02-12
`git: tags` view


divmain:
- [ ] show all local tags
- [ ] add tag
- [ ] delete tag
- [ ] show tags on remote (?)
- [ ] push tags to remote

p3lim:
I was thinking about tackling this one today, as I use tags quite a bit in my projects.

I wanted to hear with your first though, do you want it to be handled by the menus entirely, or do you want something more like the status window?
Both have their advantages and disadvantages.

divmain:
I was thinking something similar to the status view.  I like the way I can interact with stashes in the command view, and I imagine it'd be a similar productivity boost here.  Basically `git: tag` palette command takes you to the TAGS view.  Does that sound reasonable?

p3lim:
I agree on using a view for this, but I'm still going to implement some quick commands for the menu as well (just like with commit there is quick commit).

I don't mind the work that is required by this, I'm really enjoying Python so far, fantastic language.

divmain:
:+1:

divmain:
Closed by #80.

PR #80, opened 2015-03-12
Feature: Tags view

p3lim:
As discussed in #10, this is the branch/PR I will implement the tags view feature.

Features:
- [x] create view
- [x] delete tag
- [x] create tag
- [x] push tag
- [x] push all tags
- [x] list remote tags
- [x] delete remote tag
- [x] view commit diff linked to tag

Todo:
- [x] validate tag names
- [x] make more actions async so the view/window doesn't hang
- [ ] cursor is inconsistent
- [ ] document

### DIFF
--- a/core/commands/tags.py
+++ b/core/commands/tags.py
@@ -289,7 +289,7 @@ class GsTagPushCommand(TextCommand, GitCommand):
         elif hasattr(self, "items") and self.items:
             refs = ""
             for item in self.items:
-                refs += "refs/tags/" + item[1] + ":"
+                refs += "refs/tags/" + item[1] + " "
             refs = refs[:-1]
 
             self.git("push", selected_remote, refs)
--- a/core/commands/tags.py
+++ b/core/commands/tags.py
@@ -422,4 +422,14 @@ class GsTagViewLogCommand(TextCommand, GitCommand):
         items = tuple(line[4:].strip().split() for line in lines if line)
 
         if items:
-            self.git("log", "-1", "--pretty=medium", items[0][0], show_panel=True)
+            stdout = self.git("show", items[0][0])
+
+            panel_view = self.view.window().create_output_panel("GitSavvy")
+            panel_view.set_syntax_file("Packages/GitSavvy/syntax/show_commit.tmLanguage")
+            panel_view.settings().set("line_numbers", False)
+            panel_view.set_read_only(False)
+            panel_view.erase(edit, sublime.Region(0, panel_view.size()))
+            panel_view.insert(edit, 0, stdout)
+            panel_view.set_read_only(True)
+            panel_view.show(0)
+            self.view.window().run_command("show_panel", {"panel": "output.{}".format("GitSavvy")})
--- a/core/commands/tags.py
+++ b/core/commands/tags.py
@@ -263,9 +263,17 @@ class GsTagCreateCommand(WindowCommand, GitCommand):
         if not tag_name:
             return
 
-        # TODO: do some validation
+        stdout = self.git(
+            "check-ref-format",
+            "--normalize",
+            "refs/tags/" + tag_name,
+            throw_on_stderr=False
+            )
 
-        self.tag_name = tag_name
+        if not stdout:
+            return util.log.panel("\"{}\" is not a valid tag name.".format(tag_name))
+
+        self.tag_name = stdout.strip()[10:]
         self.window.show_input_panel(
             TAG_CREATE_MESSAGE_PROMPT,
             sublime.load_settings("GitSavvy.sublime-settings").get("default_tag_message"),
--- a/core/commands/tags.py
+++ b/core/commands/tags.py
@@ -235,7 +235,7 @@ class GsTagCreateCommand(WindowCommand, GitCommand):
         tag message. If the message is empty, use the pre-defined one.
         """
         # If the user pressed `esc` or otherwise cancelled.
-        if not tag:
+        if not tag_name:
             return
 
         # TODO: do some validation
@@ -261,7 +261,7 @@ class GsTagCreateCommand(WindowCommand, GitCommand):
             default_message = sublime.load_settings("GitSavvy.sublime-settings").get("default_tag_message")
             message = default_message.format(tag_name=self.tag_name)
 
-        self.git("tag", self.tag, "-F", "-", stdin=message)
+        self.git("tag", self.tag_name, "-F", "-", stdin=message)
 
 
 class GsTagPushCommand(TextCommand, GitCommand):
--- a/core/commands/tags.py
+++ b/core/commands/tags.py
@@ -6,6 +6,8 @@ from sublime_plugin import WindowCommand, TextCommand, EventListener
 from ..git_command import GitCommand
 from ...common import util
 
+TAG_DELETE_MESSAGE = "Tag deleted."
+
 VIEW_TITLE = "TAGS: {}"
 
 LOCAL_TEMPLATE = """
@@ -38,8 +40,7 @@ KEY_BINDINGS_MENU = """
   #############
 
   [c] create (NYI)
-  [d] delete (NYI)
-  [D] delete locally and remotely (NYI)
+  [d] delete
   [p] push to remote(s) (NYI)
   [P] push all tags to remote(s) (NYI)
   [l] view commit
@@ -177,6 +178,30 @@ class GsTagsFocusEventListener(EventListener):
     def on_activated(self, view):
         if view.settings().get("git_savvy.tags_view") == True:
             view.run_command("gs_tags_refresh")
+
+
+class GsTagDeleteCommand(TextCommand, GitCommand):
+
+    """
+    Delete tag(s) in selection.
+    """
+
+    def run(self, edit):
+        valid_ranges = view_section_ranges[self.view.id()][:3]
+
+        lines = util.view.get_lines_from_regions(
+            self.view,
+            self.view.sel(),
+            valid_ranges=valid_ranges
+            )
+
+        items = tuple(line[4:].strip().split() for line in lines if line)
+
+        if items:
+            for item in items:
+                self.git("tag", "-d", item[1])
+            util.view.refresh_gitsavvy(self.view)
+            sublime.status_message(TAG_DELETE_MESSAGE)
 
 
 class GsTagViewLogCommand(TextCommand, GitCommand):
--- a/core/commands/tags.py
+++ b/core/commands/tags.py
@@ -13,6 +13,11 @@ LOCAL_TEMPLATE = """
 {}
 """
 
+REMOTE_TEMPLATE = """
+  REMOTE ({}):
+{}
+"""
+
 VIEW_HEADER_TEMPLATE = """
   BRANCH:  {branch_status}
   ROOT:    {repo_root}
@@ -21,6 +26,10 @@ VIEW_HEADER_TEMPLATE = """
 
 NO_TAGS_MESSAGE = """
   Your repository has no tags.
+"""
+
+LOADING_TAGS_MESSAGE = """
+  Please stand by while fetching tags from remote(s).
 """
 
 KEY_BINDINGS_MENU = """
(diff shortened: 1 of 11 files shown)


p3lim:
What it looks like so far:

image

p3lim:
This is pretty much feature-complete, I'll be adding some validation to the creation of tags and rework some things with the listing of remote tags, but please come with suggestions or comments.

core/commands/tags.py
```diff
@@ -0,0 +1,318 @@
...
+
+class GsTagsRefreshCommand(TextCommand, GitCommand):
+
+    """
+    Get the current state of the git repo and display tags and command
+    menu to the user.
+    """
+
+    def run(self, edit, **kwargs):
+        sublime.set_timeout_async(lambda: self.run_async(**kwargs))
```
divmain: Is the `**kwargs` here left over from previous work?

core/commands/tags.py
```diff
@@ -0,0 +1,318 @@
...
+            nonlocal cursor
+            start = cursor
+            cursor += len(new_text)
+            end = cursor
+            return sublime.Region(start, end)
+
+        lines = ""
+        if tags:
+            lines = "\n".join(
+                "    {} {}".format(t.sha[:7], t.tag)
```
divmain: Can the `tag` property here be `name` instead?

core/commands/tags.py
```diff
@@ -0,0 +1,318 @@
...
+                for t in tags
+                )
```
divmain: This can be one line.

core/commands/tags.py
```diff
@@ -0,0 +1,318 @@
...
+        else:
+            lines = NO_LOCAL_TAGS_MESSAGE
+
+        view_text = LOCAL_TEMPLATE.format(lines)
+        regions.append(get_region(view_text))
+
+        self.remotes = list(self.get_remotes().keys())
+        if self.remotes:
+            for remote in self.remotes:
+                remote_text = REMOTE_TEMPLATE.format(remote, NO_REMOTE_TAGS_MESSAGE)
```
divmain: Is this in progress?

core/commands/tags.py
```diff
@@ -0,0 +1,318 @@
...
+            None,
+            None
+            )
+
+    def on_entered_tag(self, tag_name):
+        """
+        After the user has entered a tag name, prompt the user for a
+        tag message. If the message is empty, use the pre-defined one.
+        """
+        # If the user pressed `esc` or otherwise cancelled.
```
divmain: `on_entered_tag` will not be called if the user presses `esc`.  The below condition will only ever be true if the user enters an empty string.

core/commands/tags.py
p3lim: That is pretty much copied from the status view.
https://github.com/divmain/GitSavvy/blob/7713415730f8daa5907704695415b00e66bb74c1/core/commands/status.py#L118-L119

core/commands/tags.py
```diff
@@ -0,0 +1,318 @@
...
+        if not tag_name:
+            return
+
+        # TODO: do some validation
```
divmain: You may already have something in mind for this.  I think there's a regex floating around to validate branch/tag names.  `self.git("check-ref-format", "--normalize", tag_name).strip() == tag_name` would also work.

core/commands/tags.py
p3lim: Yes.

core/commands/tags.py
```diff
@@ -0,0 +1,318 @@
...
+            None
+            )
+
+    def on_entered_message(self, message):
+        """
+        Perform `git tag tag_name -F -`
+        """
+        # If the user pressed `esc` or otherwise cancelled
+        if message == -1:
+            return
```
divmain: Same comment here.  This would be a valid check in a callback for `show_quick_panel`.

core/commands/tags.py
```diff
@@ -0,0 +1,318 @@
...
+
+        if not message:
+            default_message = sublime.load_settings("GitSavvy.sublime-settings").get("default_tag_message")
+            message = default_message.format(tag_name=self.tag_name)
```
divmain: This was a good way to go about it :+1:

core/commands/tags.py
p3lim: I suppose that behavior only applies to `show_quick_panel`.

core/commands/tags.py
```diff
@@ -0,0 +1,318 @@
...
+            # Valid sections are in the Local section
+            valid_ranges = view_section_ranges[self.view.id()][:1]
+
+            lines = util.view.get_lines_from_regions(
+                self.view,
+                self.view.sel(),
+                valid_ranges=valid_ranges
+                )
+
+            self.items = tuple(line[4:].strip().split() for line in lines if line)
```
divmain: `"string".split()` will output `["string"]`.  Is that the desired behavior here?

core/commands/tags.py
```diff
@@ -0,0 +1,318 @@
...
+                self.on_select_remote,
+                flags=sublime.MONOSPACE_FONT
+                )
+
+    def on_select_remote(self, remote_index):
+        """
+        Push tag(s) to the remote that was previously selected
+        """
+
+        #if the user pressed `esc` or otherwise cancelled
```
divmain: Sorry for the nit-pick, can you add a space after the `#`.

core/commands/tags.py
p3lim: I was thinking instead of doing it like this, use the default message as the 2nd param for `show_input_panel` above. That way the user will know that `{tag_name}` can be used for a replacement, and it also voids this check.

core/commands/tags.py
p3lim: The lines are in two parts: "sha tagname", I split it up so I can grab them by a list index below.

core/commands/tags.py
```diff
@@ -0,0 +1,318 @@
...
+        if remote_index == -1:
+            return
+
+        selected_remote = self.remotes[remote_index]
+
+        sublime.status_message(START_PUSH_MESSAGE)
+        if self.push_all:
+            self.git("push", selected_remote, "--tags")
+        elif hasattr(self, "items") and self.items:
```
divmain: Will `self` ever not have an `items` property in except when doing a `push_all`?  Wondering if this could be a simple `else:`.

p3lim: You have a point, it can be empty however.

core/commands/tags.py
```diff
@@ -0,0 +1,318 @@
...
+            refs = ""
+            for item in self.items:
+                refs += "refs/tags/" + item[1] + " "
```
divmain: It looks like the `.split()` does do something, since you're grabbing the second element of the list here.  What is the first element?

core/commands/tags.py
```diff
@@ -0,0 +1,318 @@
...
+    def run(self, edit):
+        lines = util.view.get_lines_from_regions(
+            self.view,
+            self.view.sel(),
+            valid_ranges=view_section_ranges[self.view.id()]
+            )
+
+        items = tuple(line[4:].strip().split() for line in lines if line)
+
+        if items:
```
divmain: No need for `if items:` - if `items` is empty, the `for` loop will be a no-op.

core/commands/tags.py
```diff
@@ -0,0 +1,318 @@
...
+            for item in items:
+                self.git("log", "-1", "--pretty=medium", item[0], show_panel=True)
+                break
```
divmain: Not finished?

core/commands/tags.py
p3lim: First element is the sha1 hash, second is the tag ref.

core/git_mixins/tags.py
```diff
@@ -0,0 +1,27 @@
...
+    def get_tags(self, remote=None):
+        """
+        Return a list of TagDetails object. These objects correspond
+        to all tags found in the repository, containing abbreviated
+        hashes and reference names.
+        """
+        entries = []
+
+        stdout = self.git("show-ref", "--tags", throw_on_stderr=False)
+        procelain_entries = stdout.split("\n").__iter__()
```
divmain: Typo: `procelain_entries`

core/commands/tags.py
p3lim: Don't really know if there are any additions necessary for this, maybe a syntax theme if that's possible to apply to a panel.

core/git_mixins/tags.py
```diff
@@ -0,0 +1,27 @@
...
+
+        for entry in procelain_entries:
+            if not entry:
+                continue
+            sha = entry[:40]
+            tag = entry[51:]
+            entries.append(TagDetails(sha, tag))
+
```
divmain: The above can be expressed as:
```python
# First forty characters is the commit hash, followed by `refs/tags/`, followed by tag name.
entries = [TagDetails(entry[:40], entry[51:]) for entry in porcelain_entries if entry]
```

core/commands/tags.py
divmain: Got it, thanks for the clarification.

core/commands/tags.py
divmain: I'll have to remove that then. :)

core/commands/tags.py
divmain: That would be more explicit.  I suppose it would be strange to be prompted for a message, you enter an empty string (for some reason) and it turns out that the message is not what you entered.  Providing a pre-filled default is a good idea, and probably better user experience.

core/commands/tags.py
divmain: :+1: I just hadn't visualized what each line would look like.

core/commands/tags.py
divmain: I suppose I'm confused by the for loop.  If the intention is just to show a panel for the first item, this could be written as:
```python
if items:
    # items[0][0] will be the sha1 of the first item
    self.git("log", "-1", "--pretty=medium", items[0][0], show_panel=True)
```

divmain:
This is really solid work.  Thanks again!  A few minor comments, but definitely nothing major.

core/commands/tags.py
p3lim: It could, I'll change it.

core/commands/tags.py
p3lim: I'd rather go the regex route, `check-ref-format` is not 100% accurate in my tests.

p3lim:
Are you by any chance on IRC or any other form of IM?
Got an issue I'd rather talk with your on a instant-ish platform.

core/commands/tags.py
divmain: Sounds good to me.

divmain:
We use Slack internally, so I don't have an IRC client installed at the moment.  Will also be heading to bed soon - pretty late here in Seattle.  Let me see if I can find a quick way to connect.

divmain:
@p3lim can try this... http://tlk.io/gitsavvy

p3lim:
To use `git check-ref-format` I need access to stderr in GitCommand, any suggestion?
(The regex route was just too messy).

divmain:
@p3lim - If you do end up going the `check-ref-format`, you can avoid a GitSavvy error condition by passing `throw_on_stderr=False` to `self.git`.  Dropping [this link](http://stackoverflow.com/questions/12093748/how-do-i-check-for-valid-git-branch-names) to the StackOverflow article for my future reference in case you go with the regex strategy.

p3lim:
I'm aware of that argument, I've used it in 4f03b3ad005d4e57d7dce527e1193084c84dc917.
I've done some major work on how the view is displayed, filled and updated today, I'll push it soon and the PR should be ready for merging.

p3lim:
@divmain Give that a go, there are two missing pieces to this puzzle, left marked off in the PR description. It should be feature complete and ready for merge, but I'd give it a test first, just in case.

divmain:
Awesome, I'll pull it down and give it a test drive.

p3lim:
There is one future feature I was thinking about however;

Having multiple complete lists of all the tags, especially when dealing with multiple remotes, will make the list of tags become extremely large over a period of time. I was thinking of this change:

If the tag on a remote is present locally, suffix it to the line in the local section instead of adding it to it's own section. This way, only tags not present locally but on a remote would end up in the remote section(s). Other than that it would function as it does now.

This will basically reduce the view's size drastically when you're working on a large project with many tags.

p3lim:
Also another change would be to sort the tags by most recent on top.

divmain:
Re: appending remote to local tag section - that seems fine, considering the key-bindings you've configured.  If there was a `delete remote` that might be more opaque.  I don't mind waiting for that change if you'd like to include it in the PR.

p3lim:
If the tag is deleted locally it would show up again in the remote section(s).

divmain:
Yeah, that's fine.  One other thing to consider (you may have already) is what the behavior will be if the remote is unavailable.  I.e. off wifi, off VPN, etc.

p3lim:
> Yeah, that's fine. One other thing to consider (you may have already) is what the behavior will be if the remote is unavailable. I.e. off wifi, off VPN, etc.
It would simply fail to retrieve them and display a message that says "Your repository has no tags."

divmain:
Alright, I think this looks good.  I'm going to merge and any additional work can go in separately.

p3lim:
I'm going to open a few issues that would complete this feature, as mentioned here.

PR #95, opened 2015-03-17
GitHub integration doesn't work with multiple remotes, especially when the remote is not GitHub

p3lim:


divmain:
Hmm... this won't work for enterprise GitHub instances.  I've been thinking of a different approach, where first it would try `origin` (in 90% of cases, this will probably suffice) and fall back to alphanumeric order (the current behavior).

p3lim:
I've never used GitHub enterprise, but since you mention that this won't work with it, is enterprise deployed to a server? I've always thought it was just a premium version of GitHub.

divmain:
No, you're correct. As far as I've been able to see, there isn't a straightforward approach to handle all cases without user intervention. Enterprise GitHub instances are self-hosted and may not have "github" in the FQDN. Furthermore, it is also possible for multiple remotes to be GitHub, enter .com or enterprise. In this case, there's no easy way to disambiguate and determine which one to use without perhaps calling GitHub APIs and following the fork tree for a given repo. That being the case, we should probably default to trusting that origin knows about issues, by convention, and then make an override configurable either globally or at the project level.

p3lim:
Sounds good.

yyyc514:
> That being the case, we should probably default to trusting that origin knows about issues, by convention, and then make an override configurable either globally or at the project level.

Project level please, where would this go?

divmain:
Project-level config could be edited by in `Menu > Project > Edit Project`, and sticking a value in there.  I'm concerned, however, that this data would be too loosely tied to the files in a repo.  In the terminal, I could easily see myself typing `cd project folder; subl some.file`.  When the project isn't opened directly, the project data will not be present.

A better approach may be to store this meta-data in the local repo's git config, and provide a GitSavvy mechanism to make the selection.  Something along the lines of the following under the hood:

```
> git config --local --add GitSavvy.issueRemote not_origin
> git config --local --get GitSavvy.issueRemote
not_origin
```

And then `github: set remote for integration`, present the user with a quick panel of remotes, and update the Git config.

divmain:
Alright, this should be good to go.  The user's selection is saved in the repo's git config.  If none has been set, it falls back to `origin`.

p3lim:
Great idea!

yyyc514:
>  In the terminal, I could easily see myself typing cd project folder; subl some.file. When the project isn't opened directly, the project data will not be present.

Yeah, this annoys me to no end.  One of the few places where I like TM's behavior better: when you open up a folder assume it IS the project. So good call on mixing it into git config.

divmain:
I suppose we could write a plugin for that :)  When a file is opened, if a project isn't defined for the window, add the project to the current window.

yyyc514:
How?  All the event listeners are for view except `on_window_command(window, command_name, args)`.  Does opening a window fire some sort of default command we could hook?  It's true you did say "file", and I suppose that would be _close_.

> add the project to the current window

I don't see an API call for that.

yyyc514:
https://github.com/yyyc514/subl_with_brains

divmain:
:)

Issue #13, opened 2015-02-12
Full-featured `git: graph` view


divmain:
- [x] syntax highlighted view
- [x] ~~pressing `super+enter` on a branch name checks out that branch~~
- [x] pressing `super+enter` on a commit will checkout that commit as detached HEAD
- [x] pressing `enter` on a commit displays the full commit message and diff in a separate window

PR #96, opened 2015-03-17
Add syntax highlighting to Git graph output and clear selection

borgand:
This PR adds syntax highlighting to `graph` output and removes selection.

I am not a ST3 developer so please correct me if I made any mistakes.

gitsavvy graph tmlanguage

**Credit:** Inspired by https://github.com/chancancode/sublime-text-2-git


syntax/graph.YAML-tmLanguage
```diff
@@ -0,0 +1,26 @@
...
+  match: (?<= )(\(.*\))(?= )
+- comment: issue numbers
+  name: string
+  match: (?:[Ff]ix(?:e[ds])?|[Rr]esolve[ds]?|[Cc]lose[ds]?)?\s*(?:#\d+|\[.*?\])
+- comment: merges
+  name: git-savvy.graph
+  match: "Merge branch '(.*?)'(?: (?:of|into) (.*))?"
+  captures:
+    '1': {name: string}
+    '2': {name: string}
```
divmain: Please add a new-line at the end of the file.

core/commands/log_graph.py
```diff
@@ -17,10 +17,12 @@ def run(self):
         view = self.window.new_file()
         view.settings().set("git_savvy.log_graph_view", True)
         view.settings().set("git_savvy.repo_path", repo_path)
+        view.set_syntax_file("Packages/GitSavvy/syntax/graph.tmLanguage")
         view.set_name(LOG_GRAPH_TITLE)
         view.set_scratch(True)
         view.set_read_only(True)
         view.run_command("gs_log_graph_initialize")
+        view.sel().clear()
```
divmain: Instead of adding this line here, modify the last line of `GsLogGraphInitializeCommand#run` to the following:
```python
self.view.run_command("gs_replace_view_text", {"text": branch_graph, "nuke_cursors": True})
```

divmain:
Thanks for this!  Just a couple of comments, and I'll merge this in.

borgand:
Implemented your comments.

BTW, I noticed that some themes (one of my favourites - Tomorrow)  put `constant.other` together with regular `foreground` which means the branch names won't be different from foreground colour.

Where do you think the problem lies - should the graph use some other colour for branch names or should I turn to theme editor for a fix?

divmain:
GitHub is showing a conflict and disabling Merge button, but I'm not sure where the conflict is.  I'll pull down this branch and merge into master manually.

Wrt the coloring, I'm not 100% sure of the best approach.  I've only been checking against Monokai, so this may need to be addressed broader than here in this view.  If you have time, would you mind opening a ticket with screenshots of the places where this is a concern?

divmain:
Relates to #13.

PR #97, opened 2015-03-17
Add `--all` to graph command to show other branches (esp. origin)

borgand:
Add `--all` to `graph` log command so that other branches become visible.

This is helpful in branched workflow (e.g. git-flow) where after a pull on master one needs to see how feature branches have diverged.

Screenshot:

gitsavvy graph-all


divmain:
Relates to #13.

Issue #107, opened 2015-03-19
Show what will/won't be committed same as default git commit template


yyyc514:
IE:

```
Fix: this commit fixes xyz

## To make a commit, type your commit message and press SUPER-ENTER. To cancel
## the commit, close the window. To sign off the commit press SUPER-S.

## You may also reference or close a GitHub issue with this commit.  To do so,
## type `#` followed by the `tab` key.  You will be shown a list of issues
## related to the current repo.  You may also type `owner/repo#` plus the `tab`
## key to reference an issue in a different GitHub repo.

# On branch master
# Your branch is up-to-date with 'origin/master'.
# 
# Changes to be committed:
# 
# 	new file:   project.sublime-project
# 
# Untracked files:
# 
# 	project.sublime-workspace
```

We can get this from `git commit --dry-run` with a few small tweaks.  I like the single pounds to differentiate it from the above help, but we could also use ##.

yyyc514:
It would be nice to get the meat (what's changed) a little higher... could the notes about github issues be placed below?  Thoughts?

yyyc514:
https://github.com/divmain/GitSavvy/pull/111

p3lim:
Try setting the option `show_commit_diff` to true, it'll show a diff for the commit.

yyyc514:
Perfect IF what I wanted was a diff.  What I'm going for is to match the default git behavior for editing commits.

p3lim:
Ahh, now I know what you mean.

divmain:
I'm going to tackle this myself.  I don't see any reason why it can't be toggle-able, ala the remote branches in the `git: branch` dashboard.  Probably the same with the diff.  I'll include syntax highlighting in that effort.

yyyc514:
I'm open to helping with the larger refactoring also.  Just trying to hit the ground running on a few things and I didn't realize there was a whole new interface paradigm until you pointed it out. :)

divmain:
This will be included as part of #50.  Closing in favor of that issue.

Issue #50, opened 2015-03-08
Enhance commit view interactivity


divmain:
Follow-up for changes seen in #45.
- [ ] convert commit view to `ui.Interface`
- [ ] enable region-based retrieval of commit message
- [ ] toggle-able display of relevant diff
- [ ] toggle-able display of commit summary (ala `git commit --dry-run`, see #107)
- [ ] add embedded syntax-highlighting for diff (defining scope for region)
- [ ] **investigate:** stage files from commit view

divmain:
- [ ] convert commit view to `ui.Interface`
- [ ] enable region-based retrieval of commit message
- [ ] in-view toggle-able display of relevant diff
- [ ] add embedded syntax-highlighting for diff (defining scope for region)

yyyc514:
Why we're at it any objection to staging files from the git commit when users have turned on viewing commit summaries?

alextegelid:
I often forget to stage some file and I don't think it's a big deal closing the commit message, returning to the status view and stage the remaining file/files from there

yyyc514:
Yeah, I've kept doing the same thing and if I'm seeing the problem on the commit view I'm not sure why I shouldn't able to just solve it right there.

divmain:
Hmm, I'll think on that...  Updating description to reflect possible enhancement.

divmain:
Related to #89.

kaste:
Relevant parts have been implemented here.  You can unstage and discard right in the diff.  You cannot *stage* as we're only showing what gets committed.  T.i. it is either staged or we're in `--all` mode.  

Unstaging in `--all` mode is not implemented and would be interesting, maybe.  Since technically nothing is staged in that mode, we would stage everything but the hunk/line the user wanted to unstage.  And the of course switch away from `--all` mode to the normal commit-whats-staged mode.  🤔

PR #111, opened 2015-03-20
Commit: show summary when commiting

yyyc514:
Relates to #107.

### DIFF
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -63,7 +63,7 @@ class GsCommitInitializeViewCommand(TextCommand, GitCommand):
     def run(self, edit):
         template = [
             self.commit_message(),
-            COMMIT_HELP_TEXT,
+            COMMIT_HELP_TEXT.lstrip(),
             self.footers()
         ]
         content = "".join(template)
@@ -83,7 +83,7 @@ class GsCommitInitializeViewCommand(TextCommand, GitCommand):
             with open(merge_msg_path, "r") as f:
                 return f.read()
         else:
-            return ""
+            return "\n\n"
 
     def footers(self):
         footer = ""
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -5,6 +5,8 @@ from sublime_plugin import WindowCommand, TextCommand
 
 from ..git_command import GitCommand
 
+NO_STAGED_FILES = "No files are currently staged.\n" \
+                  "Perhaps you meant to commit include unstaged files?"
 
 COMMIT_HELP_TEXT = """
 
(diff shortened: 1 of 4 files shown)


yyyc514:
If this is heading the right direction I can improve the syntax further.

yyyc514:
I bumped the help to `###` cause I thought it looked better.

core/commands/commit.py
```diff
@@ -57,11 +59,37 @@ class GsCommitInitializeViewCommand(TextCommand, GitCommand):
...
+                filtered_lines.append(line)
+
+        return "\n".join(filtered_lines)
+
     def run(self, edit):
         merge_msg_path = os.path.join(self.repo_path, ".git", "MERGE_MSG")
 
-        option_amend = self.view.settings().get("git_savvy.commit_view.amend")
-        if option_amend:
+        if self.amending():
```
divmain: This is an unnecessary change.

core/commands/commit.py
```diff
@@ -71,11 +99,14 @@ def run(self, edit):
...
 
         if sublime.load_settings("GitSavvy.sublime-settings").get("show_commit_diff"):
-            if option_amend:
+            if self.amending():
                 initial_text += self.git("diff", "HEAD^")
             else:
                 initial_text += self.git("diff", "--cached")
 
+        if sublime.load_settings("GitSavvy.sublime-settings").get("show_commit_summary"):
+            initial_text += self.fetch_commit_summary()
```
divmain: Just pass in `option_amend` here.

core/commands/commit.py
```diff
@@ -57,11 +59,37 @@ class GsCommitInitializeViewCommand(TextCommand, GitCommand):
...
+        """
+
+        args = []
+        if self.amending():
+            args.append("--amend")
+
+        dryrun = self.git("commit", "--dry-run", *args, throw_on_stderr = False)
+        # detect trying to do a commit with nothing staged
+        if dryrun.find('no changes added') != -1:
+            sublime.error_message(NO_STAGED_FILES)
```
divmain: If you go this route, there should be an exit mechanism - after showing this error message, it'll continue one with executing this method.

yyyc514: I wasn't sure what the proper thing to do was since at that point we already have the view (it's created by the first command at the window level).

divmain:
I think a more straightforward approach here would be something along the lines of the following.

```
status = self.git("status")
status_comment = "\n".join("## " + line for line in status.split("\n"))
```

Really, the normal behavior here for commits is to display the verbose `git status` output.

core/commands/commit.py
yyyc514: Is there some reason we can't wrap it in a method vs passing it everywhere?  `amending` really does seem to be a property of `GsCommitInitializeViewCommand to me`.

core/commands/commit.py
```diff
@@ -5,16 +5,18 @@
...
+### the commit, close the window. To sign off the commit press {key}-S.
 
-## You may also reference or close a GitHub issue with this commit.  To do so,
-## type `#` followed by the `tab` key.  You will be shown a list of issues
-## related to the current repo.  You may also type `owner/repo#` plus the `tab`
-## key to reference an issue in a different GitHub repo.
+### You may also reference or close a GitHub issue with this commit.  To do so,
+### type `#` followed by the `tab` key.  You will be shown a list of issues
+### related to the current repo.  You may also type `owner/repo#` plus the `tab`
+### key to reference an issue in a different GitHub repo.
```
divmain: I'd prefer to keep it at two `#`.

yyyc514:
`status = self.git("status")`

There is no --amend option for status.  Since what we want is to know exactly what the commit is planning to do there is no substitute for asking the commit command itself.

divmain:
Hmm that's a very good point @yyyc514 .  Let me think on this a bit longer.  Sorry if it feels like I took an axe to this - just got off a plane and pretty tired.  Very much appreciate the PR.  As much as possible I prefer to keep things a simple as possible until a clean abstraction becomes apparent, which is where most of my comments are coming from.  I'll re-review this tomorrow.

syntax/make_commit.YAML-tmLanguage
```diff
@@ -6,7 +6,15 @@ uuid: 0d4c19fb-b197-4d90-806e-7212e542c86b
...
 - comment: commit header
   name: comment.git-savvy.make-commit
-  match: ^##.*
+  match: ^###.*
+- comment: blah
+  name: comment
+  begin: ^#[^#]
+  end: $
+  patterns:
+  - comment: git gile categories
```
divmain: `git file categories`?

yyyc514: Open to suggestions of a better name :)

divmain: I was just pointing out the typo :)

yyyc514: And I still didn't see it. LOL.  Ok.

core/commands/commit.py
yyyc514: My next idea was to clean this up so it's just content = " ".[header(), help_text(), footer()].  Since lots of different pieces of the render need to know about amend it seems logical to make it accessible on the class vs passing it into every render method.

divmain:
Alright, I'm going to hit the sack, but I'll take a look at this again tomorrow.  Thanks again.  One last nit-pick before I sign-off - the top method in each class should be `run`, for consistency.

yyyc514:
> One last nit-pick before I sign-off - the top method in each class should be run, for consistency.

Oh sure, that's no problem.

yyyc514:
I'll be more than happy to port this stuff to the new Interface framework when that is all polished up.

yyyc514:
Rebased and squashed.

divmain:
I'm traveling for the next few days and will want some time to dig into this before merging.  This PR will have to wait for a couple of days.

yyyc514:
I've rebased and removed the error case.  Currently commit is not an error case if there are no files staged, so this probably shouldn't be either.  Plus if there are no files staged it's pretty easy to notice that when you're showing the dry-run.

Ping.

divmain:
Not ignoring this one.  I'm mulling over how I want to approach the commit view, after working with the other interfaces.  Leaving this PR open and, even if it is not merged, I'll reference what you did here.

stoivo:
Hey, how is the planing going

divmain:
Closing this, but issue will be addressed as part of #50.

Issue #104, opened 2015-03-19
Reopen views in same tab


jipumarino:
Currently, when choosing "git: status" or any other view from the command palette, a new tab/sidebar entry with the interface is opened. I think it's better to just open the same one if it's already opened.

This image can clarify what I mean:
![branches scribd scribd 2015-03-19 14-12-15](https://cloud.githubusercontent.com/assets/2559/6736243/5f5d7270-ce42-11e4-886d-5376d97edd4f.png)

alextegelid:
Yes! That's something that I've been thinking of aswell. That'd be awesome👍

divmain:
I think as I refactor the status and tag dashboards, this should be easier to do.  I'll start with branch dashboard as proof of concept, and if it works out, the others will be addressed as part of the refactor.

yyyc514:
Seems like you could replace/supliment `get_read_only_view` with `create_or_find_read_only_view(context, name, title)` and then just search existing views for a matching `title`, no?  If title matches (and it's a gitsavvy view) return that view, otherwise return a new view?

Sound about right?  Trying to find someplace easy I could start helping contribute.

yyyc514:
It might work with name alone, I'd have to review how repo_path works and see if you can't open up a higher level directory that has two different git repositories in subduers.

divmain:
@yyyc514, if you'd like to take a look at this, start with the branch dashboard and the UI module.  I plan to port over existing functionality from status and tags views soon, as this will provide a more solid base for enhancement (and consistency) moving forward.

Wrt this ticket, since it is only relevant for interfaces/interfaces-to-be (and not other types of views), there is already the interfaces dictionary you can iterate over.  For each `interface` in `interfaces`, data to use for comparison should include:

- `interface.type_name`
- `interface.view.window().id()` (same Git repo could be open in multiple windows)
- `interface.repo_path` (`Interface` class may need to inherit from `GitCommand` - and I'll eventually have to rename that base class, as it doesn't really fit anymore), OR an abstraction of `repo_path` that could be called from both `GitCommand` instances and the interface

divmain:
This is in master now.  Please give it a try and let me know if you encounter any issues.

PR #113, opened 2015-03-20
reuse existing views for interfaces

yyyc514:


yyyc514:
Attempt at #104.  Thoughts?

yyyc514:
I'm thinking maybe every Interface takes a single named argument and that is `repository_path`?  We can't count on storing it in the view because an interface doesn't have a view until after we search for an existing view, and to do the search we need to know the repository_path (to build the view_path).

yyyc514:
Rather than do a super complex comparison I just let ever interface set it's own view-path (should it be interface path?) then it loops over interfaces trying to match the path.

yyyc514:
If you want to reuse the whole interface itself I'd have to go back and refactor this... right now a new Interface is created but it ties to find and use an existing view.

divmain:
Closed in favor of ecabcad7b098097c6b626cc3d71d61423bec547e.

PR #115, opened 2015-03-20
fix: reloading breaks interface refreshing

yyyc514:


yyyc514:
Anything holding this up?

yyyc514:
Order is going to matter more as the source grows, but this is a good start.  Another of my branches doesn't work without more explicit ordered because when you start to use the objects more freely you run into issues about the order in which the objects are imported into different files.

divmain:
Lost track of this one.  Merging.

yyyc514:
Thanks!

PR #116, opened 2015-03-20
fix: interfaces should survive restarts

yyyc514:


yyyc514:
I can review this on Monday and tweak it to roll a little better with the new interface changes.

divmain:
Can you clarify what this solves?  As it is, the branch dashboard will survive Sublime exit and re-open.  That logic lives here:
https://github.com/divmain/GitSavvy/blob/3098732961bc669a380011633935f9ed4600679c/common/ui.py#L222-L232

yyyc514:
- Open branches
- Change branch a few times
- Switch back to branch tab
- Quit Sublime
- Reopen
- Hit 'c' to try and switch branches.

Python blows up because there are no interfaces to return from `get_interface` (they are lost when the program exits).  That is what this fixes.

yyyc514:
We'd need to make sure a refresh was always triggered before `GsNewContentAndRegionsCommand`, which does not seem to be the case now.

yyyc514:
Ok I see that.  So I tried this instead:

```
def get_interface(view):
    iff = interfaces.get(view.id(), None)
    if not iff:
        view.run_command("gs_interface_refresh")

    return iff or interfaces.get(view.id(), None)
```

It doesn't work though because I think gs_interface_refresh is run async so it's not finished when we try and get the interface again and return it.  We would need to use a version that blocks.  And you'd still have to pass `view` around to get access to `run_command` I think.

yyyc514:
It's actually `GsBranchesCheckoutCommand` that runs when you hit 'c' and fails to get_interface.

yyyc514:
Ok, what about that?  `get_interface` calls refresh (sync) if it fails the first time... and interfaces that take an existing view need to call `render()` to fully setup the state properly.

divmain:
The issue here is that when Sublime re-opens views upon starting, the `activated` event does not fire for the view that is active when all plugins are finishing loading.  If you waited for plugins to load, went to a different view, and returned to the branch dashboard, you'd find that the dashboard functioned as expected.  I've resolved this issue in 5b70e53.

Considering the volume of changes you've been pushing up, I feel bad with the number of PRs that haven't been merged.  If you reach out and point to the issues you'd like addressed, I'd be more than happy to point you in the right direction with how to contribute.

yyyc514:
> . If you waited for plugins to load, went to a different view, and returned to the branch dashboard, you'd find that the dashboard functioned as expected. I've resolved this issue in 5b70e53.

Yep, I saw that. My first impulse was to fix the actual low-level error (interfaces disconnected from views) rather than to just prevent it.  Preventing it is a little simpler and more elegant, so it makes sense here.  And if the rest of the code is well written then the low-level error should never occur in the first place.

> Considering the volume of changes you've been pushing up, I feel bad that so few PRs have been merged.

Oh I don't mind in cases like this... where it's pretty clear your familiarity with the codebase and Sublime is trumping my growing understanding of both.  It is a bit more annoying when we simply seem to disagree on things where the "correct" solution is much less clear (or arbitrary). :-)  I think it'd be great to chat sometime on IRC or something if you have time.

GitSavvy is what drug me away from TextMate and I've seen it as a great opportunity to learn what's possible with Sublime (and learn Python in the process).  You're just starting to touch the surface with Interfaces.  I think the abstraction can get a lot higher still and would love to improve some this stuff together going forward to allow some pretty impressive interfaces to be built (and expanded by users wanting to customize things, etc).

divmain:
Great :+1:

I agree, it will be interesting to work on some higher-level abstractions that can be used across the board.  In my research, I haven't found anything truly re-usable, and it might be interesting to try and develop the Sublime-equivalent of a web framework for interfaces.  However, given that GitSavvy was originally created to scratch my own itch, I've been addressing these things where it become clear the ROI is high.

Anyway, thanks for being flexible.  And thanks for the thoughts and contributions!

PR #117, opened 2015-03-20
fix: cursor should not jump when toggling remotes

yyyc514:


yyyc514:
Perhaps util.save_cursor_position(view), util.restore_cursor_position(view) helpers?

yyyc514:
Ping.

divmain:
Please see the comment and explanation for 6eda35a.

Issue #83, opened 2015-03-12
Custom commands


sentience:
I frequently want to pull the latest upstream changes to a branch and then re-apply my own unpushed changes. This is done from the command prompt using `git pull --with-rebase`.

I’m able to do this with the SublimeGit package using its support for [custom commands](https://docs.sublimegit.net/customizations.html#customizations-commands).

divmain:
- [ ] take an array of args like `GitCommand#git`
- [ ] support string interpolation for arguments for `{FILE_PATH}` and `{REPO_PATH}`
- [ ] support non-git commands (that nonetheless rely on running in project's root dir)
- [ ] support output to panel or new scratch view
- [ ] accept flag to run as shell command, to support basic sh capabilities

yyyc514:
What is the advantage of an array of args (in this context) vs just a command string?

divmain:
@yyyc514, what comes to mind:

1. Consistency across the app.
2. More explicit arguments without needing to escape spaces and other characters, and having to know how Python, Sublime settings parser, and/or the OS would treat it.
3. Simpler implementation for special keyword interpolation like `REPO_PATH` and `FILE_PATH`.
4. Only a slight hit to UX.

yyyc514:
Why not allow both for simplify and maximum configurability? :-)  Actually I don't think you can stop that... since all you're doing (in the simple case) is simplest case is string concat. :-)  So n/m. :-)

max-mykhailenko:
Will I have ability to run sublime shortcut command (actually another plugin) on every commit view? I have plugin that count coding time in Sublime and I want write it to every commit (https://github.com/max-mykhailenko/memTask)

yyyc514:
Regarding the command class structure I'm thinking:

- ExternalCommand
- GitCommand (ExternalCommand)
- GsSomeRandomViewCommand (GitCommand, RepoPathMixin)
- GsSomeRandomViewCommand (ExternalCommand, CurrentPathMixin)

The thought being you could use ExternalCommand or GitCommand stand-alone (from globals, utils, etc)... and then if you wanted to use them from a view you'd lay in RepoPathMixin or CurrentPathMixin to set the paths properly based on the file you are currently viewing.  Though really GitCommand and ExternalCommand are mixins the way we're using them so perhaps the names should also change to reflect that?

Thoughts?

yyyc514:
Here is what I've written and started using until we have something better built-in:

https://github.com/yyyc514/Tint

[...]

PR #118, opened 2015-03-21
Add the command `git: gui`.

zlsun:


divmain:
Output on my machine:
```
> git gui
git: 'gui' is not a git command. See 'git --help'.

Did you mean one of these?
	pud
	pum
```

What version was the first to ship `git gui`?  Also, this seems to be a good use case for #83.  That's coming up soon in my queue - do you think that would suffice @zlsun?

zlsun:
@divmain , sorry for my carelessness, Git Gui is a build-in tool on Windows but not Linux/OSX.
Custom commands looks like a good solution. I will wait for it.

divmain:
No worries @zlsun!

asfaltboy:
@zlsun  - Actually, [git-gui is based on Tcl/Tk](http://git-scm.com/docs/git-gui), and it is cross-platform-compatible; e.g.: it also work on my OSX and Fedora machines (as well as windows). 

@divmain - Can I ask which version of git you use? `git-gui` was in fact added since git 1.5 series. Since it's been at least 5 years now, I believe it's safe to assume most people will have it. Additionally, we can also test for existence of this tool (and others e.g `git flow`) before enabling command.

`git-gui` exposes 4 commands that wrap other git commands such as blame, diff, log, etc.
For example, I find `git gui` very useful in some specific cases such as blame "digging"; i.e.: it's trivial to perform "Blame parent commit" from within `git gui blame`.

asfaltboy:
By the way, simply running `git gui` defaults to `git gui citool` which is for a single commit action, and is not that useful IMHO.

Perhaps we ought to add separate commands for `git gui blame`, `git gui browser` and `git gui citool` ?

divmain:
This seems to be enough of an edge case that I'm inclined to consider this "complete" with the implementation of #83.

Issue #41, opened 2015-03-07
Pedantic commit messages


p3lim:
Add a distinct style guide for the commit messages:
- First line should be max 50 characters
- Second line should be empty
- Any subsequent lines should be max 72 characters

Coloring the background of any characters that exceed limits is ideal, only warning the user.

This should be optional, of course.

References:
http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
https://github.com/blog/926-shiny-new-commit-styles

yyyc514:
See #119.  Someone just needs to make a kick-butt syntax for commits in general.

yyyc514:
Related: https://github.com/adambullmer/sublime_git_commit_syntax/pull/2

Spent a little time on this.  If I can't get it back it mainline I'll probably create a whole new repository and package.  Still tweaking the styles.

yyyc514:
![screen shot 2015-03-21 at 6 26 41 am](https://cloud.githubusercontent.com/assets/6473/6764679/607ce3a0-cf93-11e4-9549-ccb05da0a4f3.png)

yyyc514:
@divmain Would you like this kind of thing in GitSavvy proper?

yyyc514:
Having multiple syntax files for different modes gets messy at some point (though it might be ok for only two options).

divmain:
Please see my comment in #119.  I'm not averse to including a "pedantic commit message syntax" in GitSavvy proper - it can be commented out as an alternative in the default `GitSavvy.sublime-settings` file.

divmain:
For anyone interested, @yyyc514 may have a custom syntax definition for commit views that you can use.  However, as I commented in #119, I'm not 100% I can preserve this new behavior while moving the commit view to being more interface-like.

I'll leave this open until the feature is officially announced as part of a changelog.

stoivo:
hey, @yyyc514!
How did you make the commit view look like this?
I cloned https://github.com/adambullmer/sublime_git_commit_syntax
now when I open Commit view and  run in console 
```
>>> view.settings().get('syntax')
'Packages/sublime_git_commit_syntax/commit-message.tmLanguage'
```

And my diff look really boring. Now there are no color at all

yyyc514:
Try my fork:

https://github.com/yyyc514/sublime_git_commit_syntax

stoivo:
Still, not working. What theme are you using?

yyyc514:
[...]

PR #119, opened 2015-03-21
feature: allow using your own COMMIT_EDITMSG syntax

yyyc514:


yyyc514:
Ok, I've squash this down to JUST this change without anything else.

core/commands/commit.py
```diff
@@ -44,7 +45,14 @@ def run_async(self, repo_path=None, include_unstaged=False, amend=False):
...
         view.settings().set("git_savvy.repo_path", repo_path)
-        view.set_syntax_file("Packages/GitSavvy/syntax/make_commit.tmLanguage")
+
+        gitsavvy_settings = sublime.load_settings("GitSavvy.sublime-settings")
+        if gitsavvy_settings.get("use_syntax_for_commit_editmsg")
+            syntax_file = util.file.get_syntax_for_file("COMMIT_EDITMSG")
+            view.set_syntax_file(syntax_file)
+        else:
+            view.set_syntax_file("Packages/GitSavvy/syntax/make_commit.tmLanguage")
+
```
divmain: Overriding the default syntax is a good approach to resolve #41.  However, let's simplify by creating a setting in `GitSavvy.sublime-settings` called `commit_syntax` with a default value of `Packages/GitSavvy/syntax/make_commit.tmLanguage` (and remove `use_syntax_for_commit_editmsg`).  Then, replace the above chunk of code with:
```python
gitsavvy_settings = sublime.load_settings("GitSavvy.sublime-settigns")
view.set_syntax_file(gitsavvy_settings.get("commit_syntax")
```

yyyc514:
I felt this was a lot simper for veteran an novice user alike - *and makes use of the existing setup that Sublime already provides for choosing which template you want for particular files.*  We don't need to re-invent the wheel here,  Sublime already has the feature to let pros decide exactly which syntax applies to exactly which files.  We just need to make it clear if GitSavvy is in charge or some other plugin the user has decided to install and configure.

Honestly I'm not even sure how to easily get the full path of a syntax file inside a plugin I just installed without opening up the settings dir and looking manually and typing it out.

yyyc514:
If there is a clearer name I'd be for that, but I think a True/False toggle is more in the spirit of Sublime and keeping things truly simple.

yyyc514:
I'm not even so sure the default behavior should be:

- check for COMMIT_EDITMSG syntax
- if found, use that
- if not, use our built-in

But then I prefer guessing in the 90% cases and saving configuration for the remaining 10.

divmain:
I'll consider this.  Can you point me to other packages that take a similar approach?  I'm hesitant to go this direction because the commit view is an interactive view, not a file type.  That being the case, assigning syntax by a fake file extension seems like an opaque approach to me.

yyyc514:
> That being the case, assigning syntax by a fake file extension seems like an opaque approach to me.

I don't see it that way at all.  That's what git calls it's PRE-COMMIT edit file... so it just seemed the simplest thing to allow someone to reuse a theme that already targeted that file.  It really comes down to what the view is and does.  I'd even suggest changing the name of the view to COMMIT_EDITMSG, but I didn't think you'd go fo that.

Right now it's really a glorified COMMIT_EDITMSG.  It might do more in the future, but that's how I'm thinking of it.  Obviously if it starts doing really wacky things that aren't compatible with the COMMIT_EDITMSG syntax then this would be a moot argument - but if what it does it a million things WITHIN this format then it still seems maybe a good idea to allow an easy toggle.

What about two settings?

- commit_syntax
- use_syntax_for_commit_editmsg

Or would you allow special meanings of some values?

- commit_syntax: "" / "default" (internal syntax)
- commit_syntax: "COMMIT_EDITMSG"
- commit_syntax: "/your/favorite/file.tmLanguage"

yyyc514:
Of course the settings would be documented in the sample config file.

yyyc514:
> Can you point me to other packages that take a similar approach?

I have no idea.  My point is that it's the approach of Sublime itself - to color files based on what they are - determined by extension.  This is a pre-commit git message view.  The fact that we allow adding commit summary, diff, etc. to that view (just like git itself) make that even more clear.  So until it differs substantially from that type of view it seemed to make sense to "color files based on what they are", as I put it.

>This is a pre-commit git message view.

I'm not sure that changes as it gets more powerful.  I'd actually love to be able to stage files right from the summary view on the commit page... but if I could that wouldn't change me thinking of the view as a pre-commit git message view.

Obviously those using a less-specific syntax might have fewer coloring feature (if you start adding keystroke colors, etc), but that would be their choice.  I'm not suggesting make this default, only an option.

yyyc514:
> commit_syntax: "COMMIT_EDITMSG"

My chief objection is not the name of the setting but the user having to waste time figuring out the exact path name of a syntax file - when they just want the editor to have the same behavior as it does for other "file types".

divmain:
Alright, I'm not really a fan of treating the filename like a filename extension, but I've decided its not terribly important to me, seeing as this feature is disabled by default.  So I'm going to merge this in.

Keep in mind, however, that the implementation and/or behavior may change once the commit view is converted to being more interface-like, due to constraints for Sublime's region-based syntax highlighting.  I'll do my best to preserve the functionality.  For that reason, I'm not going to include the change in a changelog until I'm sure the feature won't go away or significantly change.

yyyc514:
> Keep in mind, however, that the implementation and/or behavior may change once the commit view is converted to being more interface-like, due to constraints of Sublime's region-based multiple-syntax highlighting. I'll do my best to preserve the functionality.

At the very least then GitSavvy should (please) have an option to use a toggle to use an old "git style" commit screen (and related syntax) or the "new, amazing, and so completely unlike git commit that a syntax file can't work" dashboard/commit view.  We're all waiting on the edge of our seats to see what you want to do with git commit but honestly it's one of the few screens I haven't yet been hankering for any improvements too - beyond supporting what git already does from the console.

Maybe it's because I just don't know what is possible. :-)  I wanted to stage files from there but I think the new improvements to quick stage may have done the trick to scratch that itch.

yyyc514:
> Alright, I'm not really a fan of treating the filename like a filename extension, but I've decided its not terribly important to me, seeing as this feature is disabled by default.

Thanks. :-)  I'll see what I can do to move the Pedantic stuff forward now that this is in.

PR #120, opened 2015-03-21
syntax color git when commiting

yyyc514:


divmain:
Pre-cursor to some of the work in #50.

PR #121, opened 2015-03-21
Scroll to first diff (inline)

NicoSantangelo:
Hello again!
This PR makes GitSavvy automatically scroll to the first found diff on the file:

**before**
before

**after**
after

I don't know if this is a something you had in mind for the package but I find it very useful.

Let me know if you need me to change anything,
Thanks!

### DIFF
--- a/core/commands/inline_diff.py
+++ b/core/commands/inline_diff.py
@@ -141,10 +141,13 @@ class GsInlineDiffRefreshCommand(TextCommand, GitCommand):
         self.view.replace(edit, sublime.Region(0, self.view.size()), inline_diff_contents)
 
         if cursors:
-            self.view.sel().clear()
-            pt = self.view.text_point(row, 0)
-            self.view.sel().add(sublime.Region(pt, pt))
-            self.view.show_at_center(pt)
+            if (row, col) == (0, 0) and sublime.load_settings("GitSavvy.sublime-settings").get("inline_diff_auto_scoll", False):
+                self.view.run_command("gs_inline_diff_goto_next_hunk")
+            else:
+                self.view.sel().clear()
+                pt = self.view.text_point(row, 0)
+                self.view.sel().add(sublime.Region(pt, pt))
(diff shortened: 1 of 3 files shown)


divmain:
@NicoSantangelo, I like this idea.  I think I personally prefer the current behavior, but definitely see how some would prefer what you've done here.  Let's make it optional.  Add `inline_diff_auto_scoll` to `GitSavvy.sublime-settings`, set to `false` by default (we can change the default down the road if enough people are for it).  If you'd like, also mention this option in the user docs for the inline diff view.

core/commands/inline_diff.py
```diff
@@ -289,6 +289,16 @@ def highlight_regions(self, replaced_lines):
 
         self.view.add_regions("git-better-added-lines", add_regions, scope="git_savvy.change.addition")
         self.view.add_regions("git-better-removed-lines", remove_regions, scope="git_savvy.change.removal")
+        self.scroll_to_the_first_change(l)
```
divmain: Instead of taking the action here in `highlight_regions`, move the call to `scroll_to_first_change` up [here](https://github.com/NicoSantangelo/GitSavvy/blob/scroll-to-first-diff/core/commands/inline_diff.py#L150) right after the call to `highlight_regions`.  You should have access to the relevant line number as part of the `replaced_lines` list.

divmain:
Also, quick thought: I think this could be accomplished by simply running
```python
self.view.run_command("gs_inline_diff_goto_next_hunk")
```
instead of introducing a new method.

NicoSantangelo:
I actually searched for a command to go the the next hunk and couldn't find it, my bad :). I'll make the changes, test them and let you know.

Thanks a lot for taking the time to review the PR!

divmain:
No problem, glad to have your involvement!

NicoSantangelo:
If you have the time to review the setting comment and info in the README, my english is not my strongest point :).

divmain:
This looks good!

PR #124, opened 2015-03-22
use commit -a to add unstaged files when committing

yyyc514:


yyyc514:
Awesome!

divmain:
Thanks, this was a good find!

Issue #129, opened 2015-03-23
Error: maximum recursion depth exceeded in comparison


p3lim:
This error is shown in a panel every time a file without a source (e.g. opened from a zip) is opened or in the focused view.

The console shows a flood of the following error:

```
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\Administrator\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\git_command.py", line 96, in git
    cwd=working_dir or self.repo_path,
  File "C:\Users\Administrator\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\git_command.py", line 148, in repo_path
    return self._repo_path()
  File "C:\Users\Administrator\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\git_command.py", line 173, in _repo_path
    throw_on_stderr=throw_on_stderr
  File "C:\Users\Administrator\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\git_command.py", line 103, in git
    raise_error(e)
  File "C:\Users\Administrator\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\git_command.py", line 84, in raise_error
    raise GitSavvyError(msg)
GitSavvy.core.git_command.GitSavvyError: maximum recursion depth exceeded in comparison

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\Administrator\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\commands\status_bar.py", line 34, in run_async
    if not self.file_path or not self._repo_path(throw_on_stderr=False):
  File "C:\Users\Administrator\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\git_command.py", line 173, in _repo_path
    throw_on_stderr=throw_on_stderr
  File "C:\Users\Administrator\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\git_command.py", line 103, in git
    raise_error(e)
  File "C:\Users\Administrator\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\git_command.py", line 84, in raise_error
    raise GitSavvyError(msg)
GitSavvy.core.git_command.GitSavvyError: maximum recursion depth exceeded in comparison
```

yyyc514:
I could look at this is anyone knew how to easily reproduce on Mac platform.

p3lim:
[...]

Issue #128, opened 2015-03-23
GitSavvy views won't open if no tabs are present


alextegelid:
For example the status view won't open unless I open one or more files in tabs. Executing the "git: status" command outputs this in console

```
Traceback (most recent call last):
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime_plugin.py", line 526, in run_
    return self.run()
  File "/Users/alex/Library/Application Support/Sublime Text 3/Packages/GitSavvy/core/interfaces/status.py", line 97, in run
    repo_path = self.repo_path
  File "/Users/alex/Library/Application Support/Sublime Text 3/Packages/GitSavvy/core/git_command.py", line 148, in repo_path
    return self._repo_path()
  File "/Users/alex/Library/Application Support/Sublime Text 3/Packages/GitSavvy/core/git_command.py", line 163, in _repo_path
    file_dir = os.path.dirname(file_path)
  File "./posixpath.py", line 151, in dirname
AttributeError: 'NoneType' object has no attribute 'rfind'
```

Might this have something to do with reusing gitsavvys tabs if present? Maybe that function needs to check if any tabs are open before trying to reuse them. I'm not familiar with ST plugin development, so this is just a guess ;)

yyyc514:
Try PR #130

alextegelid:
Thanks! I'll wait for @divmain to merge it.

divmain:
@alextegelid, this should be fixed in master, along with #129.  Please let me know if you encounter any issues.

alextegelid:
Thanks! I'll continue to keep an eye open for further issues. Thanks again for writing this

PR #130, opened 2015-03-23
fix: repo_path broken for window with no files open

yyyc514:


yyyc514:
Or should i push this down into file_path?

core/git_command.py
```diff
@@ -159,7 +159,9 @@ def _repo_path(self, throw_on_stderr=True):
         repo_path = view.settings().get("git_savvy.repo_path")
 
         if not repo_path:
-            file_path = self.file_path
+            # if we are in an empty tab for a project self.file_path may
+            # return None in which case we check the windows folder
+            file_path = self.file_path or self.window.folders()[0]
```
divmain: If you look a few lines down, the window's folders are already checked.  The issue in #128 is that the method is that `None` is passed to a method that expected a string.  I've just pushed a fix for #128 and #129.

yyyc514: Yeah, I didn't spend enough time looking.

Issue #14, opened 2015-02-12
GitHub issues integration


divmain:
- [ ] list all issues for remote repository (special view)
- [ ] syntax highlighting
- [ ] open new issue (special view)
- [ ] close issue
- [ ] view details for issue, including comment history
- [ ] add comment
- [ ] view issue in browser

https://developer.github.com/v3/issues/

yyyc514:
I've been looking at this a little.  You didn't have a mockup laying around, did you?

yyyc514:
Since results are paged from the server we'd need to support paging somehow so that would have to fit in the UI.

yyyc514:
I assume we'd allow paging by keystroke... thoughts?

yyyc514:
I was imagining a detailed view and then less detailed view that you could toggle.  Idea for less detail:


```
GITHUB: divmain/GitSavvy

#139   Github Integration broken? [question]
#137   How to tell GitSavvy to forget git repository? [ux]
#136   TypeError: object of type 'NoneType' has no len() [bug] [investigate]
#135   Diff current file
#133   Inline diff adding multiple instances of the same row [bug]
#131   Easier to read branch summary on status view [ux]
#127   Support for non-UTF-8 [new feature]
#125   textwrap.dedent
#119   feature: allow using your own COMMIT_EDITMSG syntax
#117   fix: cursor should not jump when toggling remotes
```

yyyc514:
![screen shot 2015-03-25 at 10 02 59 pm](https://cloud.githubusercontent.com/assets/6473/6839145/bcdb2062-d33a-11e4-8ee0-939d3cf54508.png)

Detail example.

yyyc514:
Thoughts?

yyyc514:
Idea for adding comment:

```
What makes you feel the Github integration is broken?  This was fixed in #139.

## also: [CLOSE ISSUE]

## Commenting on: #139 Github Integration broken? [question]
## To save your comment press SUPER-ENTER. To cancel just close the window. 

---------------------------------------
* yyyc514 commented 33 minutes ago:
 
  Thoughts?


---------------------------------------
* yyyc514 commented 2 hours ago:
  
I was imagining a detailed view and then less detailed view that you could toggle.  Idea for less detail:


``
GITHUB: divmain/GitSavvy

#139   Github Integration broken? [question]
#137   How to tell GitSavvy to forget git repository? [ux]
[...]

PR #143, opened 2015-03-27
first pass at opening issue in browser

yyyc514:


yyyc514:
screen shot 2015-03-26 at 10 17 37 pm

First pass at this.  Currently only loads a single page of issues (same as the issue fetch for commits).  Need some feedback on how to choose a remote.  I changed last=True so it works for GitSavvy... but really there are two big use cases I can think of:

- remote with issues is origin (you're a direct contributor)
- remote with issues is upstream or some other (you're an indirect contributor via github)

I think we need a nice way to handle both without having to make a choice each time.

yyyc514:
Related #14

yyyc514:
And this doesn't support timezones properly.  Dates from github are UTC I'm guessing?  Not sure.

divmain:
Remote-selection should be drastically simplified as of 93cf346302f7c2baad87604fbe2b80048acd2a81, so take a look there.  I'm not sure what timezone GitHub uses in their API responses, haven't had to deal with that yet.  Is there a timezone included in the timestamps themselves, or as header information somewhere in the payload?

With `git: log` quick panel, we have paging set at something like 5000 commits.  You could do something similar here, where you add an entry at the end to page forward to the next commits.  It isn't perfect, but it is still a big improvement on opening browser, clicking issues, searching for the right issue.

yyyc514:
> Remote-selection should be drastically simplified as of 93cf346, so take a look there.

Yep, I'll go back and refactor this with your recent changes in mind.

> Is there a timezone included in the timestamps themselves, or as header information somewhere in the payload?

Timestamps appear to be UTC I think... I'll have to look again for some type of header - I didn't see one.

> With git: log quick panel, we have paging set at something like 5000 commits. You could do something similar here, where you add an entry at the end to page forward to the next commits.

It's 6000. Funny number.  Ok, that makes sense as a way to get this workable and out the door.  I'll see about doing that.

yyyc514:
Thoughts?

core/git_command.py
```diff
@@ -31,6 +31,22 @@ class GitSavvyError(Exception):
...
 
+class IsolatedGitCommand(GitCommand):
+    """
+    Allow accessing GitCommand outside the scope of a WindowCommand or
+    TextCommand.  You must manually pass in a valid repository path.
+    """
+    def __init__(self, repo_path=None):
+        self._repo_path = repo_path
+
+    def git(self, *args):
```
divmain: This method does nothing.

core/git_command.py
```diff
@@ -31,6 +31,22 @@ class GitSavvyError(Exception):
     pass
 
 
+class IsolatedGitCommand(GitCommand):
+    """
+    Allow accessing GitCommand outside the scope of a WindowCommand or
+    TextCommand.  You must manually pass in a valid repository path.
+    """
+    def __init__(self, repo_path=None):
+        self._repo_path = repo_path
```
divmain: You can just say `self.repo_path = repo_path` and remove the `repo_path` property method.

core/git_command.py
```diff
@@ -31,6 +31,22 @@ class GitSavvyError(Exception):
     pass
 
 
+class IsolatedGitCommand(GitCommand):
```
divmain: Not sure this is the abstraction I'd like to go with, so will think on this a bit before moving forward.

github/github.py
```diff
@@ -80,6 +93,11 @@ def get_api_fqdn(github_repo):
 
 
 def query_github(api_url_template, github_repo):
+    resp = raw_query_github(api_url_template, github_repo)
+    return resp.payload
+
+
+def raw_query_github(api_url_template, github_repo):
```
divmain: This is an unnecessary change.

divmain:
As I mentioned, not sure `IsolatedGitCommand` is how I'd like this to work.  I'll think on this a bit more and get back to you.  Also, please add docstrings all around.

github/github.py
yyyc514: Do you want the raw response returned everywhere and all code that uses web requests to be modified then?

core/git_command.py
yyyc514: Yeah it's a little different, but I needed git outside of a command/window so I didn't need the rest of the baggage... really what should be abstracted is running commands in general then git on top of that, then GitCommand as used by window/view.

core/git_command.py
yyyc514: I thought it was allowing me to call git with fewer options but now seeing that the calls are both *args that doesn't seem to be the case.

core/git_command.py
yyyc514: Are you sure?  Because this is a subclass of GitCommand I needed to override it's behavior for retrieving repo_path... does just setting a static value kill the property method from the parent class?

yyyc514:
>  Also, please add docstrings all around.

Where specifically?  In reviewing a git diff not much was added really.  I added a doctoring to the UTC method... but most new things have them, unless I'm missing something?

divmain:
I closed the ticket but failed to close this PR.  From the issue:

> Have decided this is out of scope for GitSavvy. Could be implemented as a GitSavvy plugin. In my case, I've discovered http://overv.io for issue management, and I doubt anything I put in GitSavvy (aside from the light integration that already exists) could provide a better issue management experience.

Issue #139, opened 2015-03-25
Github Integration broken?


wldcordeiro:
I noticed that I'm getting errors for issue and contributor completion regarding my credentials. I've updated my preferences to include the API token but can't get it to run.

divmain:
Hmm, it is working for me for both GitHub.com and GH Enterprise.  Can you paste the exact error output @wldcordeiro?

wldcordeiro:
```Traceback (most recent call last):
  File "/home/wldcordeiro/.config/sublime-text-3/Packages/GitSavvy/github/commands/commit.py", line 76, in <lambda>
    sublime.set_timeout_async(lambda: self.run_async())
  File "/home/wldcordeiro/.config/sublime-text-3/Packages/GitSavvy/github/commands/commit.py", line 82, in run_async
    contributors = github.get_contributors(remote)
  File "/home/wldcordeiro/.config/sublime-text-3/Packages/GitSavvy/github/github.py", line 99, in query_github
    raise FailedGithubRequest(response.payload)
GitSavvy.github.github.FailedGithubRequest: {'message': 'Bad credentials', 'documentation_url': 'https://developer.github.com/v3'}
Traceback (most recent call last):
  File "/home/wldcordeiro/.config/sublime-text-3/Packages/GitSavvy/github/commands/commit.py", line 76, in <lambda>
    sublime.set_timeout_async(lambda: self.run_async())
  File "/home/wldcordeiro/.config/sublime-text-3/Packages/GitSavvy/github/commands/commit.py", line 82, in run_async
    contributors = github.get_contributors(remote)
  File "/home/wldcordeiro/.config/sublime-text-3/Packages/GitSavvy/github/github.py", line 99, in query_github
    raise FailedGithubRequest(response.payload)
GitSavvy.github.github.FailedGithubRequest: {'message': 'Bad credentials', 'documentation_url': 'https://developer.github.com/v3'}```

**Comment by divmain** [comment]
Do you have an entry in your `GitSavvy.sublime-settings` that looks like this?
```
{
    // ...
    "api_tokens": {
        "github.com": "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
    }
}

```

The token itself should be 40-characters.

wldcordeiro:
```json
{
  "api_tokens": {
      "github.com": "apikeyhere"
  },
  "git_graph_args": [
      "log",
      "--oneline",
      "--decorate",
      "--all",
[...]

PR #145, opened 2015-03-27
fix: add # to regex to match text before cursor

yyyc514:


divmain:
Thanks for this.  It is unrelated to #139, but it needs to be fixed anyway.

PR #148, opened 2015-03-27
feature: improve quick stage

yyyc514:
- better status display "x files staged"
    - add "git: commit" to allow quick staging to become quick commiting

### DIFF
--- a/core/commands/quick_stage.py
+++ b/core/commands/quick_stage.py
@@ -12,6 +12,8 @@ MenuOption = namedtuple("MenuOption", ["requires_action", "menu_text", "filename
 CLEAN_WORKING_DIR = "Nothing to commit, working directory clean."
 ADD_ALL_UNSTAGED_FILES = " ?  All unstaged files"
 ADD_ALL_FILES = " +  All files"
+STAGED = "--- {} files are staged for commit. ---"
+COMMIT = "    git: commit"
 
 
 class GsQuickStageCommand(WindowCommand, GitCommand):
@@ -48,6 +50,10 @@ class GsQuickStageCommand(WindowCommand, GitCommand):
             if not selection.requires_action:
                 return
 
+            elif selection.menu_text == COMMIT:
+                self.window.run_command("gs_commit")
+                return
+
             elif selection.menu_text == ADD_ALL_UNSTAGED_FILES:
                 self.git("add", "--update", ".")
                 scope_of_action = "all unstaged files"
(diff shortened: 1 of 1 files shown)


yyyc514:
screen shot 2015-03-27 at 4 06 39 am

yyyc514:
I can fix the pluralization also if you care.

divmain:
@yyyc514, I sent you an email last night.  Can you please check and let me know if you haven't received?

yyyc514:
I just responded, it's not an email address I check all the time.

core/commands/quick_stage.py
```diff
@@ -12,7 +12,8 @@
 CLEAN_WORKING_DIR = "Nothing to commit, working directory clean."
 ADD_ALL_UNSTAGED_FILES = " ?  All unstaged files"
 ADD_ALL_FILES = " +  All files"
-
+STAGED = "--- {} files are staged for commit. ---"
+COMMIT = "    git: commit"
```
divmain: Nit: Add an extra line here.  PEP8 calls for two lines in-between top-level class definitions.

core/commands/status_bar.py
```diff
@@ -6,17 +6,20 @@
 
 class GsStatusBarEventListener(EventListener):
 
-    def on_new(self, view):
+    def trigger(self):
```
divmain: This change/method is not necessary.  Please remove.

core/commands/quick_stage.py
```diff
@@ -83,18 +87,29 @@ def get_menu_options(self):
...
                 menu_text = "[{0}] {1}".format(entry.working_status, filename)
                 menu_options.append(MenuOption(True, menu_text, filename, entry.index_status == "?"))
 
-        if not menu_options:
-            return [MenuOption(False, CLEAN_WORKING_DIR, None, None)]
+        if not status_entries:
+            menu_options.append(MenuOption(False, CLEAN_WORKING_DIR, None, None))
+            return menu_options
+
+        if unstaged_count>0:
```
divmain: Spaces around `>`.

core/commands/quick_stage.py
yyyc514: I noticed, oversight.  Anything like rubocop for Python?

core/commands/quick_stage.py
```diff
@@ -83,18 +87,29 @@ def get_menu_options(self):
...
                 filename = (entry.path if not entry.index_status == "R"
                             else entry.path + " <- " + entry.path_alt)
                 menu_text = "[{0}] {1}".format(entry.working_status, filename)
                 menu_options.append(MenuOption(True, menu_text, filename, entry.index_status == "?"))
 
-        if not menu_options:
-            return [MenuOption(False, CLEAN_WORKING_DIR, None, None)]
+        if not status_entries:
+            menu_options.append(MenuOption(False, CLEAN_WORKING_DIR, None, None))
+            return menu_options
```
divmain: The change to the `if` line is okay, but change these two lines back to:
```python
return [MenuOption(False, CLEAN_WORKING_DIR, None, None)]
```
And actually, please move this block to the top of the method to return early, immediately after the definition of `status_entries`.

core/commands/quick_stage.py
```diff
@@ -48,6 +49,9 @@ def on_selection(id):
             if not selection.requires_action:
                 return
 
+            elif selection.menu_text == COMMIT:
+                self.window.run_command("gs_commit")
+
```
divmain: Add line with `return` here.

core/commands/quick_stage.py
divmain: I use SublimeLinter and the [PEP8](https://github.com/SublimeLinter/SublimeLinter-pep8) plugin.  I think you can also just `pip3 install pep8` and then run `pep8 FILENAME` from the terminal.

divmain:
I like this change!  Comments are pretty minor, once they're done I'll merge.

yyyc514:
> I like this change!

I know. See... I'm not as crazy as you think. :-)

core/commands/quick_stage.py
yyyc514: Do you have a list of which errors you care about vs ignore so I can configure this a little tighter?  Right now I'm getting flags for lines longer than 79, which seems to be  pretty common issue in this source. :)

PR #154, opened 2015-03-28
fix: regex for matching blame dividers is broken

yyyc514:
- should be "|" no | (or)


divmain:
Funny!  This worked because it was matching ` -----------------`, which worked out alright in every case I tested it.  Good catch.

yyyc514:
Yes, in it's broken form it matches too much - so it will work in a lot of cases.

PR #171, opened 2015-04-17
Add an option to append customized message to the default commit help.

old9:
So that one can display a commit rule/tip/convention to the team when they commit, something like this:
image


divmain:
Looks good to me.  Thanks!

PR #175, opened 2015-05-01
fix: BranchesMixin regex doesn't match branch names with dots

ddevlin:
Existing regex didn't match branch names like 'v1.1' or 'release/v1.1'. If you checkout or create a branch with `.` in the name, it will display correctly as the active branch, but won't appear in the list of local or remote branches:

screen shot 2015-05-01 at 16 48 40


divmain:
Looks good. Thanks!

PR #182, opened 2015-05-12
fix: ActiveBranchMixin doesn't match status when remote branch is gone

ddevlin:
## Issue

Attempting to checkout a local tracking branch when the remote branch is gone/deleted causes the following exception in `core.get_mixins.ActiveBranchMixin`:

```
  File "/Users/ddevlin/Library/Application Support/Sublime Text 3/Packages/GitSavvy/core/interfaces/branch.py", line 73, in render_branch_status
    return self.get_branch_status(delim="\n           ")
  File "/Users/ddevlin/Library/Application Support/Sublime Text 3/Packages/GitSavvy/core/git_mixins/active_branch.py", line 69, in get_branch_status
    self._get_branch_status_components()
  File "/Users/ddevlin/Library/Application Support/Sublime Text 3/Packages/GitSavvy/core/git_mixins/active_branch.py", line 48, in _get_branch_status_components
    return False, False, addl_lines[0], None, clean, None, None
IndexError: list index out of range
```

## Resolution

- Updated `short_status_pattern` to match status when remote branch is gone. 
- Pass back additional boolean from `_get_branch_status_components` to indicate if remote branch is gone. 
- Updated `get_branch_status` with new status message to indicate remote branch is gone:
screen shot 2015-05-12 at 15 38 07

### DIFF
--- a/core/git_mixins/active_branch.py
+++ b/core/git_mixins/active_branch.py
@@ -26,6 +26,7 @@ class ActiveBranchMixin():
           4) boolean indicating whether branch is clean
           5) # commits ahead of remote
           6) # commits behind of remote
+          7) boolean indicating whether the remote branch is gone
         """
         stdout = self.git("status", "-b", "--porcelain").strip()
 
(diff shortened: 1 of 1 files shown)


core/git_mixins/active_branch.py
```diff
@@ -86,6 +87,8 @@ def get_branch_status(self, delim=None):
                 secondary = "You're ahead by {}.".format(ahead)
             elif behind:
                 secondary = "You're behind by {}.".format(behind)
+            elif gone:
+                secondary = "The remote branch is gone."
```
ddevlin: @divmain Does this message look ok to you? For consistancy with ahead/behind, I could highlight "gone" via `branches.tmLanguage`, maybe something like this:
```xml
<dict>
	<key>comment</key>
	<string>ahead, behind or gone</string>
	<key>match</key>
	<string>(ahead)|(behind)|(gone)</string>
	<key>name</key>
	<string>keyword.other.git-savvy.status.ahead-behind</string>
</dict>
```

core/git_mixins/active_branch.py
```diff
@@ -34,22 +35,22 @@ def _get_branch_status_components(self):
...
 
         valid_punctuation = "".join(c for c in string.punctuation if c not in "~^:?*[\\")
         branch_pattern = "[A-Za-z0-9" + re.escape(valid_punctuation) + "]+?"
-        short_status_pattern = "## (" + branch_pattern + ")(\.\.\.(" + branch_pattern + ")( \[((ahead (\d+))(, )?)?(behind (\d+))?\])?)?$"
+        short_status_pattern = "## (" + branch_pattern + ")(\.\.\.(" + branch_pattern + ")( \[((ahead (\d+))(, )?)?(behind (\d+))?(gone)?\])?)?$"
         status_match = re.match(short_status_pattern, first_line)
 
         if not status_match:
-            return False, False, addl_lines[0], None, clean, None, None
+            return False, False, addl_lines[0], None, clean, None, None, False
```
ddevlin: Might be better to replace `addl_lines[0]` with `None if clean else addl_lines[0]` to prevent exceptions if there are additional cases where `short_status_pattern` doesn't match `first_line` but `addl_lines` is empty?

core/git_mixins/active_branch.py
divmain: Highlighting is unnecessary, I think.  Message is concise, so :+1:

core/git_mixins/active_branch.py
divmain: Unfortunately, I can't recall the case where `short_status_pattern` doesn't match - I should've left a comment for myself there!  It may be unnecessary, but it can't hurt, so go ahead and make that change.  Thanks!

divmain:
Just the one change, and this will be ready to merge.  Thanks again!

ddevlin:
@divmain Change committed and commits squashed.

divmain:
Awesome, thanks!

PR #185, opened 2015-05-19
fix: get_branch_status_short should unpack gone value

ddevlin:
In my previous commit 6b5fd98c `_get_branch_status_components` was updated to return a seventh value (boolean indicating if the remote branch was gone), but `get_branch_status_short` wasn't updated to unpack this value.

Sorry, I really should have checked for additional calls to `get_branch_status_short`!


divmain:
:)

PR #186, opened 2015-05-19
feature: per-project default rebase base ref

ddevlin:
Some branching models (e.g. [gitflow](http://nvie.com/posts/a-successful-git-branching-model/)) use something other than `master` as an integration branch. In my case, I'm integrating against a branch called `dev` and only merging to `master` when I'm doing a stable release. This means that when I open the rebase dashboard I always need to switch base ref.

As the appropriate default base ref varies from repository to repository, I think a per-project setting in a project's `.sublime-project` file is the best place to override the default base ref. In the absence of this setting, or an active `.sublime-project`, I've retained the default behaviour of defaulting to `master`

### DIFF
--- a/core/interfaces/rebase.py
+++ b/core/interfaces/rebase.py
@@ -212,8 +212,9 @@ class RebaseInterface(ui.Interface, GitCommand):
     def base_ref(self):
         base_ref = self.view.settings().get("git_savvy.rebase.base_ref")
         if not base_ref:
-            base_ref = "master"
-            self.view.settings().set("git_savvy.revase.base_ref", "master")
+            project_settings = sublime.active_window().project_data().get('settings', {})
+            base_ref = project_settings.get("rebase_default_base_ref", "master")
+            self.view.settings().set("git_savvy.rebase.base_ref", base_ref)
         return base_ref
 
     def base_commit(self):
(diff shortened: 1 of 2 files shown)


core/interfaces/rebase.py
```diff
@@ -212,8 +212,9 @@ def render_conflicts_bindings(self):
     def base_ref(self):
         base_ref = self.view.settings().get("git_savvy.rebase.base_ref")
         if not base_ref:
-            base_ref = "master"
-            self.view.settings().set("git_savvy.revase.base_ref", "master")
```
ddevlin: There was a typo here that I corrected (revase => rebase), but I don't think was actually causing any issues.

core/interfaces/rebase.py
```diff
@@ -212,8 +212,9 @@ def render_conflicts_bindings(self):
...
+            project_settings = sublime.active_window().project_data().get('settings', {})
+            base_ref = project_settings.get("rebase_default_base_ref", "master")
```
ddevlin: @divmain Does this settings key (`rebase_default_base_ref`) look ok to you?

docs/rebase.md
```diff
@@ -62,7 +62,20 @@ This will display the selected commit data and meta-data in a new window, includ
...
+
+```json
+{
+    "folders": [
+        {
+           "path": "XYZ"
+        }
+    ],
+    "settings": {
+        "rebase_default_base_ref": "develop"
```
max-mykhailenko: Maybe 
```
"settings": {
 gitsavvy : {"rebase_default_base_ref": "develop"}
}
```
will be better?

divmain:
This is a great change!

PR #191, opened 2015-06-02
Fix: super_key should be set to CTRL on Linux

### LINKED ISSUE #190: Wrong help in commit's view
As I can see from [this file](https://github.com/divmain/GitSavvy/blob/master/Default%20(Linux).sublime-keymap) I should use `Ctrl`+`Enter` to make a commit, but on the commit's view page I see a following message:
```
## To make a commit, type your commit message and press SUPER-ENTER. To cancel
## the commit, close the window. To sign off the commit press SUPER-S.
```

ddevlin:
Since 983cf67 Linux key map has used `CTRL` instead of `SUPER`, but `common.util.super_key` was still set to `SUPER` on Linux as `os.name == 'posix'` is `true` on both Linux and OS X. Condition changed to `sys.platform == 'darwin'` which is only true on OS X.

Closes #190.

### DIFF
--- a/common/util/__init__.py
+++ b/common/util/__init__.py
@@ -1,4 +1,4 @@
-import os
+import sys
 
 from .parse_diff import parse_diff
 from . import dates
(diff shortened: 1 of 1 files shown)


stoivo:
Looks fine

divmain:
Excellent, thanks for this fix.

PR #192, opened 2015-06-03
Feature: Display diffs against commit history from `git: log`

### LINKED ISSUE #172: Diff current file to earlier commits
Often use in console diff file between current state and some commit.

ddevlin:
Preview

When the `show_log_output_options` setting is set to `true`, on selection of a commit from the `git: log` quick panel, a user can choose from the following options:

`Show commit`: Display the commit
`Show diff`: Display the diff between the commit and the working tree
`Show diff (cached)`: Display the diff between the commit and staged changes

Closes #172

### DIFF
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -28,7 +28,7 @@ class GsDiffCommand(WindowCommand, GitCommand):
     def run(self, **kwargs):
         sublime.set_timeout_async(lambda: self.run_async(**kwargs), 0)
 
-    def run_async(self, in_cached_mode=False, file_path=None, current_file=False):
+    def run_async(self, in_cached_mode=False, file_path=None, current_file=False, base_commit=None):
         repo_path = self.repo_path
         if current_file:
             file_path = self.file_path or file_path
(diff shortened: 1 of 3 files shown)


GitSavvy.sublime-settings
```diff
@@ -125,5 +125,12 @@
...
         NOTE: Changes you make may break in-view functionality, so use with caution.
      */
-    "git_graph_args": ["log", "--oneline", "--graph", "--all", "--decorate"]
+    "git_graph_args": ["log", "--oneline", "--graph", "--all", "--decorate"],
+
+    /*
+        Change this to `true` to pick the output type (e.g. "Show commit" or "Show diff") when 
+        selecting a commit from `git: log`.
+    */
+    "show_log_output_options": false
```
ddevlin: @divmain I set this to false to preserve the current behaviour, but if you would prefer to make the new behaviour the default I can change this to `true`.

divmain:
Hmm... I really like this.  My gut reaction was to dislike the extra step, but seeing as "Show commit" is the default selection, an extra `Enter` keypress is pretty negligible.  I'd say remove the configuration option altogether, and make this the only behavior (making sure to preserve "Show commit" as default action).

Only other thing I'd like to see changed here is the selection text for the second and third options.  `Show diff` doesn't tell me enough, as the user, what it is I'll be looking at.  I'd have to consult the documentation.  Let's change this to `Compare commit against working directory` and `Compare commit against index`.

ddevlin:
@divmain Sure, it's nicer not to introduct an extra setting anyway, I'm just hesitant to change existing behaviour in unsolicited pull requests ;) 

Setting removed, selection text updated, docs updated, and commits squished.

divmain:
Looks great, thanks @ddevlin!

PR #193, opened 2015-06-03
Diff commit on branch

stoivo:
I created `f + m` in branch interface. This displays a graph of commits that on one of the branches but not  on the active. It have all the same functions as graph view have. Not sure if that is good or bad.

it looks like this: 
screen shot 2015-06-03 at 22 57 43

This is not done completed but I would like to get you feedback.

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -478,6 +478,13 @@
         ]
     },
     {
+        "keys": ["H"],
+        "command": "gs_branches_diff_branch_history",
+        "context": [
+            { "key": "setting.git_savvy.branch_view", "operator": "equal", "operand": true }
+        ]
+    },
+    {
         "keys": ["r"],
         "command": "gs_branches_refresh",
         "context": [
--- a/core/interfaces/branch.py
+++ b/core/interfaces/branch.py
@@ -52,6 +52,7 @@ class BranchInterface(ui.Interface, GitCommand):
       [o] checkout remote as local                  [h] fetch remote branches
 
       [f] diff against active
+      [H] diff history against active
       [e] toggle display of remote branches
       [r] refresh
 
@@ -598,9 +599,12 @@ class GsBranchesToggleRemotesCommand(TextCommand, GitCommand):
     Toggle display of the remote branches.
     """
 
-    def run(self, edit):
+    def run(self, edit, show = None):
         interface = ui.get_interface(self.view.id())
-        interface.show_remotes = not interface.show_remotes
+        if show == None:
+            interface.show_remotes = not interface.show_remotes
+        else:
+            interface.show_remotes = show
         interface.render()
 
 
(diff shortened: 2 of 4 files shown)


docs/branch_mgmt.md
```diff
@@ -82,10 +82,15 @@ The selected branch will be merged into the active branch.  If the merge cannot
...
 
-#### Diff selected branch against active (`f`)
+#### Diff code selected branch against active (`f + d`)
 
 A scratch view will be opened, showing the diff between the selected branch and the active branch.
 
+
+#### Diff commits selected branch against active (`f + m`)
+
+A scratch view will be opened, showing the commit diff between the selected branch and the active branch.
```
divmain:
```
A scratch view will be opened, showing all commits present on the selected branch but not the active branch, and vice versa.
```

Default.sublime-keymap
```diff
@@ -471,13 +471,20 @@
         ]
     },
     {
-        "keys": ["f"],
+        "keys": ["f", "d"],
```
divmain: No need to have two-key bindings here.  Change this to `"keys": ["H"]`.

docs/branch_mgmt.md
```diff
@@ -82,10 +82,15 @@ The selected branch will be merged into the active branch.  If the merge cannot
 
 This action also merged the selected branch into active.  However, it only works for remote branches, and will first fetch from the selected remote before the branch is merged.
 
-#### Diff selected branch against active (`f`)
+#### Diff code selected branch against active (`f + d`)
 
 A scratch view will be opened, showing the diff between the selected branch and the active branch.
 
+
+#### Diff commits selected branch against active (`f + m`)
```
divmain:
```
#### Diff history of selected and active branches (`H`)
```

core/interfaces/branch.py
```diff
@@ -612,3 +612,60 @@ class GsBranchesFetchCommand(TextCommand, GitCommand):
 
     def run(self, edit):
         self.view.window().run_command("gs_fetch")
+
+
+class GsBranchesDiffCommitsBranch(TextCommand, GitCommand):
```
divmain: `GsBranchesDiffBranchHistory`

core/interfaces/branch.py
```diff
@@ -612,3 +612,60 @@ class GsBranchesFetchCommand(TextCommand, GitCommand):
...
+
+        repo_path = self.repo_path
+        view = self.view.window().new_file()
+        view.set_syntax_file("Packages/GitSavvy/syntax/graph.tmLanguage")
+        view.settings().set("git_savvy.log_graph_view", True)
+        view.settings().set("git_savvy.repo_path", repo_path)
+        view.settings().set("word_wrap", False)
+        view.set_name("BRANCH COMMIT COMPARISON")
+        view.set_scratch(True)
+        view.run_command("gs_replace_view_text", {"text": diff_contents, "nuke_cursors": True})
```
divmain: This can be cleaned up with `util.get_read_only_view`.

core/interfaces/branch.py
```diff
@@ -612,3 +612,60 @@ class GsBranchesFetchCommand(TextCommand, GitCommand):
...
+        savvy_settings = sublime.load_settings("GitSavvy.sublime-settings")
+        args = savvy_settings.get("git_graph_args")
+        diff_contents = "Commits on {} and not on {}\n".format(branch1, branch2)
+        args.append("{}..{}".format(branch2, branch1))
+        diff_contents += self.git(*args)
+        diff_contents += "\n\nCommits on {} and not on {}\n".format(branch2, branch1)
+        args.pop()
+        args.append("{}..{}".format(branch1, branch2))
+        diff_contents += self.git(*args)
+        return diff_contents
```
divmain: Since this is something new, it may be best to add this functionality as a new interface, rather than the old-style view instantiation / view logic.  The branch-dashboard command would only call the interface constructor with the selected branch name.

core/interfaces/branch.py
```diff
@@ -612,3 +612,60 @@ class GsBranchesFetchCommand(TextCommand, GitCommand):
...
+        view.set_scratch(True)
+        view.run_command("gs_replace_view_text", {"text": diff_contents, "nuke_cursors": True})
+    def get_commit_branch_sting(self, branch1, branch2):
+        savvy_settings = sublime.load_settings("GitSavvy.sublime-settings")
+        args = savvy_settings.get("git_graph_args")
+        diff_contents = "Commits on {} and not on {}\n".format(branch1, branch2)
+        args.append("{}..{}".format(branch2, branch1))
+        diff_contents += self.git(*args)
+        diff_contents += "\n\nCommits on {} and not on {}\n".format(branch2, branch1)
+        args.pop()
```
divmain: This is a little too clever.  Should be abstracted into a method that takes two args, `branchA`, and `branchB`, and returns the contents from that combination.

divmain:
This is a worthy feature!  Let me know once you've addressed my comments.  I may adapt what you've written here into an interface, which is where most of these things have been heading.

Default.sublime-keymap
stoivo: Fine, is there any logic for that? History?
My question is why 'H'?

core/interfaces/branch.py
stoivo: maybe it would be good to be able to pass a hash of option that we can use to set settings like 
``` pyhon3
view.settings().set("word_wrap", False)
```

Default.sublime-keymap
divmain: Yeah, I was thinking "diff history" as the short-description.  Generally, unless grouping very-related commands together (like stash commands), I try to stick to single key-presses.  This new command relates to the diff command in that it sort of does something similar.  Sticking to single key-bindings (when possible) also avoids the issue of mistakenly invoking the wrong command when the first key wasn't registered.

core/interfaces/branch.py
divmain: Most of these read-only views behave in similar ways.  The `repo_path` will always be passed, I can't think of a time when you'd want `word_wrap` to be `True`, etc.  So unless there's a reason, I'd prefer for the helper method to set sane defaults and only make exceptions when necessary.

Default.sublime-keymap
stoivo: Make sense.

divmain:
Also, once you've made the requested changes, could you squash and rebase on top of latest master?  Looks like there is at least one conflict.

core/interfaces/branch.py
stoivo: We do want to use `word_wrap` on committing or looking at a commit. Like from the graph.

stoivo:
All I need to do is to Convert it to a interface. Is there anything more I should change or add to it?

divmain:
This looks good.  However, I'd like you to remove the `GsMoveCorsurToRemoteBranch` command and associated keybinding.  That is definitely a different feature, so should go in a separate PR.  And I have other plans for how navigation between sections should work, so this likely conflicts.

stoivo:
@divmain, now I removed GsMoveCorsurToRemoteBranch and rebased once more :)
Should I change it to a interface?

I whould like to have GsMoveCorsurToRemoteBranch somehow. Can I get  to know your plan?

divmain:
Thanks for this.  Merging now.  Can you ping me on Gitter to discuss `GsMoveCorsurToRemoteBranch`?  I need a refresher on what it did so that I can properly reply to your question.

PR #197, opened 2015-06-10
Fix: Git fetch with remotes=None should append --all option

ddevlin:
The fetch `All remotes.` option currently does not fetch all remotes (watch the upstream remote below):

The GitSavvy `RemotesMixin.fetch` function expects git to fetch all remotes if a remote isn't specified, but according to the git docs it will default to `origin`:

```
When no remote is specified, by default the origin remote will be used,
unless there’s an upstream branch configured for the current branch.
```
http://git-scm.com/docs/git-fetch/2.4.0

To fix I've appended the `--all` option if remote is falsey.

### DIFF
--- a/core/git_mixins/remotes.py
+++ b/core/git_mixins/remotes.py
@@ -17,7 +17,7 @@ class RemotesMixin():
         If provided, fetch all changes from `remote`.  Otherwise, fetch
         changes from all remotes.
         """
-        self.git("fetch", "--prune" if prune else None, remote)
+        self.git("fetch", "--prune" if prune else None, remote if remote else "--all")
 
     def get_remote_branches(self):
         """


divmain:
Damn, oversight on my part.  Thanks, I'll merge this and push it out.

stoivo:
What application do you use to record screen and key-press-viewer?
It is nice to look at and eacy to understand what the thing you want to show.

divmain:
@stoivo, on a Mac, I use [RecordIt](http://recordit.co/).

stoivo:
:+1:

ddevlin:
@stoivo I use [Screenflick](http://www.araelium.com/screenflick) to record and [GIFBrewery](http://gifbrewery.com) to convert to animated GIF, both for Mac.

stoivo:
cool

PR #201, opened 2015-06-11
Feature: git reset, using log or reflog

### LINKED ISSUE #200: git reset commands
We don't have reset commands.

ddevlin:
I've made a first attempt at a git reset feature (closes #200). Feedback welcomed :smile: 

Adds two git reset commands to the command panel:
- `git: reset`: Displays the commit selector from `git: log`, selecting a commit performs a `git reset --mixed` to that commit.
- `git: reset (reflog)` Displays a reference log selector, selecting a reference performs a `git reset --mixed` to the referenced commit. This allows you to reset an earlier reset or a rebase.

Reflog quick panel:
screen shot 2015-06-11 at 23 43 51

The reflog option could be confusing to less advanced git users. I personally would find it useful, on occasion I've reset one or two commits too far, and it would be nice if I could quickly undo that without returning to the terminal. If it's more trouble than it's worth though, I'm happy to take it out.

Doc entries are pretty rough (particularly for the `reflog` option), so welcome feedback there. Wasn't sure whether to put them in `misc.md` or elsewhere.

I'm using the default git reset behaviour `--mixed` (though I've added it to command anyway just to be explicit). I didn't add an option for `--soft` or `--hard` to keep things simple, from the status view it's easy to either stage or discard the post-reset uncommited changes anyway.

@divmain In `GsResetCommand` I subclassed `GsLogCommand` as the only difference between the two is the `on_hash_selection` method, but it might be better not to subclass and live with the duplication, as the subclassing is less explicit and future changes to `GsLogCommand` could potentially break `GsResetCommand`. What do you think?

### DIFF
--- a/docs/misc.md
+++ b/docs/misc.md
@@ -7,3 +7,13 @@ This command will initialize a new repository, or re-initialize a pre-existing o
 When run, you will asked to confirm the root directory of the new Git repository.  GitSavvy will attempt to auto-detect this for you.
 
 This command will also be suggested to you should you attempt to run a GitSavvy command on a file that is not within a valid Git repository.
+
+## `git: reset`
+
+This command will change the HEAD of the current branch to a previous commit.
+
+When run, you will be asked to select a commit from a list of previous commits. Once you select a commit, the HEAD of the current branch will be changed to the selected commit, and any changes that were committed after the selected commit will appear as unstaged changes in the working directory.
+
+## `git: reset (reflog)`
+
+Like `git: reset`, this command will change the HEAD of the current branch to a previous commit, but uses `git reflog` rather than `git log` as the source of available commits.
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -207,5 +207,13 @@
     {
         "caption": "git: rebase",
         "command": "gs_show_rebase"
+    },
+    {
+        "caption": "git: reset",
+        "command": "gs_reset"
+    },
+    {
+        "caption": "git: reset (reflog)",
+        "command": "gs_reset_reflog"
     }
 ]
--- a/core/commands/__init__.py
+++ b/core/commands/__init__.py
@@ -17,3 +17,4 @@ from .merge import*
 from .changelog import *
 from .status_bar import *
 from .tags import *
+from .reset import *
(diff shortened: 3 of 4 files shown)


stoivo:
I am not super good with git, soo git reflog is the log of everything you do With reflog you can undo merges, undo resets, undo everything. Rigth??

ddevlin:
> soo git reflog is the log of everything you do

Almost, it's a log of every change to the commit that HEAD is pointing to. Actions that don't change HEAD, e.g. changes in your working directory or staging files, don't appear in the reflog.

> With reflog you can undo merges, undo resets, undo everything. Rigth??

Yes, but it only allows you to undo things that change the commit that HEAD is pointing too. So you could undo:
- amending a commit, 
- reset, 
- rebase, 
- cherry-pick, etc. 

but you couldn't undo deleting a file or discarding a change that had never been committed. It's most useful when you need to undo an action that has rewritten your git history. Hope that helps, this might explain it better than I can: https://www.atlassian.com/git/tutorials/refs-and-the-reflog/

docs/misc.md
```diff
@@ -7,3 +7,13 @@ This command will initialize a new repository, or re-initialize a pre-existing o
...
+
+## `git: reset`
+
+This command will change the HEAD of the current branch to a previous commit.
+
+When run, you will be asked to select a commit from a list of previous commits. Once you select a commit, the HEAD of the current branch will be changed to the selected commit, and any changes that were committed after the selected commit will appear as unstaged changes in the working directory.
+
+## `git: reset (reflog)`
+
+Like `git: reset`, this command will change the HEAD of the current branch to a previous commit, but uses `git reflog` rather than `git log` to commits.
```
divmain: ... but uses `git reflog` rather than `git log` as the source of available commits.

divmain:
Huge thumbs up on this!  Very much like the reflog option, and could see myself using it.  I think inheriting from `GsLogCommand` is fine.  If I mull it over and decide differently, it is likely this functionality will just be pulled out into a superclass that both inherit from.

Just the one copy change, and this will be ready to go.

ddevlin:
Updated and squashed.

divmain:
Excellent - thanks again.

PR #202, opened 2015-06-12
Fix: Stash shortcuts do not execute when cursor is at end of line

### LINKED ISSUE #196: Status dashboard - stash shortcuts inactive when cursor is at the end of a stash row
Moving the cursor 1 symbol to the left fixed it and I can now do `[t][s] show stash` again.

I realise this is very nitty gritty, but my OCD kicks in every time. Can add easy pickings tag?

ddevlin:
Matching regex for meta.git-savvy.status.saved_stash did not include end of line.

Closes #196.

### DIFF
--- a/syntax/status.YAML-tmLanguage
+++ b/syntax/status.YAML-tmLanguage
@@ -45,7 +45,7 @@ patterns:
   end: '^$'
   patterns:
   - name: meta.git-savvy.status.saved_stash
-    match: '^    (.+)$'
+    match: '^    (.+)\n$'
 
 - comment: Key-bindings menu
   name: comment.git-savvy.status.key-bindings-menu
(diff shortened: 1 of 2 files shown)


divmain:
Awesome, thanks for this!  Can you scour that syntax file for any other instances where this might be the case?

ddevlin:
I don't think there's anything else in the status dashboard syntax file, as far as I can see the only other pattern that was used as a key binding context is `meta.git-savvy.status.file` and you already included a `\n` in that pattern. 

There may be similar issues in the syntax files of other dashboards, I'll take a look when I get a chance.

divmain:
Fair enough - I'm going to merge this for now, and if you get a change to review the other key bindings, that'd be great.  Otherwise, I'll take a look myself in the next day or two.

PR #206, opened 2015-06-22
Prompt before discarding commit message

asfaltboy:
### Changes

* display prompt by default
* allow toggling behavior with "prompt_on_abort_commit" setting
* always set to scratch on commit (no need to prompt)

### Reasoning

Coming in from writing commit message with command line editors,
quitting the view mid-commit is always considered dirty in my mind.

And so I often found myself closing a commit by habit and losing
changes without recover options. 

I've considered different solutions for this problem but could not
find anything as simple that worked for me until now.

The same behavior is also used in by the popular plugin "sublime git"
which is where some users come from (including myself).

### WIP

This is a simple change, however if anyone wants to change anything
about this behavior or make it disabled by default, please let me know
and I'll amend the PR.

### DIFF
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -134,5 +134,10 @@
         NOTE: Changes you make may break in-view functionality, so use with caution.
      */
     // "git_graph_args": ["log", "--pretty=format:%h%d (%ar) by `%an` %s", "--graph"],
-    "git_graph_args": ["log", "--oneline", "--graph", "--decorate"]
+    "git_graph_args": ["log", "--oneline", "--graph", "--decorate"],
+
+    /*
+        Set to False to abort commit and drop message without a prompt
+    */
+    "prompt_on_abort_commit": false
 }
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -55,7 +55,8 @@ class GsCommitCommand(WindowCommand, GitCommand):
 
         title = COMMIT_TITLE.format(os.path.basename(repo_path))
         view.set_name(title)
-        view.set_scratch(True)
+        if not gitsavvy_settings.get("prompt_on_abort_commit"):
+            view.set_scratch(True)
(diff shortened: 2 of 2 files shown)


divmain:
Hmm, while this hasn't been an issue for me, I can see what you're saying.  Switch the default to `false` so that this is a non-breaking change, and I'll merge this in.

As an aside, if you'd like to make my life a slight bit easier (definitely not required), you could follow my commit message structure.  In case you've missed the `git: generate changelog` command, it will build a change-log for me by automatically categorizing commit messages that follow the following format:

```
TYPE: Short commit message

Detailed
commit
info
```

For the value of `TYPE`, I've been using `Feature`, `Enhancement`, and `Bug`.

asfaltboy:
Yeah I hadn't noticed that command, looks great; by the way can I customize the commit structure template in settings?

I'll commit the change to this PR when I get to a PC in a bit

GitSavvy.sublime-settings
```diff
@@ -134,5 +134,10 @@
         NOTE: Changes you make may break in-view functionality, so use with caution.
      */
     // "git_graph_args": ["log", "--pretty=format:%h%d (%ar) by `%an` %s", "--graph"],
-    "git_graph_args": ["log", "--oneline", "--graph", "--decorate"]
+    "git_graph_args": ["log", "--oneline", "--graph", "--decorate"],
+
+    /*
+        Set to False to abort commit and drop message without a prompt
```
divmain: This comment needs an update.

divmain:
There is no mechanism to control the commit message structure, wrt the changelog generation.  I'd be open to customization, though, if you'd like to take a crack at it.

GitSavvy.sublime-settings
divmain: Nvm, I can update the message separately.

asfaltboy: oops, aight thanks

PR #210, opened 2015-06-23
Fix: color output problems if user have set there config to always

stoivo:
it look like this
screen shot 2015-06-23 at 22 16 59
in .gitconfig
```
[color]
  ui = auto
  diff = always
  status = always
  branch = always
  interactive = always
```


divmain:
This looks reasonable.  Thanks!

PR #215, opened 2015-07-25
Bug: Branch matching regex did not match dots in remote branch name

ddevlin:
Fixed branch matching regex so it matches remote branch names containing dots.

**Before:**
screen shot 2015-07-25 at 21 06 49

**After:**
screen shot 2015-07-25 at 21 26 01

### DIFF
--- a/core/git_mixins/branches.py
+++ b/core/git_mixins/branches.py
@@ -30,7 +30,7 @@ class BranchesMixin():
         if not line:
             return None
 
-        pattern = r"(\* )?(remotes/)?([a-zA-Z0-9\-\_\/\.]+(?<!\.lock)(?<!\/)(?<!\.)\b) +([0-9a-f]{40}) (\[([a-zA-Z0-9\-\_\/\\]+)(: ([^\]]+))?\] )?(.*)"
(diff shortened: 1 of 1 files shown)


core/git_mixins/branches.py
```diff
@@ -30,7 +30,7 @@ def _parse_branch_line(line):
         if not line:
             return None
 
-        pattern = r"(\* )?(remotes/)?([a-zA-Z0-9\-\_\/\.]+(?<!\.lock)(?<!\/)(?<!\.)\b) +([0-9a-f]{40}) (\[([a-zA-Z0-9\-\_\/\\]+)(: ([^\]]+))?\] )?(.*)"
+        pattern = r"(\* )?(remotes/)?([a-zA-Z0-9\-\_\/\.]+(?<!\.lock)(?<!\/)(?<!\.)\b) +([0-9a-f]{40}) (\[([a-zA-Z0-9\-\_\/\.\\]+)(: ([^\]]+))?\] )?(.*)"
```
ddevlin: @divmain: I think the `\\` at the end of the remote matching part of the pattern (`([a-zA-Z0-9\-\_\/\.\\]+)`) might be a typo, as far as I know there shouldn't be any `\`s here? If it's a typo I can update the commit to remove it, just wanted to check with you first.

divmain: Good catch.  Yes, please do make that change.

ddevlin:
@divmain: Updated and squashed.

divmain:
Awesome.  Thanks!

PR #216, opened 2015-07-25
Enhancement: Hide git flow commands if git flow is not installed

ddevlin:
Git flow commands are only usable if git flow is installed, this patch hides them if it's not.


ddevlin:
This enhancement causes issues if the project is not already a git repository, so closing PR until I can resolve the issue

PR #219, opened 2015-07-28
Enhancement: Updated git-flow documentation

ddevlin:
Docs updated as `show_git_flow_commands` is now false by default.


divmain:
Thanks!

PR #221, opened 2015-07-30
Enhancement: Rebase dashboard handles case where master doesn't exist

### LINKED ISSUE #217: Rebase doesn't work if no master branch
I tried `git rebase` in a repository with no `master` (it only has `gh-pages` and forks from that):

```
`/usr/bin/git merge-base HEAD master` failed with following output:

fatal: Not a valid object name master
```

ddevlin:
Some repositories don't use a `master` branch (e.g. GitHub Pages repos use `gh-pages` instead). If a `base_ref` hasn't been actively selected and `rebase_default_base_ref` hasn't been set in the project settings the `base_ref` will default to `master`, and if `master` doesn't exist `merge-base` will throw an exception, causing the dashboard not to display.

This commit checks that the default `base_ref` (either `master` or `rebase_default_base_ref`) is a valid branch, and if it's not falls back to the first inactive local branch, the first local branch, or `HEAD`.

Closes #217 

**After**:

### DIFF
--- a/core/interfaces/rebase.py
+++ b/core/interfaces/rebase.py
@@ -211,10 +211,27 @@ class RebaseInterface(ui.Interface, GitCommand):
 
     def base_ref(self):
         base_ref = self.view.settings().get("git_savvy.rebase.base_ref")
+
         if not base_ref:
             project_settings = sublime.active_window().project_data().get('settings', {})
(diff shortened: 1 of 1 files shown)


divmain:
Fantastic.  Thanks @ddevlin!

PR #223, opened 2015-08-04
Allowing git_path to be customized on a per-platform basis.

johnburnett:
On Windows, git isn't always in PATH, requiring you to set git_path for GitSavvy to work correctly.  For users that sync settings across platforms, this allows git_path to be (optionally) customized on a per-platform basis.


divmain:
Looks good to me.  Thanks!

PR #224, opened 2015-08-14
Enhancement: Additional git blame display options

ddevlin:
This adds additional options to the git blame view. I occasionally "blame" the wrong person for a change, because they fixed indentation, moved a function within a file, or broke a large file up into smaller files. When using git blame from the command line I can use the `-w`, `-M` and `-C` flags to find the last real change to a line. This adds similar functionality to GitSavvy.


stoivo:
:+1:

divmain:
This looks great - thanks!

PR #229, opened 2015-08-26
Feature: simple background thread for custom commands

asfaltboy:
There's no documentation, I just wanted to pitch the idea first.

The custom commands work well for short duration commands; however, for long running commands the
plugin is blocked in the `subprocess` `communicate()` call (polling for stdout/err).

### DIFF
--- a/core/commands/custom.py
+++ b/core/commands/custom.py
@@ -1,4 +1,5 @@
 import sublime
+import threading
 from sublime_plugin import WindowCommand
 
 from ..git_command import GitCommand
@@ -6,6 +7,17 @@ from ...common import util
 
 
 ALL_REMOTES = "All remotes."
+
+
+class CustomCommandThread(threading.Thread):
+    def __init__(self, func, *args, **kwargs):
+        super(CustomCommandThread, self).__init__(**kwargs)
+        self.cmd_args = args
+        self.cmd_func = func
+        self.daemon = True
+
+    def run(self):
+        return self.cmd_func(*self.cmd_args)
 
 
 class GsCustomCommand(WindowCommand, GitCommand):
@@ -21,7 +33,8 @@ class GsCustomCommand(WindowCommand, GitCommand):
                   output_to_panel=False,
                   args=None,
                   start_msg="Starting custom command...",
-                  complete_msg="Completed custom command."):
+                  complete_msg="Completed custom command.",
+                  run_in_thread=False):
 
         if not args:
             sublime.error_message("Custom command must provide args.")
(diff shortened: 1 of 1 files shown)


divmain:
Hmm... are you noticing blocking behavior?  Commands should already be running in a separate thread (this is what `sublime.set_timeout_async` does).  All commands will run through that single, separate thread, but it should avoid any problems with non-responsive UI.

asfaltboy:
No blocking in UI, but as you said, that single thread, this one blocks.


For instance while running a `git gui` command, all of the other git commands appear to queue up, and suddenly deque when I close the GUI process.





Best regards,
Pavel Savchenko

divmain:
Hmm `git gui` is something of a special case.  In almost all other cases, we do not want to be running more than one `git` command concurrently, so queuing the tasks is appropriate.

It may be worth adding a generic custom command interface (run any command, not just Git commands), exposing `REPO_PATH` and `FILE_PATH` variables but otherwise just spinning up a child process and then forgetting about it.  This seems a better way forward to me than running all custom Git commands in their own thread.  That said, aside from exposing `REPO_PATH`, this may be better suited to a different plugin altogether.

asfaltboy:
This commit does not force new thread for all custom `git` commands, rather it adds an optional argument `run_in_thread` (False by default); which, when enabled, allows the creation of a new thread for this specific custom action.

Furthermore, from my observation, the `thread`, and the `subprocess` it spawns, are properly cleaned up after shutdown. The child also stays alive after sublime close, which is a bonus.

divmain:
Can you think of cases other than `git gui` where this would be of benefit?

asfaltboy:
Any `git` command that is long running: `git svn clone`, `git mergetool`, maybe others.

For non-`git` commands, users who want to remain in sublime, have many other plugins that focus on just that; such as [Tint](https://github.com/yyyc514/Tint), or [Glue](https://github.com/chrissimpkins/glue)

divmain:
I've been thinking about this and was close to deciding against merging.  Almost no Git commands can or should be run concurrently, the blocking nature of the single non-UI thread is desirable.  However, this is obviously useful to you, and the default behavior will remain the same.

So for now, this will be an undocumented (and non-recommended) option.  Should significant interest from others reveal itself, this might become a documented feature down the line.

asfaltboy:
:+1:  my thanks

cotcomsol:
I would like to suggest that this gets documented.  It was very helpful for me, but it took me a while to stumble upon it.  I tend to leave a git gui window open lots of the time and it really throws me when the other git commands don't seem to work.

divmain:
Fair enough.  @asfaltboy or @cotcomsol, would either of you be willing to put together a quick PR to update the documentation, making sure to include a warning that it may have undesirable effects if you don't know what you're doing?

asfaltboy:
on it

PR #233, opened 2015-09-10
Added an extra check when looking up the gitsavvy repo_path

theiviaxx:
When looking up the repo_path in git_command, it was possible to find a non-existing path that would make everything fail. This verified the path it find exists and if not, it will go the long way round to try to find one.

### DIFF
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -198,7 +198,7 @@ class GitCommand(StatusMixin,
         view = self.window.active_view() if hasattr(self, "window") else self.view
         repo_path = view.settings().get("git_savvy.repo_path")
 
-        if not repo_path:
+        if not repo_path or not os.path.exists(repo_path):
             file_path = self.file_path
             file_dir = os.path.dirname(file_path) if file_path else None
             working_dir = file_path and os.path.isdir(file_dir) and file_dir


divmain:
Would you mind squashing this into a single commit?  I'll merge as soon as that is done.  Thanks, this was a good thought.

theiviaxx:
im fairly new to git in terms of contributing.  I tried running git reset --soft HEAD~3 and then pushed back to github.  It doesnt look like it did what I thought it would.  What is the proper way to squash the commits into 1?

divmain:
No worries!  If you're using GitSavvy - I'm assuming you are ;) - you can squash like so:

- In the terminal, add a new remote for upstream: `git remote add upstream https://github.com/divmain/GitSavvy.git`.  This allows you to reference branches and commits in my canonical GitSavvy repo.
- Open sublime in the GitSavvy repo.
- `git: fetch` in the command palette, select `upstream`.  Now your local version of the repo knows the status of the upstream repo.
- `git: rebase` in the command palette to open the rebase dashboard.
- Press `f` to set which branch to compare your branch against - you should select `upstream/master`

Now the Rebase dashboard will be showing the difference between your branch and `upstream/master`.  Only one more step:

- Press `Q` to squash all visible commits.  This will take the diffs and commit messages from all commits shown and squash them into a single commit.  If you'd like to clean up the commit message afterwards, select what is now the only visible commit and press `e`.

Once you've squashed, do `git push (force)` to update your PR.  Let me know if any of that is unclear or something is left ambiguous.

divmain:
Also, just realized that in your case you could do it much simpler.  Since the first commit in this branch is the only one you care about, go to the terminal and do `git reset --hard 1fa9c233bfcfd0bc3503d71c9d9b9faab9b75e91`.  Then `git push --force` and you should be good to go.

theiviaxx:
using the second suggestion worked.  PR is updated

divmain:
Excellent, thanks!

stoivo:
:+1:

Issue #240, opened 2015-09-20
Keybinding for jumping to next / previous file in git status


dvcrn:
Hi! Testing GitSavvy and coming from SublimeGit. 

I'm desperately looking for a way to have git status start with the cursor on the first changed file, or at least a keybinding that navigates between files and not lines. 

Right now status opens and I have to press `down` a few times until I am at the file I want to perform actions on. SublimeGit had this very efficient. I think n automatically went to the next file and p to the previous ones. 

I searched through the keybindings and couldn't find a way to do this.

divmain:
I've been meaning to do this for awhile.  This issue gave me the little push to take a few minutes and add it.  Navigation between files in the status dashboard can now be accomplished using `.` for the next file and `,` for the previous file.  This follows the same key-binding patterns used in the inline-diff mode and elsewhere.

This will be included with the next release, but if you're running off of the GitSavvy repo, it is now in master branch.

dvcrn:
Man that was fast! 2h and already done! Huge thanks for that :) 

Now all there's left is opening git status with the first file selected and there is nothing I'm missing about SublimeGit (even though I have a license for it)

Any estimate on when the next release will be? I'm burning to replace SublimeGit in [sublimious](https://github.com/dvcrn/sublimious/tree/develop) but would wait until this is out before releasing the change

divmain:
You caught me when I needed a distraction! Good timing on your part :)  First-file preselection is now added.  There hasn't been much activity since the most recent release, but no real reason to hold it back.  I'll push up a new tag and it should show up in package control soon after.

dvcrn:
Haha looks like I created this issue at a good time. Amazing job, works perfectly! :smile:

dvcrn:
I noticed a small thing, not sure if on purpose or not: `,` and `.` can't navigate to stashes. This is not a big deal but if you rebind navigation (arrows and j/k) inside git status completely with the new command you just implemented, you can't select them anymore. 

I read a bit through the code. Would this be as easy as adding `meta.git-savvy.status.stashes` to https://github.com/divmain/GitSavvy/blob/00ba8fc5108abffbd9169116d047eac1b5d2f9b8/core/interfaces/status.py#L762-L764? 

I might tackle this one when I have a little bit more time tomorrow :hourglass:

PR #242, opened 2015-09-22
Added navigation to stashes to `NavigateFileCommand`

dvcrn:
As explained in https://github.com/divmain/GitSavvy/issues/240, navigation to stashes with `gs_status_navigate_file` is currently not possible. This can be a problem when all navigation inside git status is bound to that action and not to arrow keys anymore.


divmain:
Nice and clean.  Thanks!

Issue #236, opened 2015-09-15
When viewing inline-diff, shortcut to open the file at specific line


guncha:
First of all, thank you for all of your work on GitSavvy! It's been an amazing boost to my productivity. Here's a feature request that I think many would appreciate:

Before every commit, I always review all the changes using the inline-diff view and the stage/unstage hunk shorcuts are great for this. Sometimes though the change needs some extra attention and I really wish there was a way to open the file in question at the specific line using the same open shortcut as on the status page.

Thoughts?

divmain:
Agreed, that's a good idea.

stoivo:
:+1: I thought about adding it myself but did't get the time yet, would you like to add it?

stoivo:
t thought about `e` for edit.

ddevlin:
`o` for open would be more consistent with Status and Rebase views.

guncha:
:+1: for `o`

stoivo:
I made it, It is working good for me :smile:

divmain:
Thanks @stoivo!  I made a few changes, so now any cursor position is supported.

ddevlin:
Thanks @guncha for suggesting this, and @stoivo and @divmain for the implementation, just tried it, it's a great addition!

stoivo:
:sparkles:

PR #250, opened 2015-10-01
Feature: Open file to change when in inline diff

stoivo:
I created the feature to open file you that is open in inline diff.
Currently it is only for while you are in a hunk, but it should be implemented to do this for outside as well. 
If you have a good idea how to make this easily, I would like to hear it ✨ 

closes divmain/GitSavvy#236

### DIFF
--- a/core/commands/inline_diff.py
+++ b/core/commands/inline_diff.py
@@ -579,6 +579,48 @@ class GsInlineDiffStageOrResetHunkCommand(GsInlineDiffStageOrResetBase):
 
         return "\n".join([stand_alone_header] + hunk_ref.hunk.raw_lines[1:])
 
+class GsInlineDiffOpenFile(TextCommand):
+
+    """ Should open file at current hunk. """
+
+    def run(self, edit):
+        selections = self.view.sel()
+        region = self.view.line(0) if len(selections) == 0 else selections[0]
+
+        # Git lines are 1-indexed; Sublime rows are 0-indexed.
+        current_line_number = self.view.rowcol(region.begin())[0] + 1
+        current_column_number = self.view.rowcol(region.begin())[1] + 1
+
+        cur_hunk = self.get_current_hunk(current_line_number)
+        if not cur_hunk:
+            # do something when it is outside a hunk
+            # TODO
+            return
+
+        # line in current hunk
+        index = current_line_number - cur_hunk.section_start - 1
+
+        if cur_hunk.hunk.changes[index][1] == "+":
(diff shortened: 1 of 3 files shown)


core/commands/inline_diff.py
```diff
@@ -579,6 +579,48 @@ def get_diff_from_line(self, line_no, reset):
 
         return "\n".join([stand_alone_header] + hunk_ref.hunk.raw_lines[1:])
 
+class GsInlineDiffOpenFile(TextCommand):
+
+    """ Should open file at current hunk. """
+
+    def run(self, edit):
+        selections = self.view.sel()
+        region = self.view.line(0) if len(selections) == 0 else selections[0]
```
divmain: Nitpick: please change this to:
```python
region = selections[0] if selections else sublime.Region(0, 0)
```

core/commands/inline_diff.py
```diff
@@ -579,6 +579,48 @@ def get_diff_from_line(self, line_no, reset):
...
+
+        # Git lines are 1-indexed; Sublime rows are 0-indexed.
+        current_line_number = self.view.rowcol(region.begin())[0] + 1
+        current_column_number = self.view.rowcol(region.begin())[1] + 1
```
divmain:
```python
line_no, column_no = self.view.rowcol(region.begin())
# Git lines are 1-indexed; Sublime rows are 0-indexed.
line_no, column_no = line_no + 1, column_no + 1
```

divmain:
I pulled in your commit and merged after making changes to support any cursor position.  Thanks @stoivo!

stoivo:
Do you know when you will make a new release?

divmain:
@stoivo, usually I cluster features together.  Mostly, I do this to avoid irritating users with a new release and pop-up message every other day.  But I do release if somebody asks for it.

I hope you don't mind that I made significant changes.  I started commenting on the PR and realized it would be much quicker to just write the code.  There was a rebase, but the first commit is still attributed to you.

stoivo:
I totally agree that is stupid to place a release every other day. I just ask in case you had any schedule for releases.  

I read through your commit thoroughly so I can learn from it since you are really good at writing python. One day, I will get closer to write good python :+1: I don't mind you correcting me ;)

divmain:
Yeah, I don't have a formal release schedule, mostly because development velocity is inconsistent. Also, I'll just say your code is definitely good, and only limited when it comes to some of the language-specific syntax and idioms.

stoivo:
Definitely, I only learn python by reading some small plugins, but mostly this repo. And Thank you!!

Issue #251, opened 2015-10-02
remote branch becomes invisible, if it's name ends with '-'


pokidovea:
For example, if name of branch is
hotfix/popover-vertical-align-

the branch isn't visible on remote branches list of BRANCHES dashboard.

stoivo:
I am on it, just have to find the regex

PR #252, opened 2015-10-02
Fix: branch names ending on -

stoivo:
closes divmain/GitSavvy#251

simply just remove the \b. (if you don't remember like me) \b mean that is should be end of a string and have a *blank* character after. Therefor it did not match `some-name-`. It exist \B that is  the opposite of \b. So it it matches with a symbol like -+!~@#$% etc..

I did not see why it was needed. I guess there were some reason it got added there, so. Do any of you remember why?


divmain:
Hmm I don't recall.  Let's merge and give those on master some time to report an issue before we push out a release.

stoivo:
:+1:

PR #268, opened 2015-10-20
Fix: display all types of unmerged conflicts

### LINKED ISSUE #267: Bug: rebase conflict resolution does not display added files
When both conflicting commits "add a file", the rebase dashboard does not display the conflict, and it fails to continue.

    rebase in progress; onto 12738af
    You are currently rebasing branch 'wip-git-flow-interface' on '12738af'.
      (fix conflicts and then run "git rebase --continue")
      (use "git rebase --skip" to skip this patch)
      (use "git rebase --abort" to check out the original branch)

    Unmerged paths:
      (use "git reset HEAD <file>..." to unstage)
      (use "git add <file>..." to mark resolution)

        both modified:   Default.sublime-commands
        both modified:   core/commands/__init__.py
        both added:      core/commands/flow.py
        both added:      docs/flow.md

What may happen is that `RewriteMixin` methods are called by rebase dashboard to find *changes* to files; however, even when resolving "both changed" conflicts, the "both added" files still remain in conflict, and do not appear in rebase dash:

screen shot 2015-10-20 at 9 10 48 am

Git status now returns:

    rebase in progress; onto 12738af
    You are currently rebasing branch 'wip-git-flow-interface' on '12738af'.
      (fix conflicts and then run "git rebase --continue")
      (use "git rebase --skip" to skip this patch)
      (use "git rebase --abort" to check out the original branch)

    Unmerged paths:
      (use "git reset HEAD <file>..." to unstage)
      (use "git add <file>..." to mark resolution)

        both added:      core/commands/flow.py
        both added:      docs/flow.md

```patch
grep core/commands/flow.py .git/rebase-apply/*
.git/rebase-apply/0001: core/commands/flow.py     | 302 ++++++++++++++++++++++++++++++++++++++++++++++
.git/rebase-apply/0001: create mode 100644 core/commands/flow.py
.git/rebase-apply/0001:diff --git a/core/commands/flow.py b/core/commands/flow.py
.git/rebase-apply/0001:+++ b/core/commands/flow.py
.git/rebase-apply/0002: core/commands/flow.py    | 99 +++++++++++++++++++++++++++++-------------------
.git/rebase-apply/0002:diff --git a/core/commands/flow.py b/core/commands/flow.py
.git/rebase-apply/0002:--- a/core/commands/flow.py
.git/rebase-apply/0002:+++ b/core/commands/flow.py
.git/rebase-apply/0003: core/commands/flow.py | 70 +++++++++++++++++++++++++++++++++++++++------------
.git/rebase-apply/0003:diff --git a/core/commands/flow.py b/core/commands/flow.py
.git/rebase-apply/0003:--- a/core/commands/flow.py
.git/rebase-apply/0003:+++ b/core/commands/flow.py
.git/rebase-apply/patch:diff --git a/core/commands/flow.py b/core/commands/flow.py
.git/rebase-apply/patch:+++ b/core/commands/flow.py
```

asfaltboy:
fix #267

### DIFF
--- a/core/interfaces/rebase.py
+++ b/core/interfaces/rebase.py
@@ -191,10 +191,31 @@ class RebaseInterface(ui.Interface, GitCommand):
         return commits_info
 
     def _get_conflicts_in_rebase(self):
+        """
+        Look for unmerged conflicts in status, which are one of:
+           DD    unmerged, both deleted
+           AU    unmerged, added by us
+           UD    unmerged, deleted by them
+           UA    unmerged, added by them
+           DU    unmerged, deleted by us
+           AA    unmerged, both added
+           UU    unmerged, both modified
+
+        FIXME: Currently, we ignore scenarios with D (delete) since they require
+        a different set of commands to handle them in rebase dashbaord.
+        """
         return [
             entry.path
             for entry in self.get_status()
-            if entry.index_status == "U" and entry.working_status == "U"
+            if (
+                # (entry.index_status == "D" and entry.working_status == "D") or
+                (entry.index_status == "A" and entry.working_status == "U") or
+                # (entry.index_status == "U" and entry.working_status == "D") or
+                (entry.index_status == "U" and entry.working_status == "A") or
+                # (entry.index_status == "D" and entry.working_status == "U") or
+                (entry.index_status == "A" and entry.working_status == "A") or
+                (entry.index_status == "U" and entry.working_status == "U")
+            )
         ]
 
     def _get_diverged_outside_rebase(self):


divmain:
Nice!  Have you tested these scenarios to confirm everything behaves as expected?

asfaltboy:
@divmain I tested only the one I encountered which was UU (both modified) and AA (both added), I assume it'll be the same for the rest, since they are documented. Though we really oughta add a test suite :zap: !

divmain:
A test suite would be best, I agree. I spent some time working on it, but it was more difficult than you might expect. It is on my list... ¯\_(ツ)_/¯

For now, can you double-check the `UD unmerged, deleted by them` scenario?  If that looks good, I'll merge this in.  Thanks!

asfaltboy:
**TL;DR** delete needs a special case, so D scenarios can be commented out for now and a separate PR opened for these (out of time here).

As expected, delete is different: though it now displays the file in conflict, the command results differ.

Firstly the `use version from X` commands cannot be run for the deleted side (the other side restores the object); an attempt to use a deleted side shows an informative message:

```
`/usr/local/bin/git checkout --theirs -- project/models.py` failed with following output:

error: path 'project/models.py' does not have their version
```

Second, mergetool (when configured) cannot be launched, and shows a less informative error:

```
`/usr/local/bin/git config mergetool.opendiff.cmd` failed with following output:

```

This is because `git` requires a decision to run the mergetool.

```bash
$ git mergetool
Merging:
project/models.py

Deleted merge conflict for 'project/models.py':
  {local}: modified file
  {remote}: deleted
Use (m)odified or (d)eleted file, or (a)bort?
```

divmain:
Here are my thoughts:

- `use version from X` would work by either using the file if it is present in version X, or deleting the file and staging that deletion if the file is not present in version X
- disable merge tool for files with one side that is deleted - it isn't very useful, anyway

divmain:
Made these changes on top of your original commit and manually merged into master.

asfaltboy:
Wow that was fast, couldn't reply on time; but was going to suggest 2 more things:
1. The format should indicate which kind of conflict we have (i.e use a more verbose syntax `"    ┃           ! {} {}".format(conflict_letters, conflict.path)`).
2. Disabling "mergetool" command should have some visual indication, such as "lighter" font color, or strike-through, or status bar indication alert.

divmain:
At this time, I'd prefer not to include the conflict type in the interface.  I've been thinking about it off and on ever since I created the rebase dashboard, and the only thing I can think of is to use the same conflict-type letter representations that Git uses.  But those are hard to remember - I've always had to go look them up to see what they stood for.  And if people can't remember what they stand for, they add clutter and result in net-negative user experience.

I'm open to adding visual indicators for files that can't be handed off to the merge tool, so long as the UX is intuitive.

Issue #261, opened 2015-10-07
Feature request: keyboard shortcuts to jump between status,branch,tags,graph


burbanczyk:
Hi
I'm using GitSavvy in my sublime workflow and I think it would be very handy to be able to jump between git:status git:branch git:tags git:graph either using keys like 1 2 3 4 or(and) TAB.
The main benefit would be less open tabs and a lot less keystrokes to jump from status to graph or branch.

max-mykhailenko:
Good idea! It will be dashboard page with one hotkey.

stoivo:
I think we have decided not to add have to many global keystrokes, since than you may have some conflicts with other packages. But you are free to add you own keystrokes:

if you do to `Preferences` > `Key bindings - User` 
and add the following.
```json
[
  { "keys": ["control+1"], "command": "gs_show_status" },
  { "keys": ["control+2"], "command": "gs_show_branch" },
  { "keys": ["control+3"], "command": "gs_show_tags" },
  { "keys": ["control+4"], "command": "gs_log_graph" },
  { "keys": ["control+5"], "command": "gs_show_tags", "args": { "all_branches": true } },
]
```
or you may like to have it on F2 ..
```json
[
  { "keys": ["F2"], "command": "gs_show_status" },
  { "keys": ["F3"], "command": "gs_show_branch" },
  { "keys": ["F4"], "command": "gs_show_tags" },
  { "keys": ["F5"], "command": "gs_log_graph" },
  { "keys": ["F6"], "command": "gs_show_tags", "args": { "all_branches": true } },
]
```

burbanczyk:
Yes I know I can do that in preferences but, that's not exactly what I meant. 
Adding key bindings would open new tab each time and it's still a two key combination.

What I'm proposing is that 1 2 3 4 keys wouldn't be global - they would work only on GitSavvy. This way you would utilize only one tab in sublime and have everything you need one keystroke away.

stoivo:
So you want to switch from status to branch on the same tab with one keystroke?

If we just close status view and open branch view? Than you get the behavior you want?

burbanczyk:
Pretty much, if the tab and focus stays in place then yes.
I'm imagining workflow like this: 
I'm on git:status, i use (A) to add files, then (c) to commit, (ctrl+enter) to submit that commit. Next i want to jump to git:graph I use (4),  Everything is fine so I use (2) to jump to git:branch next i choose a branch and hit (c) to checkout.
[...]

Issue #184, opened 2015-05-15
Help command for all views


stoivo:
I was thinking that is might be good to have easy-access-help for all views.
by pressing "?" to toggle a write-only quick panel. 
for diff, inline diff, blame, etc?
This will make the docs easy to access.

I started to read all keybindings-file for find new functions.
What do you think about this?

divmain:
There is a `GitSavvy: help` command in the Sublime palette.  This will give users access to either general, wiki-style help, or bring them to a specific page when in certain views.  Do you have something in mind other than this?

stoivo:
I was thinking something like this ![screenshot 2015-05-16 04 20 41](https://cloud.githubusercontent.com/assets/3492040/7666489/921877d2-fbe7-11e4-8280-af4241dce860.png)

and some color highlighting. But I do not know who to do it.

divmain:
It is a neat idea, but I think having two live-help systems is unnecessary.  Also, highlighting in views could be difficult if you wanted control over the colors used.  It would be possible to underline, etc, without any issues but that might be confused for plugins that show errors.

stoivo:
Yes, I totally agree with you that is is unnecessary,  but it will only contain shortcuts for each view.
it is easy to add custom colouring in the quick panel. Like this(Just some silly ruby code)
![screen shot 2015-05-19 at 21 53 38](https://cloud.githubusercontent.com/assets/3492040/7712648/c3eb64ee-fe72-11e4-9b48-f02ac94ac581.png)
)

code snippet to do it 
```
class ShowHelpView( sublime_plugin.TextCommand):
    def run( self, edit ):
        other_string = """\
class Foo
  def ee
    3
  end
end"""
        string = """\
inline diff help

[,] previous hunk
[.] next hunk

[h] stage hunk         [H] discard hunk
[l] stage line         [L] discard line

[r] refresh view

{} """

        if (sublime.platform() == "Windows"):
            undo_command = "[CTRL-z] on Windows    "
        else:
            undo_command = "[SUPER-z] on Linux/OSX "

        self.output_view = self.view.window().create_output_panel("inline-diff-help")
        self.output_view.set_read_only(False)
        # self.output_view.insert(edit, 0, string.format(undo_command))
        self.output_view.insert(edit, 0, other_string)
        self.output_view.set_syntax_file("Packages/Ruby/Ruby.tmLanguage")
        self.output_view.set_read_only(True)
[...]

PR #282, opened 2015-11-02
Enhancement: shift-tab to cycle dashboards in reverse

marcinruszkiewicz:
Tab cycling through dashboards is an awesome thing, but there was no way to go back to the previous one and in the discussion in #261 it was mentioned briefly as an issue but wasn't really resolved as the discussion there went in a different direction.

Why not just remove the unnecessary dashboards through the config? Unless you remove everything but two dashboards, this still is a nice enhancement to have.

This adds the keybinding (as is common, if I cycle one way with tab then shift-tab should cycle it back) and lists it in the various dashboards (with proper syntax coloring).

I've used [SHIFT-tab] in the dashboard keybinding lists, to be in line with [SUPER-Z] keybindings in the rebase dashboard.

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -315,6 +315,14 @@
             { "key": "setting.git_savvy.status_view", "operator": "equal", "operand": true }
         ]
     },
+    {
+        "keys": ["shift+tab"],
+        "command": "gs_tab_cycle",
+        "args": { "source": "status", "reverse": true },
+        "context": [
+            { "key": "setting.git_savvy.status_view", "operator": "equal", "operand": true }
+        ]
+    },
 
     /////////////////
     // COMMIT VIEW //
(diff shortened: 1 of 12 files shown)


divmain:
Thanks for this @marcinruszkiewicz.  I like everything about this, with the one exception of the extra keybinding in the menu.  I have found, both for myself and in talking to others, that the more keybindings are displayed in the menu, the more cluttered it appears and the less useful it becomes.  For that reason, I've tried to keep keybindings lean.

However, this is useful functionality, and addresses real concerns that have been voiced.  So I'm going to merge anyway, and think about whether we can clean up the bindings menu in some way.

stoivo:
@divmain, do you remember when we talked about #184. Maybe we can have something like this. So only show the very most used bindings. And have a '?'-command. That will show those that are  less in used.

marcinruszkiewicz:
@divmain Thanks for the merge! 

As for the keybindings, hiding some bindings like @stoivo is saying sounds allright - with the exception of hiding not the "least used" but the "universal" bindings, like the tabbing and r to refresh (the last one is not actually universal but probably should). Tabbing through dashboards is hardly least used, but it makes sense to hide it.

PR #298, opened 2015-12-01
Better commit on close behavior

### LINKED ISSUE #293: "commit_on_close" not working
Regular commit works with ctrl+enter but when I activate the `"commit_on_close": true` option and close the commit window (with a commit message written) it fails to commit. 

Console panel logs the following:

      File "core.commands.commit in C:\Program Files\Sublime Text\Data\Installed Packages\GitSavvy.sublime-package", line 131, in run_async
    IndexError: list index out of range

asfaltboy:
This PR attempts to fix #293 by explicitly passing the message on view close. It is my understanding that when run_async is called, the original view is not guaranteed to exist anymore, hence the exception seen by @laggingreflex.

Additionally, I made sure that `prompt_on_abort_commit` setting does not prevent `commit_on_close` setting, as well as added a condition to Super/Ctrl+Enter keymap to prevent GitSavvy from attempting a double commit.

### DIFF
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -125,13 +125,15 @@ class GsCommitViewDoCommitCommand(TextCommand, GitCommand):
     make a commit using the text for the commit message.
     """
 
-    def run(self, edit):
-        sublime.set_timeout_async(self.run_async, 0)
+    def run(self, edit, message=None):
+        sublime.set_timeout_async(lambda: self.run_async(commit_message=message), 0)
 
-    def run_async(self):
-        view_text = self.view.substr(sublime.Region(0, self.view.size()))
-        help_text = self.view.settings().get("git_savvy.commit_view.help_text")
-        commit_message = view_text.split(help_text)[0]
+    def run_async(self, commit_message=None):
+        if commit_message is None:
+            view_text = self.view.substr(sublime.Region(0, self.view.size()))
+            help_text = self.view.settings().get("git_savvy.commit_view.help_text")
+            commit_message = view_text.split(help_text)[0]
+
         include_unstaged = self.view.settings().get("git_savvy.commit_view.include_unstaged")
 
         show_panel_overrides = \
@@ -193,4 +195,4 @@ class GsCommitViewCloseCommand(TextCommand, GitCommand):
             message_txt = message_txt.strip()
 
             if message_txt:
-                self.view.run_command("gs_commit_view_do_commit")
+                self.view.run_command("gs_commit_view_do_commit", {"message": message_txt})
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -149,9 +149,12 @@ class GsCommitViewDoCommitCommand(TextCommand, GitCommand):
             stdin=commit_message
             )
 
-        self.view.window().focus_view(self.view)
-        self.view.set_scratch(True)  # ignore dirty on actual commit
-        self.view.window().run_command("close_file")
+        # ensure view is not already closed (i.e.: when "commit_on_close" enabled)
+        is_commit_view = self.view.settings().get("git_savvy.commit_view")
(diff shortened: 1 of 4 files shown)


laggingreflex:
Tested it out but getting this error (when closing tab with a commit message (with `"commit_on_close":true`):)

      File "C:\...\GitSavvy\core\commands\commit.py", line 129, in <lambda>
        sublime.set_timeout_async(lambda: self.run_async(commit_message=message), 0)
      File "C:\...\GitSavvy\core\commands\commit.py", line 149, in run_async
        stdin=commit_message
      File "C:\...\GitSavvy\core\git_command.py", line 135, in git
        command_str, stdout, stderr
      File "C:\...\GitSavvy\core\git_command.py", line 109, in raise_error
        raise GitSavvyError(msg)

    GitSavvy.core.git_command.GitSavvyError: `C:\Program Files\Git\bin\git.EXE commit -q -F -` failed with following output:
    On branch master
    Changes not staged for commit:
        modified:   a.txt

    no changes added to commit

asfaltboy:
Interesting... I have to ask, do you have anything staged for commit at this point?

laggingreflex:
ah my bad, I didn't have anything staged. I was hitting Shift+C earlier which was expected to auto stage. 

Anyways it works!

Although there's still an error:

      File "C:\...\GitSavvy\core\commands\commit.py", line 129, in <lambda>
        sublime.set_timeout_async(lambda: self.run_async(commit_message=message), 0)
      File "C:\...\GitSavvy\core\commands\commit.py", line 152, in run_async
        self.view.window().focus_view(self.view)
    AttributeError: 'NoneType' object has no attribute 'focus_view'

And I think this is causing another issue which is that you have to Ctrl-tab away from the STATUS view for it to refresh, otherwise it still keeps showing "staged", and not "Your working directory is clean", even if you hit `r` (manually refresh).

divmain:
Thanks a good catch @laggingreflex.  Looks like, after this new behavior is introduced, the original view has already been closed at this point and no longer has a window reference, i.e. `self.view.window() == None`.  @asfaltboy, can you add a fix for this, and double-check that behavior (both with and without commit-on-close enabled) is as expected when interacting with the commit view?

asfaltboy:
That seems to do it.

The refresh behavior is strange, seems the status view is auto-refreshed only when it's activated after commit if `commit_on_close` is disabled (with ctrl+enter) but not after commit with `commit_on_close` enabled (view is closed by user).

Moreover hitting `r` never does anything for me for some reason ... Oh right, the command to refresh has changed but not updated in keymap. I'll submit a separate PR.

core/commands/commit.py
```diff
@@ -64,9 +64,12 @@ def run_async(self, repo_path=None, include_unstaged=False, amend=False):
         else:
             view.set_syntax_file("Packages/GitSavvy/syntax/make_commit.tmLanguage")
 
+        commit_on_close = savvy_settings.get("commit_on_close")
+        view.settings().set("git_savvy.commit_on_close", commit_on_close)
```
divmain: This setting doesn't appear to be referenced later.

divmain:
Just the one comment.  I'll merge once that's fixed, and we can dogfood this for a bit to make sure it works as expected.  Thanks @asfaltboy!

core/commands/commit.py
asfaltboy: It *is used* in the [this commit](https://github.com/asfaltboy/GitSavvy/commit/4e20d1639dbb94ae1b3c39ffce001ceadb192638) to check toggle super/ctrl+enter off in commit view when using `commit_on_close`.

divmain: Ah, I had missed that somehow!

divmain:
Looks good.  Thanks again!

PR #305, opened 2015-12-03
Enhancement: Prompt for git reset mode

### LINKED ISSUE #304: reset --hard?
How to to do reset --hard?

ddevlin:
This commit adds an extra step to the `git: reset` and `git: reset (reflog)` commands that prompts for the git reset mode (e.g. --soft, --hard). By default this step is disabled, and can be enabled by setting `prompt_for_reset_mode` to `true` in GitSavvy.sublime-settings.

Closes #304 

ezgif com-video-to-gif

### DIFF
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -233,5 +233,11 @@
         "rebase",
         "tags",
         "graph"
-    ]
+    ],
+
+    /*
+        Set this value to a valid git reset mode (e.g. --hard, --soft) to skip
+        reset mode prompt and always use that reset mode.
+     */
+    "use_reset_mode": null
 }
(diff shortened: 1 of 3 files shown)


GitSavvy.sublime-settings
```diff
@@ -233,5 +233,11 @@
...
         "tags",
         "graph"
-    ]
+    ],
+
+    /*
+        Set this value to true if you'd like GitSavvy to prompt you for a
+        reset mode (e.g. --hard, --soft) on reset operations.
+     */
+    "prompt_for_reset_mode": false
```
divmain: This seems like it'd be useful for the majority of users that are doing a reset, so let's turn it on by default.

ddevlin:
@divmain Updated and squashed.

divmain:
Very nice.  Thank you, sir!

PR #306, opened 2015-12-03
Fix: Incorrect syntax highlighting in docs/README.md

ddevlin:
Small nitpicky fix, comment in settings JSON example used python comment delimiter `#` instead of JS comment delimiter `//`. Additionally, comments aren't valid syntax in a Github Flavoured Markdown `json` code blocks, so changed the highlighting mode to `javascript`.

screen shot 2015-12-03 at 22 44 53


divmain:
Heh.  Thanks!

PR #310, opened 2015-12-07
Pull request tweaks

stevage:
Enhancement: show user logins, and friendly times in pull request list.


github/commands/pull_request.py
```diff
@@ -8,13 +8,46 @@
 from ...common import interwebs
 from ...common import util
 
+def time_delta(dt):
```
divmain: We have something like this already in `common/util/dates.py`.  I'm going to test out the new PR features by pulling down your changes, and will merge manually into master :)

PR #313, opened 2015-12-09
Reset hints

stevage:
Enhancement: Show a hint next to each "git reset" option explaining what it does.

### DIFF
--- a/core/commands/reset.py
+++ b/core/commands/reset.py
@@ -3,8 +3,17 @@ import sublime
 from .log import GsLogCommand
 from ...common import util
 
-
-GIT_RESET_MODES = ["--mixed", "--soft", "--hard", "--merge", "--keep"]
+PADDING = "                                                "
+GIT_RESET_MODES = [
+    # See analysis at http://stackoverflow.com/questions/34149356/what-exactly-is-the-difference-between-all-the-git-reset-modes/34155307#34155307
+    ["--mixed" + PADDING, "unstage staged, keep unstaged, don't touch working (safe)"],
+    ["--soft", "just move HEAD, stage differences (safe)"],
+    ["--hard", "discard staged, discard unstaged, update working (unsafe)"],
+    ["--merge", "discard staged, keep unstaged, update working (abort if unsafe)"],
+    ["--keep", "unstage staged, keep unstaged, update working (abort if unsafe)"]
+    # For reference, in case we ever include the (very similar) checkout command
+    # ["--checkout", "keep staged, keep unstaged, update working, move branches (abort if unsafe)"]
+]
 
 
 class GsResetCommand(GsLogCommand):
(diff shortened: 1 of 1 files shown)


ddevlin:
Hi @stevage, nice suggestion with the reset mode hints :smile: 

@divmain has the final say on these things, but I have a few suggestions:

### Quick panel styling
If you give the `show_quick_panel` function a list of 2 element lists it will use the first item as the option title, and the second item as a subtitle, e.g. instead of this:
```python
["mixed (default): unstage staged, keep unstaged, don't touch working (safe)", ...]
```
do this:
```python
[["--mixed", "unstage staged, keep unstaged, don't touch working (safe)"], ...]
```
and it will look like this:
screen shot 2015-12-09 at 11 37 17

### DRY reset modes
One nice thing in the old implementation was that the quick panel list was derived from `GIT_RESET_MODES`, so they couldn't go out of sync. How about changing `GIT_RESET_MODES` to:
```python
GIT_RESET_MODES = [
    ["--mixed", "unstage staged, keep unstaged, don't touch working (safe)"],
    ["--soft", "just move HEAD, stage differences (safe)"],
    ["--hard", "discard staged, discard unstaged, update working (unsafe)"],
    ["--merge", "discard staged, keep unstaged, update working (abort if unsafe)"],
    ["--keep", "unstage staged, keep unstaged, update working (abort if unsafe)"]
    # For ref
    # ["--checkout", "keep staged, keep unstaged, update working, move branches (abort if unsafe)"]
]
```
Then you can pass `GIT_RESET_MODES` directly to `show_quick_panel`: 
```python
self.window.show_quick_panel(
    GIT_RESET_MODES, self.on_reset_mode_selection, flags=sublime.MONOSPACE_FONT
)
```
making sure to extract the mode to be passed to `git` from the first element of the selected list from `GIT_RESET_MODES`:
```python
def on_reset_mode_selection(self, index):
    if 0 <= index < len(GIT_RESET_MODES):
        self.on_reset(GIT_RESET_MODES[index][0])
```

### Squashing commits
As your second commit is just a tweak to the first one, it's probably better to squash them into one commit.

divmain:
Agreed on all points!

stevage:
Well, thanks for doing all the work there :) Only thing was, using the 2-element mode meant the drop down wasn't wide enough to actually read them (on my machine). So I've added some padding.

Also I've re-rebased and squashed. Not sure if that was really the right thing to do. *shrug*

divmain:
Excellent - thanks for this!

As to the re-rebase + squash, that was appreciated also.  I use the `git: generate changelog` for release notes when a new version is published.  Having single, discrete feature-commits helps with this tremendously (as well as the `TOPIC: ...` format).

stevage:
Ah, cool. I wasn't sure whether the "TOPIC: ..." thing was supposed to be a commit or in the PR comment - now I get it.

PR #320, opened 2015-12-12
fix get_rel_path

randy3k:
If the repo is symlinked and the file opened from the symlinked directory, `get_rel_path` doesn't work.

`self.repo_path` shows the real directory but `path` is in terms of the symlink directory.


divmain:
Thanks!  I'll let this sit in master for a few days, to ensure it doesn't introduce anything unintended.

PR #324, opened 2015-12-15
Fix: Syntax files don't match short SHA1 refs longer than 7

### LINKED ISSUE #323: Eight-character SHAs not properly highlighted in UI views
Occasionally, a Git SHA is displayed with eight characters instead of seven. For example, in the Rebase view:

rebase__murmur_ _culture_amp

I’m not sure if the bug is the length of the SHA or the lack of highlighting for the extra character, but there’s obviously something wrong.

ddevlin:
Git will only generate short SHA1 refs that are 7 characters long if the 7 character ref is unambiguous. For larger repositories Git will begin to generate larger short SHA1 refs to keep refs unambiguous. This fix allows GitSavvy to correctly highlight refs longer than 7 characters.

See http://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#Short-SHA-1 for more information.

Closes #323.

### DIFF
--- a/syntax/branch.YAML-tmLanguage
+++ b/syntax/branch.YAML-tmLanguage
@@ -23,7 +23,7 @@ patterns:
     match: '(ahead)|(behind)'
   - comment: head summary
     name: comment.git-savvy.status.head-commit
-    match: '^  HEAD:    ([0-9a-f]{7}) .+'
+    match: '^  HEAD:    ([0-9a-f]{7,40}) .+'
     captures:
       '1': { name: constant.other.git-savvy.status.sha1 }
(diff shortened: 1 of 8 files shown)


sentience:
Looks good! Thanks for the fix, @ddevlin!

divmain:
This is great!  I'd also like for the short-hashes to be padded on the right, so that the next column lines up.  How do you feel about adding that to this PR?  If you're not comfortable giving it a try, we can split that off into a separate ticket and merge this in as-is.

divmain:
Going to merge this as-is.

PR #327, opened 2015-12-17
prerelease tag name support

randy3k:
Two changes.

1. add per-release support, if the last tag is in form of `2.3.5-alpha.1`, `0.7.2-rc.10`, the last set of number will be increased by one: `1.4.6-alpha` to `1.4.6-alpha.1` and `2.3.5-rc.1` to `2.3.5-rc.2`.

2. the `tagname` passes to `show_input_panel` instead of running `on_entered_name` directly. It gives user a chance to modify the tag 

These things may be worth discussions:

1. How we should allow user to jump from a pre-release to a regular release. For example, if the last tag is 
`0.7.2-rc.10`, should "git: quick tag patch" yields "0.7.2"? how about "tag minor" and "tag major"?
2. on the other hand, if the last tag is a regular release, say `1.4.6`, should "git: quick tag prerelease" yields `1.4.6-alpha`? how about "beta" and "rc"?
3. other forms of prereleases? there is million ways to specify a pre-release, for examples,
`1.0.0-0.3.7`, `1.0.0-alpha+001`. How about build numbers? `1.0.0+20130313144700`, `1.0.0-beta+exp.sha.5114f85`

### DIFF
--- a/core/interfaces/tags.py
+++ b/core/interfaces/tags.py
@@ -24,26 +24,64 @@ TAG_CREATE_MESSAGE_PROMPT = "Enter message:"
 START_PUSH_MESSAGE = "Pushing tag..."
 END_PUSH_MESSAGE = "Push complete."
 
+RELEASE_REGEXP = re.compile(r"^([0-9A-Za-z-]*[A-Za-z-])?([0-9]+)\.([0-9]+)\.([0-9]+)(?:-([0-9]+))?$")
 
-def smart_incremented_tag(tag, release):
+def smart_incremented_tag(tag, release_type):
     """
     Automatic increment of a given tag depending on the type of release.
+
+    >>> smart_incremented_tag('v1.3.2', "prerelease") == 'v1.3.3-0'
+    >>> smart_incremented_tag('v1.3.2', "prepatch") == 'v1.3.3-0'
+    >>> smart_incremented_tag('v1.3.2', "preminor") == 'v1.4.0-0'
+    >>> smart_incremented_tag('v1.3.2', "premajor") == 'v2.0.0-0'
+    >>> smart_incremented_tag('v1.3.2', "patch") == 'v1.3.3'
+    >>> smart_incremented_tag('v1.3.2', "minor") == 'v1.4.0'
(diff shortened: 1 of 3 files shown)


core/interfaces/tags.py
```diff
@@ -25,24 +25,39 @@
...
-            major = str(int(major)+1)
-            minor = patch = "0"
-        elif release == "minor":
-            minor = str(int(minor)+1)
-            patch = "0"
-        elif release == "patch":
-            patch = str(int(patch)+1)
-        return prefix + major + "." + minor + "." + patch
+    if release_type in ["major", "minor", "patch"]:
+        r = re.compile(r"^([0-9A-Za-z-]*[A-Za-z-])?([0-9]+)\.([0-9]+)\.([0-9]+)$")
```
divmain: Please move these compiled regular expressions to the top of the file, with a descriptive name.  No need to compile each time the command is run, just on start-up.

divmain:
I would like to use the conventions that `npm` uses with `npm version` ([link](https://docs.npmjs.com/cli/version)).  The approach has straightforward semantics, and has the benefit of already having been adopted by a large open-source community.

I also think it may be worth considering consolidating all of these new commands into one (`git: smart tag`), and then prompt for release type (`major`, `minor`, `patch`, `premajor`, `preminor`, etc).

#### Examples

`1.4.5` → `git: smart tag` → `premajor` → `2.0.0-0`
`1.4.5` → `git: smart tag` → `minor` → `1.5.0`
`2.1.0-0` → `git: smart tag` → `minor` → `2.1.0`

randy3k:
it makes sense. I can work on it later today.

randy3k:
@divmain check the updated code.

randy3k:
Oh....there is an issue. The tags is not in ordered, `2.8.3-2` is before `2.8.3`..


screen shot 2015-12-17 at 10 41 26 pm

randy3k:
Added a function `get_lastest_local_tag` to fetch the latest tag.

divmain:
This is fantastic!

PR #330, opened 2015-12-18
Fix: the default should be a string

randy3k:
`show_input_panel` doesn't like `None`.

PS: sorry for not testing the new code well..


divmain:
No problem.

PR #331, opened 2015-12-18
move tag commands out of interfaces

randy3k:
`GsTagCreateCommand` and `GsSmartTagCommand` are not interface-related commands, they should be placed under the `core.command` submodule.


divmain:
Seems reasonable.  :+1:

PR #333, opened 2015-12-18
Improve: get_lastest_local_tag in the current branch only

randy3k:
This allows `dev` branch and `stable` branch to tag simultaneously.

For example, the `stable` is tagging `v1.2.3` while the `dev` branch, the next major version, is tagging `v2.0.0-1`

### DIFF
--- a/core/git_mixins/tags.py
+++ b/core/git_mixins/tags.py
@@ -26,11 +26,14 @@ class TagsMixin():
 
         return entries
 
-    def get_lastest_local_tag(self):
+    def get_lastest_local_tag(self, all_branches=False):
         """
         Return the latest tag. get_tags() fails to return an ordered list.
         """
+        if all_branches:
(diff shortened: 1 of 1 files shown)


core/git_mixins/tags.py
```diff
@@ -26,11 +26,14 @@ def get_tags(self, remote=None, reverse=False):
 
         return entries
 
-    def get_lastest_local_tag(self):
+    def get_lastest_local_tag(self, all_branches=False):
```
divmain: In what conditions will this flag be `True`?

randy3k: To be honest, I don't know. I just feel like we should leave it for the future.

divmain:
@randy3k, I prefer not to include dead code, even if there's a future use for it.  I'm going to close this PR for now, but we can re-open if 1) the feature is accessible via custom Sublime commands _and_ has a documented use case, or 2) the feature is utilized internally.

PR #336, opened 2015-12-28
Feature: allow prompt to specify custom command argument

asfaltboy:
This will allow for more dynamic commands to be created.

I moved custom command documentation to it's own custom.md file since it started to grow too much for README index IMO, and also added some more custom command examples to the doc.

### DIFF
--- a/core/commands/custom.py
+++ b/core/commands/custom.py
@@ -27,6 +27,23 @@ class GsCustomCommand(WindowCommand, GitCommand):
     """
 
     def run(self, **kwargs):
+        if not kwargs.get('args'):
+            sublime.error_message("Custom command must provide args.")
+            return
+
+        # prompt for custom command argument
+        if '{PROMPT_ARG}' in kwargs.get('args'):
+            prompt_msg = kwargs.pop("prompt_msg", "Command argument: ")
+            return self.window.show_input_panel(
+                prompt_msg,
+                "",
+                lambda arg: sublime.set_timeout_async(
+                    lambda: self.run_async(custom_argument=arg, **kwargs), 0
+                ),
+                None,
+                None
+            )
+
         sublime.set_timeout_async(lambda: self.run_async(**kwargs), 0)
 
     def run_async(self,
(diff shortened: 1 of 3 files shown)


core/commands/custom.py
```diff
@@ -27,8 +27,17 @@ class GsCustomCommand(WindowCommand, GitCommand):
     """
 
     def run(self, **kwargs):
+        if '{PROMPT_ARG}' in kwargs.get('args'):
+            self.last_kwargs = kwargs
+            msg = kwargs.pop("prompt_msg", "Command argument: ")
+            self.window.show_input_panel(msg, "", self.custom_argument_entered,
+                                         None, None)
```
divmain: This should have a `return` statement after it, no?

divmain: Also, please put all args on a single line, or each arg on a separate line :)

core/commands/custom.py
```diff
@@ -27,8 +27,17 @@ class GsCustomCommand(WindowCommand, GitCommand):
     """
 
     def run(self, **kwargs):
+        if '{PROMPT_ARG}' in kwargs.get('args'):
```
divmain: Line 39/48 below (`if not args:`) should probably be moved up here, if you're going to assume the presence of an `args` keyword argument.  Otherwise, this line will throw an exception.

core/commands/custom.py
```diff
@@ -27,8 +27,17 @@ class GsCustomCommand(WindowCommand, GitCommand):
...
+            self.last_kwargs = kwargs
+            msg = kwargs.pop("prompt_msg", "Command argument: ")
+            self.window.show_input_panel(msg, "", self.custom_argument_entered,
```
divmain: It seems that this could be simplified into a single `lambda`, eliminating the `custom_argument_entered` method below.

asfaltboy:
@divmain rebased commit with requested changes, please confirm.

divmain:
This looks great.  Thanks @asfaltboy!

Issue #328, opened 2015-12-18
Git Log current file > show whole file


stevage:
Could `git: log current file` allow `Show whole file`as an option after selecting the commit in question?

divmain:
Can do.

PR #337, opened 2016-01-02
Feature: Git log current file > Show file at commit

craigaaroncampbell:
Fixes issue #328 

moved array in self.window.show_quick_panel() and stored in options_array  so that additional options may be added conditionally

### DIFF
--- a/core/commands/show_file_at_commit.py
+++ b/core/commands/show_file_at_commit.py
@@ -1,0 +1,33 @@
+import sublime
+from sublime_plugin import WindowCommand, TextCommand
+
+from ..git_command import GitCommand
+
+SHOW_COMMIT_TITLE = "COMMIT: {}:{}"
+
+
+class GsShowFileAtCommitCommand(WindowCommand, GitCommand):
+
+    def run(self, commit_hash, filepath):
+        self.filepath = filepath
+        repo_path = self.repo_path
+        currentview = self.window.active_view()
+        lang = currentview.settings().get('syntax')
+        view = self.window.new_file()
+        view.set_syntax_file(lang)
+        view.settings().set("commit_hash", commit_hash)
+        view.settings().set("filepath", filepath)
+        view.settings().set("git_savvy.repo_path", repo_path)
+        view.settings().set("word_wrap", False)
+        view.settings().set("line_numbers", True)
+        view.set_name(SHOW_COMMIT_TITLE.format(commit_hash[:7], filepath[len(repo_path)+1:]))
+        view.set_scratch(True)
+        sublime.set_timeout_async(lambda: self.render_text(view), 0)
+
+    def render_text(self, view):
+        commit_hash = view.settings().get("commit_hash")
+        filepath = view.settings().get("filepath")
+        repopath = view.settings().get("git_savvy.repo_path")
+        filename = filepath[len(repopath)+1:]
+        content = self.git("show", commit_hash + ':' + filename)
+        view.run_command("gs_replace_view_text", {"text": content, "nuke_cursors": True})
--- a/core/commands/__init__.py
+++ b/core/commands/__init__.py
@@ -21,3 +21,4 @@ from .custom import *
 from .flow import *
 from .cherry_pick import *
 from .tag import *
+from .show_file_at_commit import *
--- a/core/commands/log.py
+++ b/core/commands/log.py
@@ -8,11 +8,12 @@ from ...common import util
 class GsLogCommand(WindowCommand, GitCommand):
     cherry_branch = None
 
-    def run(self, filename=None, limit=6000, author=None):
+    def run(self, filename=None, limit=6000, author=None, log_current_file=False):
         self._pagination = 0
         self._filename = filename
         self._limit = limit
         self._author = author
+        self._log_current_file = log_current_file
         sublime.set_timeout_async(self.run_async)
 
     def run_async(self):
@@ -57,6 +58,15 @@ class GsLogCommand(WindowCommand, GitCommand):
         )
 
     def on_hash_selection(self, index):
+        options_array = [
+                "Show commit",
+                "Compare commit against working directory",
+                "Compare commit against index"
+        ]
+
+        if self._log_current_file:
+            options_array.append("Show file at commit")
+
         if index == -1:
             return
         if index == self._limit:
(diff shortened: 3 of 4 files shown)


core/commands/show_commit.py
```diff
@@ -8,18 +8,23 @@
 
 class GsShowCommitCommand(WindowCommand, GitCommand):
 
-    def run(self, commit_hash):
+    def run(self, commit_hash, show_file=False, filepath=None):
```
divmain: You don't need both `show_file` and `filepath` here.  Reduce this to a single variable.

core/commands/show_commit.py
```diff
@@ -8,18 +8,23 @@
...
         view.settings().set("git_savvy.show_commit_view", True)
         view.settings().set("git_savvy.show_commit_view.commit", commit_hash)
+        view.settings().set("git_savvy.show_commit_view.filepath", filepath)
         view.settings().set("git_savvy.repo_path", repo_path)
         view.settings().set("word_wrap", False)
         view.settings().set("line_numbers", False)
         view.set_name(SHOW_COMMIT_TITLE.format(commit_hash[:7]))
         view.set_scratch(True)
-        view.run_command("gs_show_commit_initialize_view")
+        if self.show_file == True:
```
divmain: No need for `if my_var == True:` - `if my_var:` is considered to be the idiomatic pattern.

core/commands/show_commit.py
```diff
@@ -8,18 +8,23 @@
 
 class GsShowCommitCommand(WindowCommand, GitCommand):
 
-    def run(self, commit_hash):
+    def run(self, commit_hash, show_file=False, filepath=None):
+        self.show_file = show_file
         repo_path = self.repo_path
         view = self.window.new_file()
         view.set_syntax_file("Packages/GitSavvy/syntax/show_commit.tmLanguage")
```
divmain: Because of this line, the view will be syntax-highlighted as a commit (diff syntax plus a few other things) instead of the correct file-type.  For example, if you looked at the historical version of `path/to/my/file.py`, it would not be syntax-highlighted as a Python file.

Because of this, I think there is too little commonality between the original `GsShowCommitCommand` and what you're attempting to add.  Instead, create a new command that is completely distinct, `GsShowFileAtCommitCommand`.

core/commands/show_commit.py
```diff
@@ -28,3 +33,13 @@ def run(self, edit):
...
         self.view.run_command("gs_replace_view_text", {"text": content, "nuke_cursors": True})
+
+class GsShowCommitFile(TextCommand, GitCommand):
+
+    def run(self, edit):
+        commit_hash = self.view.settings().get("git_savvy.show_commit_view.commit")
+        filepath = self.view.settings().get("git_savvy.show_commit_view.filepath")
+        repopath = self.view.settings().get("git_savvy.repo_path")
+        file = filepath[len(repopath)+1:]
+        content = self.git("show", commit_hash + ':' + file)
```
divmain: This is blocking code - Sublime's UI thread will not respond while the plugin waits for Git to respond.  In this particular case, it may not take a really long while, but the user may be on a slow machine.  Instead, this should run inside the alternate thread.  Check other commands for examples of how to do this - you should see `run` methods invoking `run_async` methods with `set_timeout`.

core/commands/show_commit.py
```diff
@@ -28,3 +33,13 @@ def run(self, edit):
...
         content = self.git("show","--no-color", commit_hash)
         self.view.run_command("gs_replace_view_text", {"text": content, "nuke_cursors": True})
+
+class GsShowCommitFile(TextCommand, GitCommand):
+
+    def run(self, edit):
+        commit_hash = self.view.settings().get("git_savvy.show_commit_view.commit")
+        filepath = self.view.settings().get("git_savvy.show_commit_view.filepath")
+        repopath = self.view.settings().get("git_savvy.repo_path")
+        file = filepath[len(repopath)+1:]
```
divmain: Need a better variable name here so it is clear what it does.

divmain:
Sorry to keep you waiting - holidays got in the way!  I made a few comments - ping me when you've addressed them, and I'll give it a second pass.  Thanks for the contribution!

core/commands/show_commit.py
craigaaroncampbell: removed show_file.

core/commands/show_commit.py
craigaaroncampbell: changed to if self.filepath - no longer using self.show_file

core/commands/show_commit.py
craigaaroncampbell: changed 'file' to 'filename'

core/commands/show_commit.py
craigaaroncampbell: Moved function statements into run_async. Call run_async after a timeout of 0 ms

core/commands/show_commit.py
```diff
@@ -28,3 +29,36 @@ def run(self, edit):
...
+        view.settings().set("filepath", filepath)
+        view.settings().set("git_savvy.repo_path", repo_path)
+        view.settings().set("word_wrap", False)
+        view.settings().set("line_numbers", True)
+        view.set_name(SHOW_COMMIT_TITLE.format(commit_hash[:7]))
+        view.set_scratch(True)
+        view.run_command("gs_show_commit_file")   
+    
+
+class GsShowCommitFile(TextCommand, GitCommand):
```
divmain: I don't think there is a reason for a second command here.  Just add a second, well-named, method to the `GsShowFileAtCommitCommand` class, and `sublime.set_timeout_async(self.method_name, 0)`.

divmain:
Getting really close!  You mentioned in chat that you moved the new commands to a new module.  That's fine.  Ping me here in the PR when that change has been pushed, as well as any changes for the comment.  Thanks!

core/commands/show_file_at_commit.py
```diff
@@ -0,0 +1,39 @@
...
+        currentview = self.window.active_view()
+        lang = currentview.settings().get('syntax')
+        view = self.window.new_file()
+        view.set_syntax_file(lang)
+        view.settings().set("commit_hash", commit_hash)
+        view.settings().set("filepath", filepath)
+        view.settings().set("git_savvy.repo_path", repo_path)
+        view.settings().set("word_wrap", False)
+        view.settings().set("line_numbers", True)
+        view.set_name(SHOW_COMMIT_TITLE.format(commit_hash[:7]))
```
divmain: One other little thing.  Let's indicate which file is being viewed, as well as the revision.  Something like this.
```python
SHOW_COMMIT_TITLE = "{}:{}"

# ...
    def run(self, commit_hash, filepath):
# ...
        view.set_name(SHOW_COMMIT_TITLE.format(commit_hash[:7], filepath)
```

 replace `commit_hash[:7]` here with `

craigaaroncampbell:
I rebased into a single commit and then fixed a merge conflict to account for the new SHOW_COMMIT_TITLE = "{}:{}" change.

divmain:
This looks great.  Thanks again!

PR #342, opened 2016-01-15
Add notice for credential requirement for push in remotes.md

dharma-guardian:


divmain:
Hesitated with this a bit, because credentials may also be required for `git pull` if the repository is private.  But this addresses the more common case, for people that wouldn't already know.  So merging as is.  Thanks for the PR!

dharma-guardian:
You're welcome.

PR #344, opened 2016-01-19
check tags on current branch only

randy3k:
this allows tagging different smart-tags on stable and dev branches.

For example, the stable is tagging v1.2.3 while the dev branch, the next major version, is tagging v2.0.0-1.

This is essentially the PR #333  without the dead code.


divmain:
Thanks, this is a solid improvement.

PR #345, opened 2016-01-19
reorder smart tag order list

randy3k:
`patch` and `prerelease` are more frequently used then other options.


divmain:
That seems reasonable.  Thanks!

PR #349, opened 2016-02-10
Change all syntaxes for the new format: .sublime-syntax

gwenzek:
Hi,
This PR replace old .tmLanguage syntaxes by .sublime-syntax using the converting tool provided by Sublime (in Tools > New Syntax from xxx.tmLanguage)

I didn't modify any scope but I plan to bring some improvements in the future.
In particular I feel that the scopes name doesn't respect conventions but has been chosen to be nice on a given Theme but doesn't generalize well. (For example comments are often in italic which brokes alignement with vertical '|')
Also commit hash aren't always  bearing the same scope.

But theses changes will be in another PR.


gwenzek:
Btw, thanks for this awesome plugin !

stoivo:
I read the release note to 3103 and say the new type of syntax files. I like the new format of them. If this get merged. We wil longer support older versions of sublime. This may not be a problem for some of GitSavvy. 

This is up to @divmain. 

(I just wanted to make the note so it doesn't hit us later)

stoivo:
Do we need to update places where we set the syntax? 
like: [github/commands/pull_request.py#L124)](https://github.com/divmain/GitSavvy/blob/777d3af4526d7de94457c9a83d6d34c47a6a9b28/github/commands/pull_request.py#L124)

gwenzek:
Good catch.

I tested it without further modification and it was working. Maybe Sublime
was catching this 'error'.

My impression is that when there is a  'foo.sublime-syntax' next to a
'foo.tmLanguage', Sublime will load the first one.

But it would be cleaner to load explicitly the sublime-syntax.

pradyunsg:
I would ship both the .sublime-syntax as well as .tmLanguage files. Just do what Jon did for the default packages.

https://forum.sublimetext.com/t/dev-build-3084/14887/15

asfaltboy:
If I read correctly the thread linked by @pradyunsg, both systems are supported and there is no automatic selection (in fact, if you ship syntax with both formats, the suggestion is to add a hide tag to avoid duplicate items listed in syntax menu).

However, since GitSavvy uses scope/name and explicit references, I suggest to experiment with keeping both formats, and test on current and older ST versions to see we don't break anything.

gwenzek:
ok, I didn't think about people using older version of Sublime. I'll make another PR with both files.

But it's not anymore the behavior adopted by [Sublime Packages](https://github.com/sublimehq/Packages) which only ships the .sublime-syntax.

pradyunsg:
Well, Sublime Text Packages is meant for the latest versions.

pradyunsg:
Also, you night as well just re-add those files. And put a hide=True key into them. And squash all the current commits into one.

asfaltboy:
I'm not 100% sure hide=True would mean anything in GitSavvy, since we don't _need_ to hide the syntax file from a selection menu, but rather choose the newest supported format to render a view with.

There is also the question of maintaining 2 source files (for backward compat.), so @divmain would have to decide on that.

pradyunsg:
@asfaltboy Well, if you look at the sources, it's already being done. (I believe for all syntaxes)
Also, it makes sense because these syntaxes are highly specific to GitSavvy.

As to the question of maintaining 2 source files, I see no reason to add the `.sublime-syntax` files in GitSavvy.
The status quo works and the only compelling reason to adopt the new format is to use the greater control over the lexer stack, something not needed by GitSavvy. And the added complexity in the codebase of maintaining/choosing between the two formats.

Of course, @divmain has the deciding powers, not me.

gwenzek:
My main motivation was to improve the syntaxes, and it's really more easy to edit .sublime-syntax than .tmLanguage
This PR doesn't include any modification but I was hoping to do more in the future.

The improvements I think of are:
* uniformized scope between different view
* improve scopes for 'blame' page: the author name should be highlighted
* fix error in 'graph' syntax highlighting parenthesis inside commit as 'branch'
* improve scope for 'status' page: the current branch should be highlighted

One possibility is to add the new syntax, providing improved experience for users on ST3, and not disturbing people on ST2 (which are likely people who don't like to be disturbed)

pradyunsg:
@gwenzek I forgot that GitSavvy is a ST3-only plugin.

divmain:
Hey everybody!  Sorry for the radio silence - lots going on in real life the last couple of weeks.

Since this is a ST3-only plugin, it has long been my position to only support the latest v3 beta and onward.  Most people will want to be updating anyway, users without licenses can be on these buidls, and the releases may include bug fixes for things related to GitSavvy - one example is nested scope highlighting in the inline diff view.

For that reason, I'm comfortable moving to the new syntax format.  I do, however, want to ensure that all the programatic view syntax changes work properly, and no other functionality is effected.  I agree that the YAML format is much easier to work with :)

So @gwenzek, if you are willing to go hunting for view syntax changes and include in this PR, I'll merge it!

gwenzek:
Hi @divmain thanks for your input.
I replaced all references to '.tmLanguage' file by references to '.sublime-syntax' in 2b171925b
The plugin is still working for me without error message in the console.

I also included the change to the blame syntax (in 160f645d2dcfe79b0c47372415d861b99bb84fd1)

stoivo:
I make a new branch called `sublime_syntax`. And I will use that one for the next week. 

I will do something with the diff syntax to work with git diff and it's options.  
If you have something more you want to push into this branch you can make a PR directly to this branch so we can merge and test i.

gwenzek:
Thanks @stoivo, I added a few improvements in #362.

divmain:
How does this look @gwenzek and @stoivo?  If you've been using it with no issues, we should merge this in soon.  It may result in a major-revision bump of GitSavvy's version number, since this will be a breaking change for older ST3 versions.  But if it is working, let's move it into master.

stoivo:
I think we are where we want to be. I think there is now bug here. But I  don't use the rebase and tag dashboard more than I do for GitSavvy. So only a few times.

I will have a PR from sublime_syntax to master.

PR #354, opened 2016-02-16
Feature: resolve conflict (stage merged file) on successful merge

asfaltboy:
* check helper methods return value and abort merge on fails
* store temp. backup of merge target file and revert to it on merge fail
* if mergetool.<tool>.trustExitCode is true simply resolve
* otherwise, resolve only if file changed from original
  (git would normally prompt if unchanged as well, we simply revert)


divmain:
Great - thanks!

PR #358, opened 2016-02-24
Avoid printing trailing spaces from blame output

dnicolson:
This pull request changes the blame command to not output unnecessary trailing spaces, if the [Trailing Spaces](https://github.com/SublimeText/TrailingSpaces) package is installed it is quite noticeable.

Before:
before

After:
after

A superfluous assignment of the variable `total_lines` is also removed.


divmain:
Looks good to me.  Thanks!

ddevlin:
@dnicolson, as a Trailing Spaces user, :heart: this PR, thanks!

PR #359, opened 2016-02-25
Syntax highlighting for blame output

dnicolson:
This enables syntax highlighting for line separators of commit messages spanning 40-48 characters to match what `blame.py` outputs.

For example in commit 40cdf58 the maximum commit line length of `core/git_command.py` is 48 characters and for `core/commands/blame.py` is 40 characters. This causes line separators to be either white or grey (which both look okay).

### DIFF
--- a/syntax/blame.YAML-tmLanguage
+++ b/syntax/blame.YAML-tmLanguage
@@ -14,4 +14,4 @@ patterns:
     '3': { name: keyword.other.git-savvy.line-number }
 - comment: separator
   name: comment.block.git-savvy.separator
-  match: ^\-{40} \| \-+$
+  match: ^\-{40,48} \| \-+$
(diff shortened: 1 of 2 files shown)


divmain:
@dnicolson, this looks reasonable, so I'll merge it in.  But having trouble visualizing what this changes.  If you get some time, would you mind posting a screen-shot here for posterity?

divmain:
P.S.  For something like this, if you prepend the commit summary with `Enhancement:`, it helps with GitSavvy's changelog auto-generattion (`git: generate changelog`).

dnicolson:
Sure, I've composed a left/right before/after with `git_command.py` and `blame.py`:



Before:
post

After:
pre

divmain:
Got it - thanks!

PR #362, opened 2016-03-02
Improve graph and blame syntaxes

gwenzek:
extension of #349

### DIFF
--- a/syntax/blame.sublime-syntax
+++ b/syntax/blame.sublime-syntax
@@ -6,13 +6,22 @@ hidden: true
 scope: git-savvy.blame
 contexts:
   main:
-    - match: '^([^|]+) (\|) (( |\d){1,4})'
-      comment: line
-      scope: meta.git-savvy.blame-line
-      captures:
-        1: comment.block.git-savvy.commit-info
-        2: comment.block.git-savvy.splitter
-        3: keyword.other.git-savvy.line-number
-    - match: '^\-{40,48} \| \-+$'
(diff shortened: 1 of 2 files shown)


stoivo:
This is good. We should also write some tests for our syntaxes.
There is a paragraph about it under Testing https://www.sublimetext.com/docs/3/syntax.html.

Links to some examples
https://github.com/sublimehq/Packages/blob/master/Python/syntax_test_python.py
https://github.com/sublimehq/Packages/blob/master/HTML/syntax_test_html.html
https://github.com/sublimehq/Packages/blob/master/JavaScript/syntax_test_js.js
https://github.com/sublimehq/Packages/blob/master/SQL/syntax_test_sql.sql
https://github.com/sublimehq/Packages/blob/master/Ruby/syntax_test_ruby.rb

stoivo:
Grach looks good, but can you add a screenshot of who the syntax look?

gwenzek:
Hi @stoivo, thanks for testing my changes.
I added a few tests in gwenzek/develop as you requested. I only did tests for graph/blame/branch for now. The others are coming soon.
Should I make another PR with the tests?  The simplest would be for you to checkout my repo on your computer for testing instead of creating PR each times.

stoivo:
I merged your develop branch into sublime_syntax. You tests looks good.
Make a new PR for the other tests. 

We will probably make a new release with all this syntax fixes before we get to write tests for the other syntaxes.

PR #363, opened 2016-03-03
WIP: GitLab support

divmain:
See #364 for summary.


stoivo:
We'll make it small step at the time.

Issue #81, opened 2015-03-12
local variable 'stdout' referenced before assignment


alextegelid:
Appears in console when a new window is opened since updating to ver 1.2.0

divmain:
@alextegelid - sorry about that, 1.2.1 will fix the issue once its picked up by package control.

alextegelid:
Thanks, I'll pull it from GitHub right away!

divmain:
There was a typo, so use 1.2.2.  Thanks for giving me the heads-up.

alextegelid:
Will do, thanks!

PR #365, opened 2016-03-04
show git status in status-bar when there is no view.

### LINKED ISSUE #356: Status bar should show status of the project directory when no view is opened.
Right now, if no file is opened, the status bar shows nothing. It only pops up if a file is opened. It would be nice to show info about the first project directory if no file is opened.

I did a quick look of the code, it seems that the current code depends heavily on `self.view`, which may make this feature hard to be implemented.

randy3k:
Status bar is erased when there is no view because `self.file_path` is empty. In fact, there is no need to check `self.file_path` because `self._repo_path` already checks that the directory is a git repo.

There is also side effect, not sure if it is a good/bad thing. Status-bar is now shown also in the dashboard pages.

fix #356

### DIFF
--- a/core/commands/status_bar.py
+++ b/core/commands/status_bar.py
@@ -31,7 +31,7 @@ class GsUpdateStatusBarCommand(TextCommand, GitCommand):
 
     def run_async(self):
         # Short-circuit update attempts for files not part of Git repo.
-        if not self.file_path or not self._repo_path(throw_on_stderr=False):
+        if not self._repo_path(throw_on_stderr=False):
             self.view.erase_status("gitsavvy-repo-status")
             return
 


stoivo:
Sorry for making you wait!

I looked at the log for the current file and I found this commit 
[40f42cf982baca83f73502c4df66d424d91cc312](https://github.com/divmain/GitSavvy/commit/40f42cf982baca83f73502c4df66d424d91cc312)
https://github.com/divmain/GitSavvy/issues/81
```diff
diff --git a/core/commands/status_bar.py b/core/commands/status_bar.py
index 45ebdc3..0883a4f 100644
--- a/core/commands/status_bar.py
+++ b/core/commands/status_bar.py
@@ -31,7 +31,7 @@ class GsUpdateStatusBarCommand(TextCommand, GitCommand):
 
     def run_async(self):
         # Short-circuit update attempts for files not part of Git repo.
-        if not self._repo_path(throw_on_stderr=False):
+        if not self.file_path or self._repo_path(throw_on_stderr=False):
             self.view.erase_status("gitsavvy-repo-status")
             return
```

This is almost a revert of your commit. I don't really get the issue on #81. Maybe @alextegelid or @divmain can explain it here so we can test if that works.

alextegelid:
I'm not that familiar with the actual code in #81, I just reported what seemed to be an error and that was enough for @divmain to find and fix the issue

divmain:
I pulled down the change, and whatever error occurred before no longer seems to be an issue.  I'll merge this in and see if any issues surface with it before pushing out a release.

alextegelid:
Sweet! I'll pull this down and keep my eyes open

randy3k:
Thanks @divmain

PR #367, opened 2016-03-09
Goto symbol

stoivo:
Feature: Use `Goto Symbol` to navigate in all views

We use .tmPreferences to define witch scopes that should be added to SymbolList, by using a custom scope like `gitsavvy.totosymbol` it can not match any other scopes unless the is a theme defines this scope. So all view will look the same way that are doing now.


divmain:
This looks great!  I'll merge it in.  How does the `sublime_syntax` branch look, @stoivo?  Have you noticed any problems?  I'd prefer to merge that in soon, if it seems to be working well.  I'd prefer to avoid long-running branches if possible.

PR #368, opened 2016-03-13
Sublime syntax

stoivo:
A follow up on #349.


common/util/file.py
```diff
@@ -24,6 +24,7 @@ def determine_syntax_files():
 def get_syntax_for_file(filename):
     extension = get_file_extension(filename)
     syntaxes = syntax_file_map.get(filename, None) or syntax_file_map.get(extension, None)
+    # watch out if this one has changed https://github.com/sublimehq/Packages/tree/master/Text
```
divmain: This is unnecessary - if the package changes to the new `sublime-syntax`, we will know it sooner by a new bug than by running into this comment.  If you feel we need a reminder, let's open an issue, indicating that we should review all the official syntax packages in a month or two.

syntax/diff.sublime-syntax
```diff
@@ -0,0 +1,62 @@
...
+      |Index:[ ]
+      |---\ [^%]
+      |\*\*\*.*\d{4}\s*$
+      |\d+(,\d+)* (a|d|c) \d+(,\d+)* $
+      |diff\ --git[ ]
+      )
+contexts:
+  main:
+    - match: '(^(-{3}) a/(.+)$\n?)'
+      comment: Needs to be first.
```
divmain: This comment should be improved or removed.  It doesn't tell me why it needs to be first.

divmain:
This is good, and I especially like that we can test the syntaxes now!  A couple of things before this is merged:

- Please do a pass over any comments, and add/remove as needed.  In particular, use comments to clarify what something is (for example, a non-intuitive regular expression) or why something needs to be a certain way.
- We'll need to clean up the commit history.  It looks like `master` was merged into the branch at one point, which may complicate things a bit.  If you need help with this, let me know and I'll lend a hand.

syntax/diff.sublime-syntax
stoivo: Good catch!

common/util/file.py
stoivo: It was meat like a comment to the developers that may search for '.tmLanguage' and replacing those with `.sublime-syntax`.

It will not change to sublime-syntax in near future to keep backward compatibility.

stoivo:
I did some rebasing now. Is this what you want or should I squash some of the commits.

common/util/file.py
stoivo: What do you think?

stoivo:
Show have some updated comments now.

common/util/file.py
divmain: Oh, I understand now what you intended.   Still, I think it is unnecessary.  After this PR, it isn't very likely that anyone will go looking for `.tmLanguage` in the codebase anyway.  Basically: 6 months from now, I won't know why this comment is here.  And since it won't remove confusion, I say we remove the comment.

syntax/diff.sublime-syntax
```diff
@@ -0,0 +1,66 @@
...
+      |==== \s* // .+ \s - \s .+ \s+ ====
+      |Index:[ ]
+      |---\ [^%]
+      |\*\*\*.*\d{4}\s*$
+      |\d+(,\d+)* (a|d|c) \d+(,\d+)* $
+      |diff\ --git[ ]
+      )
+contexts:
+    - match: '(^(-{3}) a/(.+)$\n?)'
+      comment:
```
divmain: I think YAML expects multi-line values to have something here denoting as much.  I'm guessing that nothing broke with this comment, otherwise you would have fixed it.  But could you change this to `      comment: |`, just so it is consistent?

divmain:
This is really great.  I just have one small piece of feedback.  That can be fixed before I merge, or we can merge it now and I'll fix it afterwards.  Thanks for all the work that went into this!

stoivo:
Merge? @divmain

divmain:
Done.

stoivo:
Thanks!

PR #372, opened 2016-03-15
Be vintageous friendly - an idea

mreq:
Would you mind adding vintageous support in this way?

It's a not-100% working example binding "j" and "k" to navigate the status page.

Basically there'd be two sets of keybinds and page-templates based on a single settings switch.

I'd be able to finish this PR to be merge-able if you see this as an option.

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -964,6 +964,14 @@
         ]
     },
     {
+        "keys": ["o"],
+        "command": "gs_vintageous_enter_insert_mode",
+        "context": [
+            { "key": "setting.git_savvy.vintageous_friendly", "operator": "equal", "operand": true },
+            { "key": "vi_command_mode_aware" }
+        ]
+    },
+    {
         "keys": ["escape"],
         "command": "gs_vintageous_enter_normal_mode",
         "context": [
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -8,6 +8,7 @@
         "keys": ["l"],
         "command": "gs_inline_diff_stage_or_reset_line",
         "context": [
+            { "key": "setting.git_savvy.handle_keys", "operator": "equal", "operand": true },
             { "key": "setting.git_savvy.inline_diff_view", "operator": "equal", "operand": true }
         ]
     },
@@ -15,6 +16,7 @@
         "keys": ["h"],
         "command": "gs_inline_diff_stage_or_reset_hunk",
         "context": [
+            { "key": "setting.git_savvy.handle_keys", "operator": "equal", "operand": true },
             { "key": "setting.git_savvy.inline_diff_view", "operator": "equal", "operand": true }
         ]
     },
@@ -23,6 +25,7 @@
         "command": "gs_inline_diff_stage_or_reset_line",
         "args": { "reset": true },
         "context": [
+            { "key": "setting.git_savvy.handle_keys", "operator": "equal", "operand": true },
             { "key": "setting.git_savvy.inline_diff_view", "operator": "equal", "operand": true }
         ]
     },
@@ -31,6 +34,7 @@
         "command": "gs_inline_diff_stage_or_reset_hunk",
         "args": { "reset": true },
         "context": [
+            { "key": "setting.git_savvy.handle_keys", "operator": "equal", "operand": true },
             { "key": "setting.git_savvy.inline_diff_view", "operator": "equal", "operand": true }
         ]
     },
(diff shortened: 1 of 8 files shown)


divmain:
Hi @mreq - definitely willing to add this, so long as it is a transparent change to users that don't use vintageous.

Default.sublime-keymap
```diff
@@ -308,6 +317,15 @@
...
     {
+        "keys": ["k"],
+        "command": "gs_status_navigate_file",
+        "args": { "forward": false },
+        "context": [
+            { "key": "setting.git_savvy.status_view", "operator": "equal", "operand": true },
+            { "key": "setting.git_savvy.vintageous_friendly", "operator": "equal", "operand": true }
+        ]
+    },
+    {
```
divmain: Does Vintagious use [command mode](http://docs.sublimetext.info/en/latest/reference/key_bindings.html#command-mode)?  That might be a more robust determinant for these key-binding contexts.

core/interfaces/status.py
```diff
@@ -31,56 +31,108 @@ class StatusInterface(ui.Interface, GitCommand):
...
+
+          ###########
+          ## OTHER ##
+          ###########
+
+          [r]         refresh status
+          [tab]       transition to next dashboard
+          [SHIFT-tab] transition to previous dashboard
+          [.]         move cursor to next file
+          [,]         move cursor to previous file
```
divmain: Keep this template string as-is, but accept new template options here and in the line above.  Something like:
```
          [{cursor_next_key}]         move cursor to next file
          [{cursor_prev_key}]         move cursor to previous file
```
And then evaluate those based on `sublime.load_settings("GitSavvy.sublime-settings").get("vintageous_friendly")` when the template string is evaluated.

core/interfaces/status.py
```diff
@@ -31,56 +31,108 @@ class StatusInterface(ui.Interface, GitCommand):
...
-
-      [r]         refresh status
-      [tab]       transition to next dashboard
-      [SHIFT-tab] transition to previous dashboard
-      [.]         move cursor to next file
-      [,]         move cursor to previous file
-
-    -
-    """
+    if sublime.load_settings("GitSavvy.sublime-settings").get('vintageous_friendly', False):
```
divmain: Try to use double-quote marks unless you'd have to escape a double-quote in the string.

mreq:
That's definitely a key point - it must affect vintageous users only. I'll keep adding commits and mention you once I find it ready.

Default.sublime-keymap
mreq: It does, can be overriden with `"context": [{ "key": "vi_command_mode_aware" }]`. As I mentioned, this commit was made as an idea. I'll handle command mode properly now, since I know  that you're willing to merge vintagoues support.

core/interfaces/status.py
mreq: Yeah, that'd be better probably.

divmain:
Thanks much - looking forward to this PR!

mreq:
Using a `handle_keys` switch seems like a clean solution. Works well.

mreq:
Am actually happy with the way it works now. Care to take a look @divmain? Thanks

stoivo:
@mreq, Do you think this is ready?

divmain:
Been thinking about this for the past week or so.  I'm hesitant to put in something that integrates narrowly with Vintageous, and would prefer a more generalized solution.  Have you looked into adding [command_mode](http://docs.sublimetext.info/en/latest/reference/key_bindings.html#command-mode) context to the key-bindings versus the specialized context that you add dynamically?

mreq:
I've been using my fork exclusively for the past few weeks and it indeed works nice. There's still a small issue with bindings in one of the views (log I believe) that I have to address though.

@divmain Definitely up to you. The idea behind this is to still be able to use all the power of vintageous. If  you were to add command_mode_aware keybinds you would interfere with that. This solution doesn't cause any difference to non-vintageous users (which is a key point). What do you mean by generalized there?

divmain:
What I'm wondering is, instead of doing this:
```
{ "key": "setting.git_savvy.handle_keys", "operator": "equal", "operand": true }
```
If this could be made to work:
```
{ "key": "setting.command_mode", "operator": "equal", "operand": false }
```

If Vintageous correctly utilizes this Sublime feature, it seems like setting `handle_keys` would be unnecessary.  This would be more idiomatic, and the changes in the key-binding contexts would improve GitSavvy compatibility with other plugins (which is what I meant by generalized).

mreq:
continued in #395

PR #374, opened 2016-03-22
Improve `create pull request` and rename the function `get_remote_branches`

randy3k:
I have renamed `list_remote_branches` to `get_remote_branches` and upgraded the related code 4e3f012.

> `get_branches` and `get_remote_branches` have similar names but have very different outputs.
`get_branches` returns a named tuple `Branch` and `get_remote_branches` returns a list of remote branches names. In order to reduce confusion, `get_remote_branches` is renamed
as `list_remote_branches`. `list_remote_branches` also has a new argument `remote` which
filters the results with a specified remote.

Also, `create pull request` now checks if remote tracking branch has been set.

PS: `on_select_remote` and `on_select_branch` have been used several times, but I couldn't find a easy way to reuse the code.

### DIFF
--- a/core/commands/checkout.py
+++ b/core/commands/checkout.py
@@ -82,7 +82,7 @@ class GsCheckoutRemoteBranchCommand(WindowCommand, GitCommand):
         sublime.set_timeout_async(self.run_async)
 
     def run_async(self):
-        self.remote_branches = self.get_remote_branches()
+        self.remote_branches = self.list_remote_branches()
         self.window.show_quick_panel(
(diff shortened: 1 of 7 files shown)


divmain:
For future PRs, if you could remember to use our [preferred commit message structure](https://github.com/divmain/GitSavvy/blob/master/CONTRIBUTE.md#commit-message-structure), it really helps when we generate the changelogs.  Thanks for the PR!

randy3k:
Got it. Thanks for considering the PR anyway. :+1:

PR #375, opened 2016-03-22
use active remote if empty

randy3k:
`last_remote_used ` would return the active tracking remote if last remote is not set.


divmain:
Seems reasonable.  Thanks again!

PR #392, opened 2016-04-18
Feature: Cherry-pick from Branch Commit Comparison view

### LINKED ISSUE #11: `git: cherry pick`

stoivo:
closes #11

Some of the changes is so Branch Commit Comparison view can be refreshed.

Read my comment on core/commands/cherry_pick.py L28.
Maybe I can rewrite the cherrypick to be defined in BranchesMixin instead, and call that one. This way I we don't need to call the asynk call.

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -650,7 +650,7 @@
     },
     {
         "keys": ["H"],
-        "command": "gs_branches_diff_branch_history",
+        "command": "gs_branches_diff_commit_history",
         "context": [
             { "key": "setting.command_mode", "operator": "equal", "operand": false },
             { "key": "setting.git_savvy.branch_view", "operator": "equal", "operand": true }
@@ -829,6 +829,48 @@
         "context": [
             { "key": "setting.command_mode", "operator": "equal", "operand": false },
             { "key": "setting.git_savvy.log_graph_view", "operator": "equal", "operand": true }
+        ]
+    },
+
+    //////////////////////////////
+    // BRANCH COMMIT COMPARISON //
+    //////////////////////////////
+
+    {
+        "keys": ["enter"],
+        "command": "gs_branches_diff_commit_history_action",
+        "context": [
+            { "key": "setting.git_savvy.branch_commit_history_view", "operator": "equal", "operand": true }
+        ]
+    },
+    {
+        "keys": ["."],
+        "command": "gs_log_graph_next_commit",
+        "args": {"forward": true},
+        "context": [
+            { "key": "setting.git_savvy.branch_commit_history_view", "operator": "equal", "operand": true }
+        ]
+    },
+    {
+        "keys": [","],
+        "command": "gs_log_graph_next_commit",
+        "args": {"forward": false},
+        "context": [
+            { "key": "setting.git_savvy.branch_commit_history_view", "operator": "equal", "operand": true }
+        ]
+    },
+        {
+        "keys": ["m"],
+        "command": "gs_log_graph_toggle_more_info",
+        "context": [
+            { "key": "setting.git_savvy.branch_commit_history_view", "operator": "equal", "operand": true }
+        ]
+    },
+    {
+        "keys": ["r"],
+        "command": "gs_branches_diff_commit_history_refresh",
+        "context": [
+            { "key": "setting.git_savvy.branch_commit_history_view", "operator": "equal", "operand": true }
         ]
     },
 
--- a/core/commands/cherry_pick.py
+++ b/core/commands/cherry_pick.py
@@ -1,13 +1,15 @@
 import sublime
+from sublime_plugin import TextCommand
 
 from .log import GsLogCommand
 from ..git_mixins.branches import BranchesMixin
+from ...common import util
 
 
 class GsCherryPickCommand(GsLogCommand, BranchesMixin):
-    def run_async(self, target_hash=None):
-        if target_hash:
-            return self.cherry_pick(target_hash)
+    def run_async(self):
+        if self._target_hash:
+            return self.cherry_pick(self._target_hash)
 
         self.select_commit = super().run_async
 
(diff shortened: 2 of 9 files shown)


Default.sublime-keymap
```diff
@@ -673,6 +673,48 @@
...
+        "keys": ["enter"],
+        "command": "gs_log_graph_action",
+        "context": [
+            { "key": "setting.git_savvy.log_graph_history_view", "operator": "equal", "operand": true }
+        ],
+        "args": { "action": "view" }
+    },
+    {
+        "keys": ["r"],
+        "command": "gs_branches_diff_branch_history_refrech",
```
divmain: Typo?  Should be `_refresh`.

core/commands/cherry_pick.py
```diff
@@ -1,13 +1,40 @@
 import sublime
+from sublime_plugin import TextCommand
 
 from .log import GsLogCommand
 from ..git_mixins.branches import BranchesMixin
+from ...common import util
+
+
+class GsCherryPickHashCommand(TextCommand, BranchesMixin):
```
divmain: If this command relies on how text is laid out in the branch history view, this class should be placed in that file.

core/commands/log.py
```diff
@@ -9,12 +9,13 @@
...
     cherry_branch = None
 
-    def run(self, filename=None, limit=6000, author=None, log_current_file=False):
+    def run(self, filename=None, limit=6000, author=None, log_current_file=False, target_hash=None):
         self._pagination = 0
         self._filename = filename
         self._limit = limit
         self._author = author
         self._log_current_file = log_current_file
+        self._target_hash = target_hash
```
divmain: What is this used for?

docs/branch_mgmt.md
```diff
@@ -88,7 +88,7 @@ A scratch view will be opened, showing the diff between the selected branch and
 
 #### Diff history of selected and active branches (`H`)
 
-A scratch view will be opened, showing the commit diff between the selected branch and the active branch.
+A scratch view will be opened, showing the commit diff between the selected branch and the active branch. To cherry-pick a commit, move one courser over the commit you want to cherry-pick and press `c`. The commit might not appear at once then refresh the view  by pressing `r`.
```
divmain: This is probably okay, but as I have mentioned before, I have a preference for discoverable UX.  Since there is no dashboard to remind users of the key-bindings here, how would you feel about them pressing `Enter` or `Super+Enter`, and then being asked what they want to do with that commit?  It could be:

- view commit
- checkout commit
- cherry-pick commit
- ??? maybe something else

core/commands/log.py
stoivo: As you can see  in cherry pick command we use _target_hash if it is in use. 
```python 
class GsCherryPickCommand(GsLogCommand, BranchesMixin):
    def run_async(self):
        if self._target_hash:
            return self.cherry_pick(self._target_hash)
```
This class inherit from GsLogCommand. I think the way it was written was impossible to set self._target_hash.

docs/branch_mgmt.md
stoivo: than it will be easier to find for everyone. I don't think there is anyone that would hate us if the change the keybindings other then for any of the interfaces.

Should that we a separate PR of in this one. I will do it for the graph view too.

core/commands/cherry_pick.py
stoivo: That is in core/interfaces/branch.py, should I move it?

core/commands/cherry_pick.py
```diff
@@ -1,13 +1,40 @@
...
+
+        if not len(hashes) == 1:
+            sublime.status_message("You can only cherry-pick one commit at a time.")
+            return
+
+        self.view.window().run_command("gs_cherry_pick", {"target_hash": hashes[0]})
+        # This line doest always works correct. We want to refresh the
+        # view after cherry-picking it, but the cherry-pick command is
+        # async so this method get called before the cherry-pick is done
+        self.view.run_command("gs_branches_diff_branch_history_refrech")
```
stoivo: What is your idea here to solve this? Should I add it to BranchesMixin?

core/commands/cherry_pick.py
divmain: Hmm... so actually, I'd like for the branch history stuff to be extracted from `interfaces/branch.py`.  The history view may be triggered from the branch dashboard, but that doesn't mean it can only be triggered from there, and it isn't really related to the branch dashboard directly.

If you'd like, I can extract that first, and then you can move `GsCherryPickHashCommand` to that new file.

core/commands/log.py
divmain: Ah, I missed that it inherited from `GsLogCommand`.  I suppose that makes sense - never mind!

core/commands/cherry_pick.py
divmain: Here's what we should do:

- This line should be removed.
- Then, [at the end of `GsCherryPickCommand`](https://github.com/divmain/GitSavvy/blob/be3f6afc3d8955567b3842787fe769533cb43ad8/core/commands/cherry_pick.py#L39-L39), we should add [`util.view.refresh_gitsavvy(self.window.active_view())`](https://github.com/divmain/GitSavvy/blob/be3f6afc3d8955567b3842787fe769533cb43ad8/core/commands/pull.py#L92-L92).
- Finally, modify [`refresh_gitsavvy`](https://github.com/divmain/GitSavvy/blob/be3f6afc3d8955567b3842787fe769533cb43ad8/common/util/view.py#L61-L68) so that it checks whether the active view is a branch history view and, if so, it should run `gs_branches_diff_branch_history_refresh`.

core/commands/cherry_pick.py
stoivo: I can do it, just give me some time :)

docs/branch_mgmt.md
stoivo: Should `r` be refresh of an option in dropdown?
I think it should be on `r`.

core/commands/cherry_pick.py
stoivo: Note to self: Done

core/commands/cherry_pick.py
stoivo: Done

stoivo:
@divmain, Sorry for all this waiting. Can you take a look at this now?

divmain:
This looks fantastic!  Merged!

PR #395, opened 2016-04-20
Be vintageous friendly

mreq:
continued from  #372

### DIFF
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -64,10 +64,7 @@ class GsCommitCommand(WindowCommand, GitCommand):
         else:
             view.set_syntax_file("Packages/GitSavvy/syntax/make_commit.tmLanguage")
 
-        if savvy_settings.get("vintageous_friendly", False) is True:
-            view.settings().set("git_savvy.vintageous_friendly", True)
-            if savvy_settings.get("vintageous_enter_insert_mode", False) is True:
-                view.run_command("_enter_insert_mode")
+        view.run_command("gs_handle_vintageous")
 
         commit_on_close = savvy_settings.get("commit_on_close")
         view.settings().set("git_savvy.commit_on_close", commit_on_close)
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -64,6 +64,11 @@ class GsCommitCommand(WindowCommand, GitCommand):
         else:
             view.set_syntax_file("Packages/GitSavvy/syntax/make_commit.tmLanguage")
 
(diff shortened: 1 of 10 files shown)


mreq:
Somehow this didn't work when I first tried it. It does, however, work now and is definitely a better approach.

`{ "key": "setting.command_mode", "operator": "equal", "operand": false }` is enough

core/commands/commit.py
```diff
@@ -64,6 +64,11 @@ def run_async(self, repo_path=None, include_unstaged=False, amend=False):
         else:
             view.set_syntax_file("Packages/GitSavvy/syntax/make_commit.tmLanguage")
 
+        if savvy_settings.get("vintageous_friendly", False) is True:
```
divmain: Can you add a comment block above this line, explaining why it is here and what it does?  Add the same for `log_graph.py` and `ui.py`.

divmain:
Awesome, I like this much better.  Once you add the comments I requested, I'll merge this in!  I may revisit at some point to see if this can be further abstracted, so that Vintageous-specific settings become unnecessary.  But this seems a reasonable step for now.  Thanks for your work on this!

mreq:
I've created a GsHandleVintageousCommand which has it's own docstring and keeps the code DRY. Hopefully placed it right.

divmain:
That's perfect.  Thanks again!

asfaltboy:
This PR actually causes a bit of pain for me. Since the default mode when opening a view is command mode, to use GitSavvy shortcuts I must change to insert. Even changing between dashboards using TAB requires changing to insert mode every time. Honestly, even though the previous mode was lacking some vim-friendliness , at least it required much less keystrokes.

mreq:
@asfaltboy set the `vintageous_enter_insert_mode` option to `true`

asfaltboy:
Brilliant! Why won't we set [these to True by default][1]?

[1]: https://github.com/divmain/GitSavvy/blob/master/common/commands/view_manipulation.py#L68-L70

mreq:
Indifferent to me. Might be a good idea, as it mimics the default non-vintageous behaviour. What do you think, @divmain?

asfaltboy:
We should definitely document this feature if possible

mreq:
Well, it is explained in the [default settings](https://github.com/divmain/GitSavvy/blob/master/GitSavvy.sublime-settings#L258) file.

asfaltboy:
Yes, this is why I just opened issue #430 .

PR #400, opened 2016-04-22
Enhancement: Remember which option got selected last in quick panel

### LINKED ISSUE #394: blame current file: remember last blame mode
I always use the last blame mode (detect moved or copied across all commits). Would be handy if it was auto-selected every time.

Many of the sub-menus remember the last position, so I suspect this is low-hanging and generally desirable. 

For your consideration. Thanks for all your work so far! 😄

stoivo:
fix #394

### DIFF
--- a/core/commands/blame.py
+++ b/core/commands/blame.py
@@ -32,13 +32,15 @@ class GsBlameCommand(WindowCommand, GitCommand):
                 "Detect moved or copied lines within same commit",
                 "Detect moved or copied lines across all commits",
             ],
-            self.on_option_selection
+            self.on_option_selection,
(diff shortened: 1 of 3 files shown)


divmain:
Would prefer to use the same pattern as [here](https://github.com/divmain/GitSavvy/blob/3d89abab3f3304120549176a461c816837278321/core/git_command.py#L303).  This doesn't remember selections when you close and re-open Sublime, but that is mostly a non-issue and it keeps Sublime settings less cluttered.

stoivo:
This is nicer, a lot nicer. I didn't know about this.

divmain:
Excellent!

PR #403, opened 2016-04-30
Improve: the slow startup time

randy3k:
Detail: `util.file.determine_syntax_files` has been significantly slowing down the start up time of Sublime Text (at least 6 seconds on my computer).

Solution: use `set_timeout_async` to run it.


stoivo:
This looks good.

PR #404, opened 2016-05-07
Navigate with `,` and `.` in Tag and Rebase dashboard

stoivo:


bharath144:
Can this be done on the BRANCH view as well? It makes sense to navigate different branches (local and remote) with `.` and `,`

stoivo:
@bharath144, that was already added but it is not released yet.

PR #407, opened 2016-05-09
Ensure Unicode is not decomposed

dnicolson:
Git commit names can contain decomposed Unicode, which causes line length calculation to be incorrect for the output of the blame command. In the below example the "ö" in "Schöne decomposed" is `\x6F\xCC\x88` but in "Schöne original" it is `\xC3\xB6`.

The solution is to normalize the Unicode of the output from the git blame command.

Before:
before

After:
after

### DIFF
--- a/core/commands/blame.py
+++ b/core/commands/blame.py
@@ -1,5 +1,6 @@
 import re
 from collections import namedtuple, defaultdict
+import unicodedata
 
 import sublime
 from sublime_plugin import WindowCommand, TextCommand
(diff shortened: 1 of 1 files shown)


stoivo:
I will test this myself next couple of days. This look good.

dnicolson:
This is the repo I used for testing: https://github.com/dnicolson/Decomposed-Unicode-Example.

divmain:
This looks great - thanks!

stoivo:
@divmain, Did you check the different platforms? Just so we don't do the same mistake I did. :)

divmain:
@stoivo, you generally shouldn't need to do that until you're about to release a version.  Otherwise it is a lot of overhead for each PR to pull up a Mac, Windows, and Linux box to check for regressions.

stoivo:
ok, I check the same way.

Issue #321, opened 2015-12-13
git: 'flow' is not a git command


rimmer333:
I have a strange issue with git flow commands in GitSavvy. I have git-flow installed (on a Mac from Homebrew), and I use it from command line all the time. I have enabled flow commands in GitSavvy, but whenever I call one of those from command drawer in Sublime upon an already flow-enabled repo, it ends up with this in a popping console on the bottom:
```
`/usr/bin/git flow version` failed with following output:

git: 'flow' is not a git command. See 'git --help'.

Did you mean one of these?
	reflog
	show

```
, as if I didn't have git-flow. Calling this same `/usr/bin/git flow version` from command line results in a perfectly legit
```
0.4.1
```

I don't run Sublime as a different user or root, I am a single user on this Macbook. `ls -lha /usr/local/bin/git-flow` results in a very usual
```
lrwxr-xr-x  1 myusername  admin    37B 23 июл  2014 /usr/local/bin/git-flow -> ../Cellar/git-flow/0.4.1/bin/git-flow
```

So, I'm clueless as to what could be wrong here.

divmain:
Have you installed a new version of `git` with Homebrew?  If so, it is likely that Sublime is using the Git packaged with OS X, whereas git flow is setup for the Git installed by Homebrew.  The easiest way to resolve this is to modify the `git_path` in `GitSavvy.sublime-settings`.  Give this a try, and let me know what happens.

rimmer333:
Man, you're quick to react! Unfortunately, that is not the case. I don't have git installed as a Homebrew package, and I always had the system git path set in my settings:

```
Last login: Sun Dec 13 21:15:06 on console
~ » which git
/usr/bin/git
~ » brew list
ansible			curlftpfs		ffmpeg			gettext			haxe			jpeg			libgpg-error		libtool			multimarkdown		objective-caml		pcre			rbenv			tmux			x264
autoconf		elixir			fontconfig		git-flow		help2man		lame			libimobiledevice	libusb			mysql			openssl			phantomjs		readline		tor			xvid
automake		eot-utils		fontforge		glib			hub			legit			libplist		libxml2			neko			ossp-uuid		pidof			ruby-build		unixodbc		xz
bdw-gc			erlang			freetype		gmp			icu4c			libevent		libpng			libyaml			nettle			osxfuse			pixman			spidermonkey		usbmuxd			yasm
[...]

PR #410, opened 2016-05-16
Documentation: Add note on `git-flow` path discoverability

asfaltboy:
To help users get over issues such as #321.


stoivo:
Thanks.

PR #429, opened 2016-06-10
Feature/custom environ for custom commands

asfaltboy:
This is a pretty selfish feature. I found myself running rebase in shell for squashing fixup commits, e.g `git rebase -i --autosquash origin` and just found out that I could do this non-interactively if setting a non-interactive EDITOR environment variable.

I wanted to add a custom commands and found I couldn't. Here's an example custom command:

    {
        "caption": "git: rebase fixup commits",
        "command": "gs_custom",
        "args": {
            "output_to_panel": true,
            "args": ["rebase", "-p", "-i", "--autosquash", "{PROMPT_ARG}"],
            "custom_environ": {"EDITOR": "cat"},
            "prompt_msg": "Rebase from commit: "
        }
    }

### DIFF
--- a/core/commands/custom.py
+++ b/core/commands/custom.py
@@ -6,18 +6,17 @@ from ..git_command import GitCommand
 from ...common import util
 
 
-ALL_REMOTES = "All remotes."
-
-
 class CustomCommandThread(threading.Thread):
-    def __init__(self, func, *args, **kwargs):
+    def __init__(self, func, *args, custom_environ=None, **kwargs):
+        self.custom_environ = custom_environ
         super(CustomCommandThread, self).__init__(**kwargs)
         self.cmd_args = args
         self.cmd_func = func
         self.daemon = True
 
     def run(self):
-        return self.cmd_func(*self.cmd_args)
+        return self.cmd_func(*self.cmd_args,
+                             custom_environ=self.custom_environ)
 
 
 class GsCustomCommand(WindowCommand, GitCommand):
@@ -52,7 +51,8 @@ class GsCustomCommand(WindowCommand, GitCommand):
                   start_msg="Starting custom command...",
                   complete_msg="Completed custom command.",
                   run_in_thread=False,
-                  custom_argument=None):
+                  custom_argument=None,
+                  custom_environ=None):
 
         for idx, arg in enumerate(args):
             if arg == "{REPO_PATH}":
@@ -65,10 +65,10 @@ class GsCustomCommand(WindowCommand, GitCommand):
         sublime.status_message(start_msg)
         if run_in_thread:
             stdout = ''
-            cmd_thread = CustomCommandThread(self.git, *args)
+            cmd_thread = CustomCommandThread(self.git, *args, custom_environ=custom_environ)
             cmd_thread.start()
         else:
-            stdout = self.git(*args)
+            stdout = self.git(*args, custom_environ=custom_environ)
         sublime.status_message(complete_msg)
 
         if output_to_panel:
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -70,7 +70,13 @@ class GitCommand(StatusMixin,
     _quick_panel_log_idx = 1
     _quick_panel_branch_diff_history_idx = 1
 
-    def git(self, *args, stdin=None, working_dir=None, show_panel=False, throw_on_stderr=True, decode=True):
+    def git(self, *args,
+            stdin=None,
+            working_dir=None,
+            show_panel=False,
+            throw_on_stderr=True,
+            decode=True,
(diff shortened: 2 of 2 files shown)


core/commands/custom.py
```diff
@@ -44,6 +42,7 @@ def run(self, **kwargs):
                 None
             )
 
+        print("sending run_async kwargs: %s" % kwargs)
```
stoivo: Do you think we should have the print here or did you leave it my mistake?

stoivo:
Other than the small comment it looks good. 

Could you update the docs too? [docs/custom.md](https://github.com/divmain/GitSavvy/blob/675ad2ec3f77ace31c3a3a30b8dd027c61e59e86/docs/custom.md)

asfaltboy:
@stoivo thanks for the review, the print was indeed just forgotten to be removed. Amended the last commit.

stoivo:
I did some testing today.
I copied [this](https://github.com/divmain/GitSavvy/blob/675ad2ec3f77ace31c3a3a30b8dd027c61e59e86/docs/custom.md#examples) form the docs
```json
    {
        "caption": "git: gui blame",
        "command": "gs_custom",
        "args": {
            "output_to_panel": false,
            "args": ["gui", "blame", "{FILE_PATH}"],
            "start_msg": "Starting gui blame...",
            "complete_msg": "Gui blame started",
            "run_in_thread": true  // SEE WARNING ABOVE !
        }
    }
```
and it did not work. I got this output in console.
```python
  File "/Users/simon/Library/Application Support/Sublime Text 3/Packages/GitSavvy/core/commands/custom.py", line 45, in <lambda>
    sublime.set_timeout_async(lambda: self.run_async(**kwargs), 0)
  File "/Users/simon/Library/Application Support/Sublime Text 3/Packages/GitSavvy/core/commands/custom.py", line 67, in run_async
    cmd_thread = CustomCommandThread(self.git, *args, custom_environ=custom_environ)
  File "/Users/simon/Library/Application Support/Sublime Text 3/Packages/GitSavvy/core/commands/custom.py", line 12, in __init__
    super(CustomCommandThread, self).__init__(**kwargs)
TypeError: __init__() got an unexpected keyword argument 'custom_environ'
```

If I set `run_in_thread` to `false` is starts working.

Can you check if this is working for you or not? 
Could you update the docs, just add your command as one example that uses custom_environ and add custom_environ to the argument list with a description?

asfaltboy:
I could  not figure out why I can't just pass around kwargs from the `run` until `git` methods, which is why I had to explicitly use `custom_environ` argument in the below calls. I think I simply forgot to test `CustomCommandThread`, but it looks like exactly the same issue as I had before. Why kwargs cannot be passed is beyond me, so lets just use explicit `custom_environ` argument here too.

stoivo:
It look like you have to do that, but I no professional python developer...
Maybe @divmain knows how to do it?

asfaltboy:
Well in  that case, the threading. Thread parent class doesn't accept custom_environ as a keyword argument, so removing it from kwargs should do it. I guess this also has something to do with passing kwargs to a `sublime_plugin.WindowCommand` subclass (even though we don't explicitly call super).

stoivo:
Now I am having this error
```python
Exception in thread Thread-209:
  File "./python3.3/threading.py", line 901, in _bootstrap_inner
  File "/Users/simon/Library/Application Support/Sublime Text 3/Packages/GitSavvy/core/commands/custom.py", line 19, in run
    custom_environ=custom_environ)
NameError: global name 'custom_environ' is not defined
```

This is the Custom command
```json
    {
        "caption": "git: gui blame",
        "command": "gs_custom",
        "args": {
            "output_to_panel": false,
            "args": ["gui", "blame", "{FILE_PATH}"],
            "start_msg": "Starting gui blame...",
            "complete_msg": "Gui blame started",
            "run_in_thread": true  // SEE WARNING ABOVE !
        }
    }
```

asfaltboy:
Sorry I'm not paying attention and unable to test my commits. We should really work adding tests

stoivo:
You should be able to test your commit.  If you save the file and run `GitSavvy: reload modules (debug)` you can test it without restarting sublime.

asfaltboy:
Yes I am aware, I meant the automated kind of testing, as in unit-testing. Out of time for this for now, will continue next time.

stoivo:
I manually merged your PR and added some docs. https://github.com/divmain/GitSavvy/commit/42f6e412bad42ab1f7d2dee2c9cfb2ba70efa5fc

stoivo:
I forgot to say, thanks for submitting pull request.

asfaltboy:
Cool stuff thanks!

PR #431, opened 2016-06-20
Syntax improvements for custom color schemes

aziz:
These are subtle syntax changes to be able to write a better color scheme and create a better UI

-  Adding a spacial scope to the last `-` of the key-binding-menu
   -  The extra \n helps the background color set by color scheme to expand to the
    end of viewport.
-    better commit help section
    - The extra \n helps the background color set by color scheme to expand to
    the end of viewport.
    - the spacing between help section is replace with empty ## to be captured
    by syntax and colored as a single section by color scheme

screen shot 2016-06-21 at 1 19 44 am
screen shot 2016-06-21 at 1 21 15 am

### DIFF
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -7,7 +7,7 @@ from ..git_command import GitCommand
 from ...common import util
 
 
-COMMIT_HELP_TEXT_EXTRA = """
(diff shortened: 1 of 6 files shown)


divmain:
This looks good to me.  Thanks!

PR #437, opened 2016-06-25
Syntax improvements for various syntaxes

aziz:
## Rebase
- Improved rebase dashboard syntax + test
- Changed rebase graph unicode characters to make it look better
- The syntax still support old unicode characters 

## Graph syntax
- more detailed scopes 
- added test
- replaced `*` in graph with `●`

## Diff syntax
- improved diff syntax and added test 
- Fixed a bug where diff can not detect new file where the header is like `--- /dev/null`

## Show commit
- improved show_commit syntax and added test
- show commit buffer should be read-only

## Make commit
- make commit should use the GitSavvy diff syntax not the built-in diff syntax 
- improved make commit syntax and added test

## Blame 
- improved blamed syntax and test

## Interactive rebase
- Improved interactive rebase syntax and added test

Here is how it looks. you can try it by using [my custom color scheme](https://github.com/aziz/sublimeText3-Userfiles/blob/master/Color%20Schemes/GitSavvy.tmTheme)

screen shot 2016-06-25 at 12 13 47 pm

### DIFF
--- a/core/commands/log_graph.py
+++ b/core/commands/log_graph.py
@@ -5,6 +5,7 @@ from ..git_command import GitCommand
 from ...common import util
 
 LOG_GRAPH_TITLE = "GRAPH"
+COMMIT_NODE_CHAR = "●"
 
 
 class GsLogGraphCommand(WindowCommand, GitCommand):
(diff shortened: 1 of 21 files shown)


divmain:
This is really great.  Would you mind rebasing and cleaning up the Git history somewhat?  I use `git: generate changelog` when announcing new versions, and this would add a bit of noise for our next release.  [This](https://github.com/divmain/GitSavvy/blob/master/CONTRIBUTE.md#commit-message-structure) is the preferred format, but just squashing related commits together would be more than enough.

divmain:
As a separate thought, perhaps we should provide official GitSavvy color schemes and provide a mechanism to switch between them.  I like the colors you're using, and if we added a dark theme too, that might be received well.

aziz:
Rebase is done. 
Regarding default color scheme, I agree. my custom color scheme needs a lot of cleaning. The colors are borrowed from the offical git website. I'll send another pull request for the color scheme when I cleaned it up.

divmain:
Sounds great - thanks again.

PR #439, opened 2016-06-26
Feature: Toggle help menu, on demand or through settings

aziz:
- toggle help menu using `?` shortcut
- added `hide_help_menu` settings to diable help
- help menu can always be shown again using the shortcut

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -320,6 +320,14 @@
     //////////////////////////////
 
     {
+        "keys": ["?"],
+        "command": "gs_interface_toggle_help",
+        "context": [
+            { "key": "setting.command_mode", "operator": "equal", "operand": false },
+            { "key": "setting.git_savvy.status_view", "operator": "equal", "operand": true }
+        ]
+    },
+    {
         "keys": ["r"],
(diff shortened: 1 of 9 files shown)


stoivo:
I think this is helpful for those you use GitSavvy a lot and start to remember most of the keys  you use. Also This will not break the UX in any whey, since the first visit is the same as before.

stoivo:
I would like to hear @divmain s opinion.

We had one issue on this earlier #184.

aziz:
Having the `?` as the help shortcut, we can later on add help sections to those pages which do not have help using the new `show_popup` API. I'm working on a proof of concept and it looks good. 

screen shot 2016-06-28 at 12 05 59 am

divmain:
Heh.  This has come up again!  Since it seems to be a popular request and since the default behavior is the same, I suppose I'm okay with it.  I do think the `show_popup` API provides potential for UX affordances in non-dashboard views down the line.  I'm okay moving forward with this, provided that it is clear that there should always be a single source of truth for what help information is displayed to the user.

stoivo:
Great

PR #445, opened 2016-06-28
Improvement: Using rulers to fake a solid line for graph view

aziz:
screen shot 2016-06-28 at 2 30 34 am


divmain:
Hmm, I'm not excited about this one.  Improving fidelity for one piece of the graph and not others (which we can't mimic with rulers) seems mismatched.  I wonder if there are unicode characters available that could get us closer to what we want - fewer gaps and better alignment between graph characters, and a cleaner looking graph in general.

aziz:
I understand. It was just a cool idea 😄 
Anyone could do this on their own without touching the code by adding the rulers config to their syntax specific settings for graph view.

PR #447, opened 2016-06-28
Popup shortcuts for graph view

### LINKED ISSUE #316: Help at top of Graph page
Are there operations you can perform on the Graph view? How about a short summary of the keyboard shortcuts available at the top? I see that <kbd>,</kbd> and <kbd>.</kbd> navigate commits, but guessing there are other things you can do here - and now way to discover them. (Obvious things to want to do are diff, checkout, reset...)

aziz:
fixed #316 

screen shot 2016-06-28 at 4 41 20 pm

The shortcut key to view the help popup is `?`
The look and feel could be customized using the color scheme's  `popupCSS` setting

### DIFF
--- a/popups/log_graph_view.html
+++ b/popups/log_graph_view.html
@@ -1,0 +1,31 @@
+<html>
+<style>
+{css}
+</style>
+<body>
+<div class="header-wrapper"><h2>Keyboard Shortcuts</h2></div>
+
+<h3>Actions</h3>
+<ul>
+  <li><code><span class="shortcut-key">enter&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</span>view commit</code></li>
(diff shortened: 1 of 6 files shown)


popups/log_graph_view.html
```diff
@@ -0,0 +1,31 @@
+<html>
+<style>
+{css}
+</style>
+<body>
+<div class="header-wrapper"><h2>Keyboard Shortcuts</h2></div>
+
+<h3>Actions</h3>
+<ul>
+  <li><code><span class="shortcut-key">enter&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</span>view commit</code></li>
```
divmain: Does `minihtml` support any sort of layout through CSS?  If there isn't any way to align columns without including spaces, this is fine.  But if it can be done with CSS, all the better.

divmain:
Just the one question.  My guess is that `minihtml`'s CSS doesn't have anything like layout.  If it does, you or I can take another look at it later.  Merging this in, since it includes a fix for the log view commands.

PR #448, opened 2016-06-30
Handle Vintageous in Inline Diff

asfaltboy:
* this view relies heavily on keyboard interaction, it only makes sense
  to enable insert mode if we're vintageous friendly


stoivo:
Thanks for contribution.

PR #450, opened 2016-06-30
Preventing generated color schemes to appear

aziz:
Preventing generated color schemes to appear in Sublime's color scheme menu.

This is a trick that sublime itself uses to hide some color schemes and syntaxes


divmain:
Neat trick!

PR #408, opened 2016-05-13
dev_mode: Reliable plugin reloads

abusalimov:
This PR implements reloading modules in a more robust and reliable way using import hooks. The key feature is that the modules get reloaded in the very same order they would be loaded during fresh startup.

Here's a part of debug output showing the proper reload sequence (I don't include debug trace facilities for now though, that is a subject for a separate PR):
```
GS [reload] begin ══════════════════════════════════════════════════════
reloading plugin GitSavvy.git_savvy
GS [reload]  ┡━─ GitSavvy.git_savvy
GS [reload]  ╿ ┡━─ GitSavvy.common
GS [reload]  ╿ ┡━─ GitSavvy.common.commands
GS [reload]  ╿ ╿ ┡━─ GitSavvy.common.commands.debug
GS [reload]  ╿ ╿ ╿ ┡━─ GitSavvy.common.util
GS [reload]  ╿ ╿ ╿ ╿ ┡━─ GitSavvy.common.util.parse_diff
GS [reload]  ╿ ╿ ╿ ╿ ┡━─ GitSavvy.common.util.dates
GS [reload]  ╿ ╿ ╿ ╿ ┡━─ GitSavvy.common.util.view
GS [reload]  ╿ ╿ ╿ ╿ ┡━─ GitSavvy.common.util.file
GS [reload]  ╿ ╿ ╿ ╿ ╿ ┡━─ GitSavvy.common.vendor
GS [reload]  ╿ ╿ ╿ ╿ ╿ ┡━─ GitSavvy.common.vendor.yaml
GS [reload]  ╿ ╿ ╿ ╿ ╿ ╿ ┡━─ GitSavvy.common.vendor.yaml.error
GS [reload]  ╿ ╿ ╿ ╿ ╿ ╿ ┡━─ GitSavvy.common.vendor.yaml.tokens
GS [reload]  ╿ ╿ ╿ ╿ ╿ ╿ ┡━─ GitSavvy.common.vendor.yaml.events
GS [reload]  ╿ ╿ ╿ ╿ ╿ ╿ ┡━─ GitSavvy.common.vendor.yaml.nodes
GS [reload]  ╿ ╿ ╿ ╿ ╿ ╿ ┡━─ GitSavvy.common.vendor.yaml.loader
GS [reload]  ╿ ╿ ╿ ╿ ╿ ╿ ╿ ┡━─ GitSavvy.common.vendor.yaml.reader
[...]

PR #451, opened 2016-07-08
dev_mode: Add debug trace facility controlled through "dev_trace" setting

abusalimov:
As a use case, add tracing to the `reload` module to get the output advertised in #408.

### DIFF
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -32,7 +32,8 @@
         Only takes its effect when "dev_mode" is set to `true`.
      */
     "dev_trace": [
-        "debug"
+        "debug",
+        "reload"
     ],
 
     /*
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -28,6 +28,14 @@
     "dev_mode": false,
 
     /*
+        A list of tags (subsystems) to filter the output of debug.trace().
+        Only takes its effect when "dev_mode" is set to `true`.
+     */
+    "dev_trace": [
+        "debug"
+    ],
+
+    /*
         Enter your github API key in the field below.  To interact with
         a GitHub Enterprise instance, add a field with the FQDN as the key
         and the API key as the value.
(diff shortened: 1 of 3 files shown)


divmain:
This is really interesting.  Thanks for your work on the debug mode stuff - it's not typically where we get contributions!

stoivo:
You are so write. I started looking at this but I don't understand what is happening in 3bb59cb33da4f6e10fd4b0eb3df4c6361e936ff3. Can one of you explain it a bit and how to use it?

divmain:
@stoivo, `trace` is essentially a wrapper around `print`, mostly adding formatted meta-data and consistent output.  In the context of this larger, it allows for pretty debug output when reloading like you see in https://github.com/divmain/GitSavvy/pull/408.

stoivo:
How would you use it? 

When to use `dump_var` vs `trace`? That is the difference I don't see.

abusalimov:
@divmain you're welcome! I was experimenting on inline diff view -> switched to refactoring `ui.py` -> encountered reloading issues -> switched to reload -> and here we are, debugging it :wink: 

@stoivo `trace` is essentially an ad-hoc replacement for the standard `logging`:
```py
log = trace.for_tag("reload")  # read as: logger = logging.getLogger(__name__)
```
And then scatter around `log()` calls. As like using the standard logging, one can filter the output using trace tags.

---

On the other hand, `dump` and `dump_var` are more like handy helpers during hacking, useful for printf-style debugging. Usage of these is not supposed to be checked into the repository.
```py
def some_nasty_stuff(foo):
    ... # the real logic

    # XXX remove!
    # and if something performs wrong:
    dump(foo)         # arg0 = ...
    dump(foo=foo)     # foo = ...

    # or, to show all arguments and local variables:
    dump(**locals())  # foo = ...
                      # bar = ...
```

`dump_var` is essentially an internal function outputting a single value, a call to `dump(foo=foo)` is then handled as `dump_var("foo", foo)`. The former is just a bit less to type. :)

stoivo:
Thanks @abusalimov

PR #459, opened 2016-07-19
Enhancement: return the only remote when there is only one

randy3k:
as title and added various fall backs when the integrated remote cannot be determined


divmain:
Looks good to me - thanks again!

PR #461, opened 2016-07-19
Enhancement: github: open file and open repo

### LINKED ISSUE #335: github: open repo and open file should open in the tracked remote
Right now, `open repo` and `open file` commands open browser at the `get_integrated_remote_url()`. However, in some cases, the active branch is actually tracking a forked remote and I actually wish to open the repo/file at the tracking remote. 

Here is my suggestion, if upstream is not set or the same as the `get_integrated_remote_url`, we should directly open the browser. However,  if the active upstream is not `get_integrated_remote_url()`, we could display a panel to allow user to open the file at either remotes.

randy3k:
`github: open file` and `open repo` are now remote sensitive. They will check if the current tracking remote is the same as the integrated remote. If they are the same (or there is only one remote), the browser will be opened directly. Otherwise, the list of remotes will be displayed to choose from.
untitled

depends on #459 
close #335

### DIFF
--- a/github/commands/open_on_remote.py
+++ b/github/commands/open_on_remote.py
@@ -1,3 +1,4 @@
+import sublime
 from sublime_plugin import TextCommand
 
 from ...core.git_command import GitCommand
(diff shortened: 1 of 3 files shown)


randy3k:
I have also update the code such that it handles opening multiple files but only asks for remote once.

github/commands/open_on_remote.py
```diff
@@ -13,17 +14,67 @@ class GsOpenFileOnRemoteCommand(TextCommand, GitCommand, git_mixins.GithubRemote
...
     At present, this only supports github.com and GitHub enterprise.
     """
 
-    def run(self, edit, preselect=False, fpath=None):
-        fpath = fpath or self.get_rel_path()
+    def run(self, edit, remote=None, preselect=False, fpath=None):
+        self.fpath = fpath or self.get_rel_path()
+        self.preselect = preselect
+
+        self.remotes = self.get_remotes()
```
divmain: This (and later commands) should run on the non-UI thread, so as to avoid blocking user interaction.  Look for anywhere else we're using `set_timeout_async` and `run_async` pattern.

randy3k:
@divmain. Done, using `sublime.set_timeout_async`.

github/commands/open_on_remote.py
```diff
@@ -13,17 +14,60 @@ class GsOpenFileOnRemoteCommand(TextCommand, GitCommand, git_mixins.GithubRemote
...
+    to choose from.
 
     At present, this only supports github.com and GitHub enterprise.
     """
 
-    def run(self, edit, preselect=False, fpath=None):
-        fpath = fpath or self.get_rel_path()
+    def run(self, edit, remote=None, preselect=False, fpath=None):
+        sublime.set_timeout_async(
+            lambda: self.run_async(edit, remote, preselect, fpath))
```
divmain: No need to pass on `edit` here.  In fact, for future reference, that `edit` object is unusable once `run` has completed execution.  Any edits need to happen synchronously, which is why we trigger an event whenever we want a content change to occur.

github/commands/open_on_remote.py
```diff
@@ -47,8 +92,37 @@ class GsOpenGithubRepoCommand(TextCommand, GitCommand, git_mixins.GithubRemotesM
     Open a new browser window to the GitHub remote repository.
     """
 
-    def run(self, edit):
-        open_repo(self.get_integrated_remote_url())
+    def run(self, edit, remote=None):
+        sublime.set_timeout_async(lambda: self.run_async(edit, remote))
```
divmain: Same thing here

divmain:
Just a minor change, and the rest looks good.  It seems like maybe this could be DRY'd up a little, but I don't mind merging it in once `edit` is removed.  Thanks again for all the PRs this evening.  I pushed out the already-merged ones as a patch-release since they addressed regressions.  This will go out in the next feature release.

github/commands/open_on_remote.py
randy3k: Thanks for the information. Fixed and rebased.

randy3k:
The `edit`s are removed. Thanks for considering the PRs. 

Thanks again for this amazing Package, it is the reason I stay using Sublime Text but not Atom or other editors.

divmain:
Same here!  I've considered moving over to Visual Studio Code (I still have trouble with performance in Atom), but GitSavvy is a pretty integral part of my workflow these days.

Anyway, this PR looks good.  Merging.

PR #464, opened 2016-07-19
Enhancement: add github integration config GitSavvy.ghBranch

randy3k:
This is needed to produce a correct comparision when the integered
remote is a forked remote.

For example, if the integrated remote is configed as randy3k,
the original code opens
https://github.com/randy3k/GitSavvy/compare/randy3k:comparision?expand=1

However, github redirects it to

https://github.com/divmain/GitSavvy/compare/master...randy3k:comparision?expand=1

To adderss the issue, the new link used is

https://github.com/randy3k/GitSavvy/compare/randy3k:master...randy3k:comparision?expand=1

### DIFF
--- a/github/commands/configure.py
+++ b/github/commands/configure.py
@@ -5,6 +5,7 @@ from ...core.git_command import GitCommand
 
 
 NO_REMOTES_MESSAGE = "You have not configured any remotes."
+NO_REMOTE_BRANCHES_MESSAGE = "Remote does not have any branches."
 
 
 class GsConfigureGithubRemoteCommand(WindowCommand, GitCommand):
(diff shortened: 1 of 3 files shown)


randy3k:
Actually, I was thinking whether I should use `GitSavvy.ghBranch` or a project specific settings
 as in [rebase](https://github.com/divmain/GitSavvy/blob/6466acad3925b4e7a541259b55b52218d2be1293/core/interfaces/rebase.py#L273).

divmain:
Yeah, I think this would be an appropriate place for the project config.  If we really want to get fancy, we could use the GitHub API to see if the specified remote is a fork, and act accordingly.  Should be able to borrow a lot from [here](https://github.com/divmain/GitSavvy/blob/master/github/commands/add_fork_as_remote.py#L20-L45).

randy3k:
I think there is no need for now to check if the remote is a fork as we switched to a more general url format: `https://github.com/{}/{}/compare/{}:{}...{}:{}?expand=1` to replace the old format
`https://github.com/{}/{}/compare/{}:{}?expand=1`

The current implementation should be good enough.

github/commands/configure.py
```diff
@@ -5,6 +5,7 @@
 
 
 NO_REMOTES_MESSAGE = "You have not configured any remotes."
+NO_REMOTE_BRANCHES_MESSAGE = "Remote does not has any branches."
```
divmain: `Remote does not have any branches.`

divmain:
Cool, that sounds reasonable.  One minor change, and then I'll merge this in.

github/commands/configure.py
randy3k: Oh.....shame on my poor grammar...

randy3k:
Fixed the typo.

PR #467, opened 2016-07-20
Feature: Copy the full SHA and show commit in log quick panel

randy3k:
- Copy the full SHA in log quick panel
- Show commit in output panel while scrolling the log quick panel
- fix default indexes

untitled 2

### DIFF
--- a/core/commands/log.py
+++ b/core/commands/log.py
@@ -65,6 +65,7 @@ class GsLogCommand(WindowCommand, GitCommand):
     def on_hash_selection(self, index):
         options_array = [
                 "Show commit",
+                "Copy the full SHA",
                 "Compare commit against working directory",
                 "Compare commit against index"
         ]
@@ -76,7 +77,7 @@ class GsLogCommand(WindowCommand, GitCommand):
             return
         if index == self._limit:
             self._pagination += self._limit
-            sublime.set_timeout_async(lambda: self.run_async(), 1)
+            sublime.set_timeout_async(self.run_async, 1)
             return
 
         self._selected_hash = self._hashes[index]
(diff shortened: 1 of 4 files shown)


divmain:
I actually really like this, but am concerned that it might add some unfortunate overhead to interaction on slower machines.  I can't tell if the `on_highlight` method will run in the primary thread but - if so - this would seriously impact responsiveness on older machines, or just machines without SSD.  Can you comment?

randy3k:
You are right, `on_highlight` runs on the primary thread. I have made it as an async call now.
Anyway, there is an option `log_show_more_commit_info` to disable the output panel.

core/commands/log.py
```diff
@@ -56,12 +56,42 @@ def run_async(self):
...
+            pre_selected_index = 0
+
+        # _on_hash_selection has to be called by set_timeout_async
+        # otherwise, on_highlight_commit_async will be called after
+        # _on_hash_selection is executed.
         self.window.show_quick_panel(
             self._entries,
-            self.on_hash_selection,
-            flags=sublime.MONOSPACE_FONT
+            lambda index: sublime.set_timeout_async(lambda: self._on_hash_selection(index), 10),
```
stoivo: Is there a reason why we have a 10 milliseconds delay?

stoivo:
I have one small comment, other that that this looks great.

core/commands/log.py
randy3k: To make sure it runs after `on_highlight_commit`. It seems that both `on_highlight_commit` and `_on_hash_selection` are executed when enter is hit

divmain:
Looks good to me!

Issue #346, opened 2016-01-20
git push resulting in "fatal: could not read Username"...


cvax:
When I try to do a git push I get this error
```fatal: could not read Username for 'https://github.com': Invalid argument```

Here is the full message:
```
`c:\program files\git\cmd\git.EXE push` failed with following output:

bash: /dev/tty: No such device or address
error: failed to execute prompt script (exit code 1)
fatal: could not read Username for 'https://github.com': Invalid argument
```

cvax:
Figured it out. Had to save a credential to git for gitsavvy to work.

stoivo:
Glad you made it and come back if there are anything we can help with.

hnbeck:
Hi cvax
what exactly you have done by "saving git credentials" ? I've the same problem although I have an valid git API token configured in user settings

Cheers

Hans

stoivo:
Hi, @hnbeck please open an issue with your problem.

And add the details for os and what you want to accomplish. 

git API token? Github personal token?

PR #468, opened 2016-07-20
Feature: Prompt for credentials

gwenzek:
Instead of failing when no credentials are found, ask them to the user.
This feature is opt-in through the "prompt_credentials" setting, and is only used on 'push', 'pull' and 'fetch'.
Solves #346.

The main drawback is that it add a dependency to "pexpect" for passing the username and password to Git from Sublime.
I used an [old version of pexpect](https://github.com/pexpect/pexpect/blob/3.x/pexpect/__init__.py) where the code was regrouped in only one file.

### DIFF
--- a/common/vendor/pexpect.py
+++ b/common/vendor/pexpect.py
@@ -1,0 +1,2134 @@
+# Pexpect from https://github.com/pexpect/pexpect/blob/aac5897aa12daf056b8fe08f1b6512d9f60c2d27/pexpect/__init__.py
+
+
+'''Pexpect is a Python module for spawning child applications and controlling
+them automatically. Pexpect can be used for automating interactive applications
+such as ssh, ftp, passwd, telnet, etc. It can be used to a automate setup
+scripts for duplicating software package installations on different servers. It
+can be used for automated software testing. Pexpect is in the spirit of Don
+Libes' Expect, but Pexpect is pure Python. Other Expect-like modules for Python
+require TCL and Expect or require C extensions to be compiled. Pexpect does not
+use C, Expect, or TCL extensions. It should work on any platform that supports
+the standard Python pty module. The Pexpect interface focuses on ease of use so
+that simple tasks are easy.
+
+There are two main interfaces to the Pexpect system; these are the function,
+run() and the class, spawn. The spawn class is more powerful. The run()
+function is simpler than spawn, and is good for quickly calling program. When
+you call the run() function it executes a given program and then returns the
+output. This is a handy replacement for os.system().
+
+For example::
+
+    pexpect.run('ls -la')
+
+The spawn class is the more powerful interface to the Pexpect system. You can
+use this to spawn a child program then interact with it by sending input and
+expecting responses (waiting for patterns in the child's output).
+
+For example::
+
+    child = pexpect.spawn('scp foo user@example.com:.')
+    child.expect('Password:')
+    child.sendline(mypassword)
+
+This works even for commands that ask for passwords or other input outside of
+the normal stdio streams. For example, ssh reads input directly from the TTY
+device which bypasses stdin.
+
+Credits: Noah Spurrier, Richard Holden, Marco Molteni, Kimberley Burchett,
+Robert Stone, Hartmut Goebel, Chad Schroeder, Erick Tryzelaar, Dave Kirby, Ids
+vander Molen, George Todd, Noel Taylor, Nicolas D. Cesar, Alexander Gattin,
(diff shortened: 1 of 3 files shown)


core/git_command.py
```diff
@@ -150,19 +156,66 @@ def raise_error(msg):
...
+
+                self._stdout = None
+                is_window_command = hasattr(self, "window")
+                window = self.window if is_window_command else self.view.window()
+                window.show_input_panel('username:', '', lambda username: self.on_username(window, command_str, username, on_result), None, self.on_cancel)
+
+                # wait that the Git command has been executed in the show_input_panel thread.
+                if is_window_command:
+                    while self._stdout is None:
+                        time.sleep(1)
```
gwenzek: Here I'm waiting the user to input his login and password. This is only possible with a `WindowCommand`. In case of a `TextCommand` I return immediately, meaning that the caller won't have access to the stdout.
This is generally not a big deal since nobody uses the output of "push", "pull" or "fetch".

core/git_command.py
```diff
@@ -150,19 +156,66 @@ def raise_error(msg):
...
 
+    def on_username(self, window, command_str, username, on_result):
+        """ Asks for the password, before calling the given command. """
+        window.show_input_panel(username + "'s password:", '', lambda password: self.on_credentials(command_str, username, password, on_result), None, self.on_cancel)
+
+    def on_credentials(self, command_str, username, password, on_result):
+        """ Launch the given command, stores the ouput and return code in self. """
+        # pexpect runs the given command in a subprocess, parses the output and feeds the username and password when asked by Git.
+        # [pexpect](https://github.com/pexpect/pexpect)
+        stdout, return_code = pexpect.run(command_str, events={'Username for': username + '\n', 'Password for': password + '\n'}, withexitstatus=True)
```
gwenzek: This is the only part of the code where I use "pexpect", if someone manage to use Popen for doing the same thing, we could drop the dependency.

common/vendor/pexpect.py
```diff
@@ -0,0 +1,2134 @@
...
+                    if err.args[0] not in (errno.EINVAL, errno.ENOTTY):
+                        raise
+
+            # gwenzek, HACK: I removed this line because I wanted to get rid of 'rresource' dependency,
+            # but I have no idea what it was supposed to do.
+            #
+            # Do not allow child to inherit open file descriptors from parent.
+            # max_fd = resource.getrlimit(resource.RLIMIT_NOFILE)[0]
+            # os.closerange(3, max_fd)
+            os.closerange(3, 1024)
```
gwenzek: This part is a bit hacky, I don't have any knowledge of how "os.closerange" works. On my computer `resource.getrlimit(resource.RLIMIT_NOFILE)` returns `(1024, 65536)` that's why I hardcoded 1024 here

randy3k:
AFAIK, old versions of Pexpect doesn't support window. Shall we move to newer version instead?

gwenzek:
Good point, I only tested on Ubuntu 16.04. But as I said, I'm not sure this dependency is mandatory, Popen should allow to pass Username and Password through stdin by I didn't manage to make it work.

randy3k:
I would not use the handcrafted code with Popen when pexpect, which is more reliable, exists

stoivo:
Have you tested this on Windows?

divmain:
Yeah, current docs state:
> The main features of Pexpect require the pty module in the Python standard library

I really _would_ like to merge this, but I think this is 1) very heavy, and 2) not portable.  As mentioned in another ticket, when I have time I'll be revisiting how GitSavvy interacts with `git` in the same way that SublimeRepl does it.  That would be cross-platform and hopefully will provide a foundation for interactivity with `git` when it prompts for input.

Also, this seems to mostly work for `pull`, `push`, and `fetch`, but I'd rather see a solution that works for all commands, since there are other cases when interactivity is desired.  The return-early return-nothing when `fatal: could not read Username` is detected seems brittle.

For now, I'm going to have to close this without merging.  If you can think of another way to make this happen, I'm definitely open to ideas!

gwenzek:
@divmain I agree it's a bit hacky.
I put the limitation on "push", "pull" and "fetch" to avoid catching errors not related to authentication.

I haven't try it on Windows mostly because Git for Windows seems to permanently store credential by default.

Maybe I could take a look at SublimeRepl code to see how they provide this kind of interaction.

divmain:
If you'd like to take a look (and/or create an issue to track), please feel free.  It'll be a significant change, but definitely worth investigating!

PR #469, opened 2016-07-20
Fix: allow remote tags using gs_tag_delete and gs_tag_view_log

randy3k:
df95ea3f7276bb5b50bbdf79f3064ff5b8054a55 did not allow remote tags.


stoivo:
Thanks.

PR #471, opened 2016-07-21
Fix: quick stage handles merge conflicts

randy3k:
move to `sort_status_entries` to mixins and use `sort_status_entries` to fix `quick stage` issues of handling merge conflicts.


divmain:
Can you explain a bit more about what this solves?  The code seems fine, just want to make sure I understand what _didn't_ work before.

randy3k:
When a file has a merge conflict, the file status is "UU". The previous code doesn't handle it and shows the file as staged.

divmain:
Ah, thanks!  LGTM.

PR #472, opened 2016-07-23
Enhancedment: use thread.Thread to execute util.file.determine_syntax_files

randy3k:
It seems that sublime.set_timeout_async still slow down the start up of
Sublime Text. For instance, the quick panel (e.g., `git: log`) is not ready until a few second
after ST opens. I suspect that they share the same second thread.

Changing `sublime.set_timeout_async` to `threading.Thread` so that the quick panel
(and other UI) does not have to wait for util.file.determine_syntax_files

### DIFF
--- a/git_savvy.py
+++ b/git_savvy.py
@@ -1,13 +1,14 @@
 import sys
-
+import threading
 import sublime
+
 
 if sys.version_info[0] == 2:
     raise ImportWarning("GitSavvy does not support Sublime Text 2.")
 else:
     def plugin_loaded():
         from .common import util
-        sublime.set_timeout_async(util.file.determine_syntax_files)
+        threading.Thread(target=util.file.determine_syntax_files).start()
 
         # Ensure all interfaces are ready.
         sublime.set_timeout_async(


stoivo:
This is good stuff.

divmain:
Hmm... this one makes me nervous, since it introduces a race condition in accessing `syntax_file_map`.  I'm going to revert this merge for now, but feel free to re-submit with proper locking mechanisms in place.

stoivo:
@divmain, I don't understand. Can you explain a bit more in detail?

divmain:
When talking about threading, we have to consider thread safety.  [This block](https://github.com/divmain/GitSavvy/blob/master/common/util/file.py#L8-L20) of code would run in one thread, while [this one](https://github.com/divmain/GitSavvy/blob/master/common/util/file.py#L23-L26) would run in a separate thread.  Python does have a GIL, which means that no two threads will be executing simultaneously - it's essentially a thread-level lock that must be acquired by a thread before it can access the shared memory and proceed with execution.

However, this only protects _atomic_ operations.  By my understanding, the GIL will transition from one thread to the next in between bytecode instructions.  In Python, this roughly translates to: single statements on built-ins will operate atomically, without introducing race conditions (where one thread is reading/writing an object while another thread is simultaneously reading/writing).  However, [these lines](https://github.com/divmain/GitSavvy/blob/master/common/util/file.py#L15-L18) are not atomic.  And even if they were, we don't have any protections in our codebase to ensure that they remain that way.

What I'd prefer here is to `Lock` the `syntax_file_map` until writes are complete, and then acquire that lock whenever `get_syntax_for_file` is invoked.  

Here are the [docs](https://docs.python.org/3.3/library/threading.html#lock-objects).

randy3k:
I may be wrong.
when `get_syntax_for_file` is executed in the main thread, and we acquire a lock for `get_syntax_for_file`, it may block the main thread. So I use a different approach in https://github.com/divmain/GitSavvy/pull/479 to check the thread of running `determine_syntax_files` is still alive.

PR #470, opened 2016-07-21
Feature: Compare commit against a different commit

randy3k:
This depends on #466 and #467. Close #373 

untitled
(the screen shot may be reflect the latest code)

PR #473, opened 2016-07-29
Fix: resolve possible symlink in "Show File at Commit"

randy3k:
`repo_path` always returns a "realpath", so we also need to resolve `filepath`.


randy3k:
The fix depends on #470, so the fix is included in #470 instead.

PR #474, opened 2016-07-29
Enhancement: show the list of commits again when reset mode selection is canceled.

randy3k:
Enhancement: show the list of commits again when reset mode selection is canceled.


stoivo:
Thanks

PR #477, opened 2016-07-29
Enhancement: use quick commit instead of commit

randy3k:
I don't know if it makes sense or not. But I think it is more intuitive to call `git: quick commit` other than `git: commit` in `quick stage`.


stoivo:
I don't use this at all so I would like to have @divmain  or @yyyc514 option since they made this.

divmain:
Seems like a reasonable change.  If folks complain, we can always switch it back or make it configurable.

stoivo:
Would you make a configuration for what a command should be called? :P

PR #479, opened 2016-07-30
Enhancement: use thread.Thread to execute determine_syntax_files

randy3k:
Another attempt of #472 

- If the threading is still running, return "Plain text". (Locking mechanism is hard to implement- when get_syntax_for_file is executed in the main thread, and we acquire a lock for get_syntax_for_file, it may block the main thread?)

- Do not run `determine_syntax_files` again when reloading the GitSavvy.

### DIFF
--- a/common/util/file.py
+++ b/common/util/file.py
@@ -1,11 +1,24 @@
 import sublime
+import threading
 from ..vendor import yaml
 
 
-syntax_file_map = {}
+if 'syntax_file_map' not in globals():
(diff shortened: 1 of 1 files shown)


stoivo:
I don't have the courage to merge this one now... Better check with @divmain

divmain:
This'll do.  Thanks @randy3k, and sorry for the reverse course :)

randy3k:
No problem. It was because I didn't think carefully.

PR #481, opened 2016-07-30
Feature: Compare commit against a different commit

### LINKED ISSUE #373: Git Diff against Specific Commit
Maybe I am just missing something obvious, but is there a way to view the differences against a specific commit, rather than just the most recent one?

randy3k:
The main feature of this PR is the comparison of commits. It refactors `gs_branches_diff_commit_history` to `gs_compare_commit` to allow comparing any commits.

It also adds various keybinds and help pages.

Close #373

untitled mov


core/commands/commit_compare.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+
+    def run(self, edit, base_commit, target_commit=None, file_path=None, title=None):
+        self.base_commit = base_commit
+        self.target_commit = target_commit if target_commit else "HEAD"
+        self._file_path = file_path
+        if not title:
+            title = "COMMIT COMPARISON"
+        self.title = title
+        sublime.set_timeout_async(lambda: self.run_async())
```
divmain: `sublime.set_timeout_async(self.run_async)`

core/commands/commit_compare.py
```diff
@@ -0,0 +1,245 @@
...
+
+class GsCompareCommitCommand(TextCommand, GitCommand):
+
+    """
+    Show a view of all commits diff between branches.
+    """
+
+    def run(self, edit, base_commit, target_commit=None, file_path=None, title=None):
+        self.base_commit = base_commit
+        self.target_commit = target_commit if target_commit else "HEAD"
```
divmain: Change this to `self.target_commit = target_commit` and change in the function signature to `target_commit="HEAD"`.

core/commands/commit_compare.py
```diff
@@ -0,0 +1,245 @@
...
+        self._file_path = file_path
+        if not title:
+            title = "COMMIT COMPARISON"
+        self.title = title
```
divmain: self.title = title or "COMMIT COMPARISON"

core/commands/commit_compare.py
```diff
@@ -0,0 +1,245 @@
...
+        else:
+            file_args = []
+            diff_contents = ""
+
+        diff_contents += "Commits on {} and not on {}\n".format(target_commit, base_commit)
+        args.append("{}..{}".format(base_commit, target_commit))
+        diff_contents += self.git(*(args + file_args))
+        diff_contents = diff_contents.replace("*", COMMIT_NODE_CHAR)
+        diff_contents += "\n\nCommits on {} and not on {}\n".format(base_commit, target_commit)
+        args.pop()
```
divmain: Would much prefer taking a slice to mutation here.

core/commands/diff.py
```diff
@@ -24,14 +24,19 @@
...
+    disable Ctrl-Enter in the diff view.
     """
 
     def run(self, **kwargs):
         sublime.set_timeout_async(lambda: self.run_async(**kwargs), 0)
 
-    def run_async(self, in_cached_mode=False, file_path=None, current_file=False, base_commit=None):
+    def run_async(self, in_cached_mode=False, file_path=None,
+                  current_file=False, base_commit=None, target_commit=None, disable_stage=False,
+                  title=None):
```
divmain: The line-breaks here suggest `title` is special in some way.  Since it can't fit on one line, would prefer either even distribution of args on each subsequent line, or a line break for every arg, even thought its verbose.

popups/compare_commit_view.html
```diff
@@ -0,0 +1,23 @@
...
+  <li><code><span class="shortcut-key">f&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</span>show commits diff</code></li>
+</ul>
+
+<h3>Other</h3>
+<ul>
+  <li><code><span class="shortcut-key">?&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</span>show this help popup</code></li>
+  <li><code><span class="shortcut-key">r&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</span>refresh the page</code></li>
+</ul>
+</body>
+</html>
```
divmain: How/where are these displayed?

divmain:
Some minor feedback, but generally looks very good.

core/commands/commit_compare.py
randy3k: this does not work since other functions may pass `target_commit=None` to `gs_compare_commit`, you have to make sure that if `target_commit is None` then we set `target_commit` to HEAD.

popups/compare_commit_view.html
randy3k: When hitting the `git: compare against ...` command or `git: log` -> select commit -> `compare commit against ...`, the `commit_commit_view` is shown. And pressing `?` in `commit_commit_view` will show this popup.

core/commands/commit_compare.py
divmain: Ah, that's fair.  :+1:

divmain:
Thanks for all your work on this!

PR #482, opened 2016-07-30
Fix: match hash explictly in log_graph

randy3k:
This also fixes the navigation issue of commit comparison in log_graph and #481 .


divmain:
:+1:

PR #483, opened 2016-07-31
Enhancement: graph syntax shows author/date and highlights merge branches

randy3k:
Fix the highlight of "Merge Branches" and show date and author info.

Before

screen shot 2016-07-30 at 8 09 13 pm

After

screen shot 2016-07-31 at 1 21 32 am

The update is backward compatible to the original default
```
"git_graph_args": ["log", "--oneline", "--graph", "--decorate"]
```

Also, it changes the scope of the graph lines from comment to punctuation. As comments are italic for most color schemes and italic lines do not align.

### DIFF
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -177,8 +177,8 @@
 
         NOTE: Changes you make may break in-view functionality, so use with caution.
      */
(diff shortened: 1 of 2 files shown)


divmain:
This is a good change!  Thanks!

PR #484, opened 2016-07-31
Enhancement: rename `git: graph current branch` to `git: graph`

randy3k:
There are two reasons for such a rename.

1. `git: graph` is now consistent with `git: log` to show commits of the current branch.

2. typing `git: gr..` in command palette always select `git: graph all branches` first. However `git: graph current branch` is far more frequently used (at least for myself). Renaming it to `git: graph` makes it appear before `git: graph all branches`.

### DIFF
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -55,7 +55,7 @@
         "args": { "amend": true }
     },
     {
-        "caption": "git: graph current branch",
+        "caption": "git: graph",
         "command": "gs_log_graph"
     },
     {


divmain:
Hmm... I get what you're saying here, but 1) it seems like this is an issue of preference and not functionality, and 2) the `git: graph` command is now less explicit, i.e. it's not clear exactly what it does.  I'm not going to revert this, since this is a small change.  But it's worth having a conversation about what might be a better path going forward.

One option would be to combine both commands into a single command, `git: graph`, where you are then prompted for `Current branch` or `All branches`.  We could tweak the ordering there without losing any explicitness.

Thoughts @randy3k @stoivo?

randy3k:
I am now preparing another PR to combine `git: log` and `git: graph` since these two commands have a lot of common code. Perhaps it will be a good chance to revise the naming.

Right now, there are 
- log
- log current file
- log by branch
- log by author
- graph
- graph all branches

I guess we should somehow make sure we have the same options for `log` and `graph`.

randy3k:
I should also add that git command line also has such a ambiguity, consider `git: diff` and `git: diff branch` and `git: diff -- file`

divmain:
I have some thoughts, documented in #486.  Let's continue the conversation there!

PR #489, opened 2016-08-04
Fix: allow renamed/copied status to correctly render

### LINKED ISSUE #339: Odd listings in status view, when git interprets splitting a file as copying
The status view adds an incomplete filename:
bildschirmfoto 2016-01-09 um 17 19 26

Git in the shell looks like this:
```sh
felix@xBook:Signal-Desktop (material-design) $ git status
On branch material-design
Your branch is up-to-date with 'dharma-guardian/material-design'.
All conflicts fixed but you are still merging.
  (use "git commit" to conclude merge)

Changes to be committed:

	modified:   js/views/inbox_view.js
	copied:     stylesheets/_variables.scss -> stylesheets/_material-colors.scss
	modified:   stylesheets/_variables.scss
```

What I did in git: I tried to merge the [android-colors branch](https://github.com/dharma-guardian/Signal-Desktop/tree/android-colors) into the [material-design branch](https://github.com/dharma-guardian/Signal-Desktop/tree/android-colors) and this is a merge commit due to conflicts.

What I did in the code: I split up `_variables.scss` into `_material-colors.scss` and `_variables.scss`by manually copying parts of it to a new file.

Maybe related: I have committed identical code changes to both branches, when I wanted to isolate code from the material design branch, because cherry picking was not really an option. But I don't think this causes the problem.

asfaltboy:
- both files will now be displayed according to the rename/copy direction:
  old file -> new file

More info: https://git-scm.com/docs/git-status

Close #339

### DIFF
--- a/core/git_mixins/status.py
+++ b/core/git_mixins/status.py
@@ -21,8 +21,8 @@ class StatusMixin():
     def get_status(self):
         """
         Return a list of FileStatus objects.  These objects correspond
-        to all files that are 1) staged, 2) modified, 3) new, or 4)
-        deleted, as well as additional status information that can
+        to all files that are 1) staged, 2) modified, 3) new, 4) deleted,
+        5) renamed, or 6) copied as well as additional status information that can
         occur mid-merge.
         """
         stdout = self.git("status", "--porcelain", "-z")
@@ -36,7 +36,7 @@ class StatusMixin():
             index_status = entry[0]
             working_status = entry[1].strip() or None
             path = entry[3:]
-            path_alt = porcelain_entries.__next__() if index_status == "R" else None
+            path_alt = porcelain_entries.__next__() if index_status in ["R", "C"] else None
             entries.append(FileStatus(path, path_alt, index_status, working_status))
 
         return entries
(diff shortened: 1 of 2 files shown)


stoivo:
Who do you get it to return `C`?

else LGTM

asfaltboy:
As in the referenced doc above, `C` is a status code, used in "short format" as the output of [`self.git("status", "--porcelain", "-z")`][1]. Following this commit, we will support all valid status codes:

- ' ' = unmodified
- M = modified
- A = added
- D = deleted
- R = renamed
- C = copied
- U = updated but unmerged

[1]: https://github.com/divmain/GitSavvy/blob/master/core/git_mixins/status.py#L28

stoivo:
I have read the documentation. But I can't make it return a C what ever I do. Which command do I need to use to get to this point?

asfaltboy:
Oh, my bad I misunderstood the question.

The test case for copied status is such:
1. Take a file A with content already committed in head.
2. Move some of the file's content out into a new file B, but leave some content in.
3. Stage both:

  - the removed content in file A
  - the added content in file B

git should now call that a "copied" in staged status area.

stoivo:
Thanks @asfaltboy.

PR #495, opened 2016-08-15
[FIX] limit graph to 20k commits

### LINKED ISSUE #494: `git graph` very slow on huge codebase
I experience very slow git graph on very huge codebase.
It seems that there is missing a sensible limit on some point in order to make it more responsive.

I would argument, that you would mostly be interested in the recent history, rather than what banching has happened some 10 years ago.

EDIT: Last line of graph is:

`171535`

blaggacao:
From then on, risks are it becomes slow on huge repos. You still can adapt the setting, if need be.

Fix #494


blaggacao:
@divmain is this it?

blaggacao:
See plans on #494

Issue #486, opened 2016-08-01
Robust `git:diff` and `git: graph`


divmain:
Here are the options that I'd like to see.

- `git: log`
  - `All`
  - `Filtered by author`
    - Display list of all committers
    - When committer is selected, display commits filtered by author
  - `Filtered by commit message regexp`
    - Prompt user for regexp text
    - When text is entered, display commits filtered by commit message; `git log --grep="[Pp]rovided regexp"`
  - `Filtered by diff contents`
    - Prompt user for regexp text
    - When text is entered, display commits filtered by commit changes; `git log -G"[Pp]rovided regexp"`
- `git: log current file`
  - same options as those for `git: log`
- `git: graph`
  - `Current branch`
  - `All branches`
  - `Filtered by author`
  - `Filtered by commit message regexp`

randy3k:
i guess we could duplicate all the options of `git: log` to `git: graph`.

divmain:
Yes, that's what I intended, although it didn't come across clearly.  I've edited the description.

stoivo:
👍  for unifying these three commands in what they prompt for.

stoivo:
When we are talking about rewrite git log,  graph especially. There is many option like `--no-merges`, it would be nice to have a way to toggle these?

I have two proposals
1.
- `git: log`
  - `--no-merges (inactive)` -> would open the same prompt but change from inactive to active
  - `For current branch`
  - `For all branches`
  - `Filtered by author`
    - Display list of all committers
    - When committer is selected, display commits filtered by author
  - `Filtered by commit message regexp`
    - Prompt user for regexp text
    - When text is entered, display commits filtered by commit message; `git log --grep="[Pp]rovided regexp"`
  - `Filtered by diff contents`
    - Prompt user for regexp text
    - When text is entered, display commits filtered by commit changes; `git log -G"[Pp]rovided regexp"`
- `git: log current file`
  - same options as those for `git: log`
- `git: graph`
  - same options as those for `git: log` 

2.
 When the graph is open press `?`(ether a html view or a commad pannel) to toggle all kind of options

asfaltboy:
I believe what is left to do here is:

- [ ] Toggle off/on merge commits
- [ ] Filtered by commit message regexp
        Prompt user for regexp text
        When text is entered, display commits filtered by commit message; git log --grep="[Pp]rovided regexp"
- [ ] Filtered by diff contents
        Prompt user for regexp text
[...]

PR #505, opened 2016-09-21
Improve log, log graph, compare against, diff view and compare commit view

randy3k:
It is to follow up #486 

I haven't implemented the `all branches`, `Filtered by commit message regexp` and `Filtered by diff contents` features. However, all other functionalities should be there.

untitled


Default (Linux).sublime-keymap
```diff
@@ -112,16 +112,6 @@
     // GRAPH VIEW //
     ////////////////
```
divmain: This comment block should be removed.

core/commands/commit_compare.py
```diff
@@ -5,22 +5,20 @@
...
-from ...common import ui
-from .log import GsLogBranchCommand
 
 
 COMMIT_NODE_CHAR = "●"
 COMMIT_NODE_CHAR_OPTIONS = "●*"
 COMMIT_LINE = re.compile("[%s][ /_\|\-.]*([a-z0-9]{3,})" % COMMIT_NODE_CHAR_OPTIONS)
 
 
-class GsCompareCommitCommand(TextCommand, GitCommand):
```
divmain: Was this removed simply because it wasn't used?

core/commands/commit_compare.py
```diff
@@ -236,11 +159,39 @@ def on_branch_selection(self, index):
...
+
+        self.window.show_quick_panel(
+            options_array,
+            self.on_option_selection,
+            flags=sublime.MONOSPACE_FONT,
+            selected_index=self.quick_panel_compare_against_idx
+        )
+
+    def on_option_selection(self, index):
+        if index < 0:
```
divmain: Very small nitpick: is there any condition under which this would be `-2` or some other negative value?  I believe we're using `if index == -1` in most places, since it has an explicit meaning in the context of a quick panel selection.  Would prefer that here if there isn't an opposing reason.

core/commands/log.py
```diff
@@ -6,169 +6,74 @@
...
+    def do_action(self, commit_hash):
+        self.window.run_command("gs_log_action", {
+            "commit_hash": commit_hash,
+            "file_path": self._file_path
+        })
 
 
-class GsLogCurrentFileCommand(WindowCommand, GitCommand):
+class GsLogCurrentBranchCommand(GsLogBase):
+    pass
```
divmain: Why is this still here?

core/commands/log.py
```diff
@@ -233,4 +133,104 @@ def on_branch_selection(self, index):
...
+            "Filtered by branch",
+        ]
+        self.window.show_quick_panel(
+            options_array,
+            self.on_option_selection,
+            flags=sublime.MONOSPACE_FONT
+        )
+
+    def on_option_selection(self, index):
+        if index < 0:
```
divmain: Same question.

core/commands/log_graph.py
```diff
@@ -1,152 +1,258 @@
...
-        self.view.run_command("gs_replace_view_text", {"text": branch_graph, "nuke_cursors": True})
+        self.view.run_command("gs_replace_view_text", {"text": graph_content, "nuke_cursors": True})
         self.view.run_command("gs_log_graph_more_info")
 
         self.view.run_command("gs_handle_vintageous")
 
 
-class GsLogGraphActionCommand(TextCommand, GitCommand):
+class GsLogGraphCurrentBranch(GsLogGraphBase):
+    pass
```
divmain: ?

core/commands/log_graph.py
```diff
@@ -1,152 +1,258 @@
...
+            "Filtered by branch",
+        ]
+        self.window.show_quick_panel(
+            options_array,
+            self.on_option_selection,
+            flags=sublime.MONOSPACE_FONT
+        )
+
+    def on_option_selection(self, index):
+        if index < 0:
```
divmain: Same.

core/commands/log.py
randy3k: Originally, I wanted to define a base class and every command would be a subclass of it. It turns out that `GsLogCurrentBranchCommand` does not need extra definitions to work. 

One possible way is to rename `GsLogBase` as `GsLogCurrentBranchCommand`. But it seems strange to do

```python
class GsLogByBranchCommand(GsLogCurrentBranchCommand):
    ....

```
rather than

```python
class GsLogByBranchCommand(GsLogBase):
    ....

```

core/commands/log_graph.py
randy3k: The same reason as above.

core/commands/commit_compare.py
randy3k: I guess WindowCommand is more appropriate since we don't need anything from the active view.

randy3k:
All are fixed except for `class GsLogCurrentBranchCommand(GsLogBase)` and `GsLogGraphCurrentBranch(GsLogGraphBase)`.

`GsLogBase` and `GsLogGraphBase` are base classes. It just happens that `GsLogCurrentBranchCommand` and `GsLogGraphCurrentBranch` work out of the box because showing current branch requires no extra arguments.

divmain:
That's fair @randy3k.  Thanks again!

PR #513, opened 2016-10-21
Add: typechange to status dashboard

### LINKED ISSUE #510: Not picking up `typechange`
Gitsavvy isn't picking up the type changes in the status dashboard.

Steps to reproduce:
- create a symlink for a file
- commit the file
- delete the symlink and add a file (not a symlink) with the same name back in

After these steps, Gitsavvy reports that the directory is clean, whereas running `git status` reports that there is a `typechange` (as opposed to `modified` and others).

stoivo:
close #510

### DIFF
--- a/core/git_mixins/status.py
+++ b/core/git_mixins/status.py
@@ -105,7 +105,7 @@ class StatusMixin():
             if f.index_status == "?":
                 untracked.append(f)
                 continue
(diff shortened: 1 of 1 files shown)


divmain:
Thanks @stoivo! 🎆

PR #514, opened 2016-10-22
Fix: Always resolve symlink in self.file_path and self.repo_path

randy3k:
I always encounter random errors when the working repo is a symlink. Always resolving symlink should fix the issue.

### DIFF
--- a/core/commands/log_graph.py
+++ b/core/commands/log_graph.py
@@ -1,7 +1,6 @@
 import sublime
 from sublime_plugin import WindowCommand, TextCommand
 import re
-import os
 from ..git_command import GitCommand
 from .log import GsLogActionCommand
 from .navigate import GsNavigate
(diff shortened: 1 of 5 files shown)


core/commands/log_graph.py
```diff
@@ -39,8 +39,7 @@ def get_graph_args(self):
         savvy_settings = sublime.load_settings("GitSavvy.sublime-settings")
         args = savvy_settings.get("git_graph_args")
         if self._file_path:
-            file_path = self._file_path
-            file_path = os.path.realpath(file_path)[len(self.repo_path)+1:]
+            file_path = self.self._file_path(self._file_path)
```
stoivo: self.self?

randy3k: good catch

stoivo:
LGTM ✨

divmain:
LGTM.  Feel free to merge this @randy3k.  You may have to accept the repo invite but, once you do, you should have merge rights.

randy3k:
For consistency with diff and inline diff,I have also refactored some code to use `git_savvy.file_path` to store current file name.

PR #515, opened 2016-10-23
Fix: cycle tabs should trigger log_graph_current_branch

randy3k:


asfaltboy:
Checked out and reviewed PR (what a useful feature!!) works well!

PR #518, opened 2016-10-25
Fix: copied sha needs to be trimed

randy3k:


stoivo:
Looks good, but would like for someone to comment before we merge. I know it feels kind a stupid to do it for one line patches. Still I think we should make it as a practice and always follow it. 

Not tried to be harsh, just want to make an agreement.

asfaltboy:
Also, a little background around this would help, e.g in the body of the commit or PR

divmain:
Yep, let's wait for another reviewer to give their :+1:, except possibly in cases of emergency bug fixes.

randy3k:
👌

stoivo:
I concur

PR #519, opened 2016-10-25
Feature: Fixup commit and autosquash

### LINKED ISSUE #220: "Squash commit with previous" action in rebase view
We have "Squash commit with next", but I'd really like an equivalent of "fixup" from Git's interactive rebase mode — e.g., for squashing a small amendment into a large commit. Usually it's not possible to move the small commit up above the large one because the small commit may be fixing something that doesn't exist yet. And if we squash the large commit into the small one, we lose the large commit's commit message.

### LINKED ISSUE #503: Support for creating fixup commits
It would be useful to have support for creating the fixup commits (`git commit --fixup sha`). The UI could work by providing a list of commits in the command palette.

Questions:
 - How many commits should command palette show (should it determine the base same way as rebase view)?
 - How to determine which commits to show? Any commits that already start with [squash|fixup]! could probably be skipped as creating a commit that has two of those is in theory invalid. At least in the sense that it would make git rebase --interactive not recognize those as proper fixups.
 - Should `fixup! <sha>` or `fixup! <summary>` be created and should it be configurable? `fixup! <summary>` is preferable IMO as it's more resilient to partial rebases.

Of course GitSavvy wouldn't need to use `git commit --fixup` internally but rather just open commit view with pre-populated, valid fixup/squash summary. That would allow adding description to the commit in addition to the summary.

randy3k:
Allow fixup commit and autosquash the fixup commit. If autosquash fails, it will leave the fixup commit.

close #220 
close #503

### DIFF
--- a/core/git_mixins/rewrite.py
+++ b/core/git_mixins/rewrite.py
@@ -1,5 +1,8 @@
 import os
 from types import SimpleNamespace
+import re
+
+fixup_command = re.compile("^fixup! (.*)")
 
 
 class ChangeTemplate(SimpleNamespace):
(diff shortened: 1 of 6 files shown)


randy3k:
Basically done, not sure how to interactive with Rebase dashboard though.

stoivo:
If this is all, it look good.

randy3k:
I am good for now. I guess rebase dashboard related issues should be addressed in another PR.

divmain:
@randy3k, can you add an entry to the rebase docs?

core/git_mixins/rewrite.py
```diff
@@ -14,6 +17,47 @@ class RewriteMixin():
...
+                        if commit_chain[j].msg.startswith(orig_msg):
+                            break
+                        j = j - 1
+                    if j >= 0:
+                        commit_chain.insert(j+1, commit_chain.pop(i))
+                        commit_chain[j].do_commit = False
+                        if i - j >= 2:
+                            # if the fixup commit moves, do not decrease i
+                            continue
+                i = i - 1
```
divmain: Can you replace `i` and `j` here with variable names that represent the real values you're dealing with?  Looks like it would be something like `commit_idx`.

randy3k:
Done (I was able to use this new feature to fixup this PR.)

divmain:
Haha awesome.  This LGTM!  Merge away.

randy3k:
🎉

PR #520, opened 2016-10-25
add inline diff help popup

scott-vsi:


randy3k:
LGTM.

divmain:
This is great.  Thanks!

PR #526, opened 2016-10-27
Enhancement: refresh_gitsavvy also refreshes sidebar

randy3k:
It is particular handful when checkout other branches.


stoivo:
Clean and nice

asfaltboy:
I don't think I ever had the issue where sidebar was not refreshing, but cool that this exists 👍 !

randy3k:
Not sure why it was a problem with me. Every time I have to refresh the sidebar when checkout a branch.

Issue #531, opened 2016-10-28
Feature request: allow overriding GitSavvy settings in project settings


asfaltboy:
These "per-project" settings should take precedence over user/default GitSavvy settings when parsed. Of course, not all settings apply here, we have to refine a  list.

We already do this in some  places, just need to gather all these settings (defaulting to user/defaults) into a single entry point.

I good example of a fallback strategy can be seen in [SublimeLinter settings "stack"](http://www.sublimelinter.com/en/latest/settings.html).

randy3k:
For git repo, we have one more option to save the settings - `git config`, I think we should also put it into consideration.

divmain:
@stoivo, any thoughts on this?  I'd like to standardize on one mechanism for persisting project-specific settings.  My preference would be to go with project settings, since the options we're saving aren't pertinent outside of GitSavvy itself.

stoivo:
I think we should use project setting which will override the user package setting which again override the Package default settings. I would make it so all settings would we overwritten.

In this case I we should create a wrapper around reading settings or create a separate repo and make it a sublime depenence like [pyyaml](https://github.com/divmain/GitSavvy/blob/master/dependencies.json)

kaste:
This has been implemented (in some way).

Issue #369, opened 2016-03-14
rebase: select remote tracking branch as default


benmosher:
I often fetch + rebase against `origin/master` + push.

Assuming this is a common workflow, I thought it would be useful to auto-select the tracking remote branch in the `rebase on top of...` => `Rebase on top of branch` list.

Possible additional option, if this is _very_ common: add as a third option next to `Rebase on top of branch` and `Rebase on top of ref`: `Rebase on top of remote-tracking branch`, and auto-select it if it was the last option used. Then this workflow is reduced to `r` + `Return` on the rebase screen.

Just a thought. Curious if others find this useful.

Thanks! :smile:

PR #532, opened 2016-10-28
Feature: better selection of base ref branch

### LINKED ISSUE #499: Improve detection of root commit for branches when opening Rebase dashboard
We use a Git Flow-esque branching structure. Consequently, most of my branches are rooted on the `develop` branch; however, the Rebase dashboard always seems to guess I want to use `master` as the base ref. It would seem the logic for detecting the root commit of a branch could use some work.

The `"rebase_default_base_ref"` preference is useful for relieving the pain of this in most cases; however, the branches that _are_ rooted on `master` are then incorrectly compared to `develop` when I open the Rebase dashboard.

asfaltboy:
Before, we used a base ref that was specified in settings, otherwise defaulting to hardcoded "master".

Now, we find the nearest common ancestor branch and use that as our default base, still allowing override in settings.

Fixes #499

-----

 _(required)_

- [ ]  major
- [ ]  minor
- [ ]  patch
- [ ]  documentation only

### DIFF
--- a/core/interfaces/rebase.py
+++ b/core/interfaces/rebase.py
@@ -930,6 +930,10 @@ class GsRebaseDefineBaseRefCommand(PanelActionMixin, TextCommand, GitCommand):
         ["select_ref", "Use ref as base."],
     ]
 
+    def run(self, *args):
+        self.interface = ui.get_interface(self.view.id())
+        super().run(*args)
+
     def _get_branches(self):
         branches = [branch.name_with_remote
                     for branch in self.get_branches()
@@ -940,9 +944,11 @@ class GsRebaseDefineBaseRefCommand(PanelActionMixin, TextCommand, GitCommand):
         if branches is None:
             sublime.set_timeout_async(self._get_branches, 0)
         else:
+            base_ref = self.interface.base_ref()
             self.view.window().show_quick_panel(
                 branches,
-                filter_quick_panel(lambda idx: self.set_base_ref(branches[idx]))
+                filter_quick_panel(lambda idx: self.set_base_ref(branches[idx])),
+                selected_index=branches.index(base_ref) if base_ref in branches else None
             )
 
     def select_ref(self):
--- a/core/interfaces/rebase.py
+++ b/core/interfaces/rebase.py
@@ -312,7 +312,7 @@ class RebaseInterface(ui.Interface, NearestBranchMixin, GitCommand):
 
             if not base_ref:
                 # use remote tracking branch as a sane default
-                remote_branch = self.get_active_remote_branch().name_with_remote
+                remote_branch = self.get_upstream_for_active_branch()
                 base_ref = self.nearest_branch(self.get_current_branch_name(),
                                                default=remote_branch or "master")
                 util.debug.add_to_log('Found base ref {}'.format(base_ref))
--- a/core/interfaces/rebase.py
+++ b/core/interfaces/rebase.py
@@ -17,6 +17,7 @@ from ..ui_mixins.quick_panel import show_log_panel
 COMMIT_NODE_CHAR = "●"
 COMMIT_NODE_CHAR_OPTIONS = "●*"
 COMMIT_LINE = re.compile("\s*[%s]\s*([a-z0-9]{3,})" % COMMIT_NODE_CHAR_OPTIONS)
+NEAREST_NODE_PATTERN = re.compile(r'.*\*.*\[(.*?)(?:(?:[\^\~]+[\d]*){1})\]')  # http://regexr.com/3gm03
 
 
 def filter_quick_panel(fn):
@@ -121,7 +122,6 @@ class RebaseInterface(ui.Interface, GitCommand):
 
     _base_commit = None
     _active_conflicts = None
-    nearest_node_pattern = r'.*\*.*\[(.*?)(?:(?:[\^\~]+[\d]*)*)\]'  # http://regexr.com/3ehtn
 
     def __init__(self, *args, **kwargs):
         self.conflicts_keybindings = \
@@ -363,8 +363,7 @@ class RebaseInterface(ui.Interface, GitCommand):
 
         relatives = []
         for rel in branch_tree:
-            match = re.search(self.nearest_node_pattern, rel)
-            # print(match.groups(), branch, default)
+            match = re.search(NEAREST_NODE_PATTERN, rel)
             if not match:
                 continue
             branch_name = match.groups()[0]
--- a/core/interfaces/rebase.py
+++ b/core/interfaces/rebase.py
@@ -1,4 +1,3 @@
-import difflib
 import os
 import re
 
@@ -8,10 +7,10 @@ from sublime_plugin import WindowCommand, TextCommand
 from ...common import ui, util
 from ..commands import GsNavigate
 from ..constants import MERGE_CONFLICT_PORCELAIN_STATUSES
-from ..git_command import GitCommand
-from ..ui_mixins.quick_panel import PanelActionMixin
 from ..exceptions import GitSavvyError
-from ..ui_mixins.quick_panel import show_log_panel
+from ..git_command import GitCommand
+from ..git_mixins.rebase import NearestBranchMixin
+from ..ui_mixins.quick_panel import PanelActionMixin, show_log_panel
 
 
 COMMIT_NODE_CHAR = "●"
@@ -47,7 +46,7 @@ class GsShowRebaseCommand(WindowCommand, GitCommand):
         RebaseInterface(repo_path=self.repo_path)
 
 
-class RebaseInterface(ui.Interface, GitCommand):
+class RebaseInterface(ui.Interface, NearestBranchMixin, GitCommand):
 
     """
     Status dashboard.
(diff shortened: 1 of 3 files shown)


randy3k:
Shall we somehow combine [github integrated branch name](https://github.com/divmain/GitSavvy/blob/master/github/git_mixins/remotes.py#L3) with this?

It seems a bit inconsistant that we are using `git config` to set github remote branch while using project settings to set the rebase base.

asfaltboy:
We're not setting the base in project settings, the base is calculated on every render of the view.

The project settings are for getting a default override, which prompted me to suggest #531 , since I believe this should be in (global) default/user settings too.

As for github integrated branch name, I'm not sure there's a better way to set settings persistently in sublime.

randy3k:
The github integrate remote is also [computed](https://github.com/divmain/GitSavvy/blob/33719ef9185fd1fc8d0e77966857dad855af1666/github/git_mixins/remotes.py#L46) every time. `git config` is used to store the default.

randy3k:
Anyway, this PR looks good to me.

randy3k:
Ah....

The `git show-branch` trick does not work if that a branch has merged into the current branch.
The current code will pick the branch `subfeature`, but subfeature has been merge into the current branch (feature).

```
* [feature] Merge branch 'subfeature' into feature
 ! [master] master new commit
  ! [subfeature] subfeature commit
---
 +  [master] master new commit
-   [feature] Merge branch 'subfeature' into feature
* + [subfeature] subfeature commit
*++ [master^] before creating feature branch
```

asfaltboy:
I'm not sure there's a way to intelligently decide for the user which  of the common branches he should rebase on. As far as I know Git doesn't store the original "branching-out" node. In other words, the user could have easily merged "master" into "feature" and it would detect that. If they do merge a sub-feature branch, deleting the merged branch will allow them to rebase on the next detected base.

On the up side, it's still better than what we had (hardcoded "master"), and it can still be overridden in settings if they prefer to keep "base-ref" static.

asfaltboy:
Another option, would be to find the _oldest_ common ancestor, rather than the _nearest_, it would look something like:

```bash
git branch --contains `diff -u <(git rev-list --first-parent develop) \
    <(git rev-list --first-parent master) | sed -ne 's/^ //p' | head -1`
```

But it feels that there's still no "right" answer for all scenarios.

asfaltboy:
@randy3k can you try the command in the above comment on your test case. If it would indicate it's better for you, maybe we can allow both options with a toggle?

randy3k:
The above returns `master` for me, so it stills need to hard code `master`? I have no idea how the above works....

asfaltboy:
Sorry I copy-pasted hastily, I probably needed to update the branches like so: 

```bash
git branch --contains `diff -u <(git rev-list --first-parent feature) \
<(git rev-list --first-parent subfeature) | sed -ne 's/^ //p' | head -1` --no-merged

# or even
git branch --contains `git merge-base feature subfeature` --no-merged
```

Here is how I understand this works, hope it helps:

So, the current base detection (in this PR) works by getting all relative branches, then simply picks the first of those branches, and if it's different than self, returns it.

However, with the above suggested solution, we would not stop at "nearest", rather we would then continue and compare current branch to the potential base branch, looking for the node these trees met (oldest common commit in rev-list). Then check if there is a branch that has this commit and is unmerged, and use that branch as base ref.

randy3k:
The former one returns `master` as expected and the latter one returns nothing...

How about this?

```bash
for s in $(git rev-list --first-parent feature); do 
    branch=`git branch --contains $s --no-merged`
    [ -n "$branch" ] && echo $branch && break
done
```

asfaltboy:
The loops works nicely on my feature branches, but not at all on my master branch, for those it returns a single feature branch (unmerged!), which would be wrong as a base ref.

Correct me if I'm wrong but, the only issue we currently have with the `show-branch` solution above is if we have _many local feature branches_ that share ancestry with current branch, then we may choose those instead of their parents (since children appear before their parents in `show-branch`).

I think I'll add the filter (earliest common no-merge branches) suggested earlier, and see how that works for us.

I still don't think there is a "one size fits all" solution, so we may also want to research a bit more of how others do it.

randy3k:
Sounds a good idea. It would be also handy to preselect the corresponding branch in [GsRebaseOnTopOfCommand](https://github.com/divmain/GitSavvy/blob/master/core/interfaces/rebase.py#L873).

asfaltboy:
Ok ready for review guys! This was a bit troublesome and I want to avoid adding more functionality in this  PR. And to be clear, this is by no means a perfect solution, there may still be cases where the detected base is clearly not what the users wants.

Here's some stuff we need to address:

1. _Refreshing:_ We currently refresh the base only when building the view from scratch again. We could call `base_ref(reset_ref=True)` in `pre_render`, only when a ref is already set, to make sure we have the right ref for the selected branch. This however, will overwrite any custom base ref, meaning `GsRebaseDefineBaseRefCommand` will not work. To address this, we need to change the view setting this command uses, e.g to `user_defined_base_ref` and have it take precedence over the auto-detected / defined in settings. This will make refreshing base in `pre_render` "safe"
2. The base should be used as the pre-selected branch in `GsRebaseOnTopOfCommand`, and other commands of similar nature.
3. A setting can be added to toggle auto-detection of base, though this can already be done by the `rebase_default_base_ref` project setting.

core/interfaces/rebase.py
```diff
@@ -284,6 +296,95 @@ def base_ref(self):
...
+
+        util.debug.add_to_log('nearest_branch: filter out branches that do not'
+                              ' share branch-out nodes')
+        diff = difflib.Differ()
+        branch_commits = self.git("rev-list", "--first-parent", branch).splitlines()
+        nearest = None
+        max_iterations = 100
+        for relative in relatives:
+            util.debug.add_to_log('nearest_branch: Getting common commits with %s' % relative)
+            relative_commits = self.git("rev-list", "--first-parent", relative).splitlines()
```
randy3k: You may set the max iterations by `self.git("rev-list", "-{}".format(max_iterations), "--first-parent", relative)`

core/interfaces/rebase.py
```diff
@@ -258,13 +262,21 @@ def _get_diverged_outside_rebase(self):
...
 
-        if not base_ref:
+        if not base_ref or reset_ref:
             project_data = sublime.active_window().project_data() or {}
             project_settings = project_data.get('settings', {})
-            base_ref = project_settings.get("rebase_default_base_ref", "master")
+            base_ref = project_settings.get("rebase_default_base_ref")
+
+            if not base_ref:
+                # use remote tracking branch as a sane default
```
randy3k: sane -> same

core/interfaces/rebase.py
```diff
@@ -258,13 +262,21 @@ def _get_diverged_outside_rebase(self):
...
+                remote_branch = self._get_branch_status_components()[3]
```
randy3k: `self.get_active_remote_branch().name_with_remote` may be better

core/interfaces/rebase.py
```diff
@@ -284,6 +296,95 @@ def base_ref(self):
 
         return base_ref
 
+    def nearest_branch(self, branch, default="master"):
```
divmain: It's probably worth breaking this out into a helper method, since there's some complexity here and it isn't strictly view-related.

core/interfaces/rebase.py
```diff
@@ -284,6 +296,95 @@ def base_ref(self):
...
+        Solution snagged from:
+        http://stackoverflow.com/a/17843908/484127
+        http://stackoverflow.com/questions/1527234
+        """
+        try:
+            branch_tree = self.git("show-branch", "--no-color").splitlines()
+        except GitSavvyError:
+            return default
+
+        util.debug.add_to_log('nearest_branch: found %s show-branch results' %
```
divmain: Nit: Prefer `"... {}".format(val)` over `"... %s" % val`for consistency.

core/interfaces/rebase.py
```diff
@@ -284,6 +296,95 @@ def base_ref(self):
...
+                continue
+            branch_name = match.groups()[0]
+            if branch_name != branch and branch_name not in relatives:
+                relatives.append(branch_name)
+
+        util.debug.add_to_log('nearest_branch: found %s relatives: %s' %
+                              (len(relatives), relatives))
+
+        if not relatives:
+            util.debug.add_to_log('No relatives found. Possibly on a root branch!')
```
divmain: I like this pattern of verbose debug logging!

core/interfaces/rebase.py
```diff
@@ -284,6 +296,95 @@ def base_ref(self):
...
+                              ' share branch-out nodes')
+        diff = difflib.Differ()
+        branch_commits = self.git("rev-list", "--first-parent", branch).splitlines()
+        nearest = None
+        max_iterations = 100
+        for relative in relatives:
+            util.debug.add_to_log('nearest_branch: Getting common commits with %s' % relative)
+            relative_commits = self.git("rev-list", "--first-parent", relative).splitlines()
+            common = None
+            for i, l in enumerate(diff.compare(branch_commits, relative_commits)):
```
divmain: Prefer descriptive variable names here.

core/interfaces/rebase.py
```diff
@@ -284,6 +296,95 @@ def base_ref(self):
...
+                                  '/ detached branch!')
+            return default
+
+        util.debug.add_to_log('nearest_branch: found best candidate %s' % nearest)
+
+        # if same as branch, return default instead
+        if branch == nearest:
+            return default
+        return nearest
+
```
divmain: This is a huge function.  Would be great if it could be decomposed somehow, but its okay if that would add extra complexity or damage readability.

divmain:
Oh, and last comment.  Would like to make sure this doesn't adversely effect performance too much.  GitSavvy can already be slow, particularly on Windows machines without SSD.  I'd like to avoid making it worse if possible.  I have an older machine that I pull out for things like this, if you would like me to check this.

core/interfaces/rebase.py
asfaltboy: Do you mean into it's own helper class? Initially I had this in `BranchesMixin` since it's basically a branch detection mechanism, then I moved it here since felt it's very much related to rebase interface.

But, perhaps it's better if we move this into it's own mixin module: `git_mixins.rebase` to potentially be used by other (rebase) commands. How do you feel about this?

asfaltboy:
@divmain It does add a tiny bit of overhead when opening rebase dashboard, but it's still surprisingly fast on my machine. I would appreciate if we could run this on an old machine! We'd also need a code base with branches that have diverged significantly, to test further.

Other than that, I'd have to use this a bit more to test out more edge cases.

randy3k:
Also related to #369

#369 suggests using remote tracking branch as the base. Perhaps we can use it as a fallback instead of `master`.

asfaltboy:
It's already the default passed by the interface to the helper method. I.e `remote_branch or "master"`

core/interfaces/rebase.py
divmain: Yep, moving to `git_mixins.rebase` sounds exactly right.

divmain:
@asfaltboy, let me know when all review comments have been addressed and I'll pull this down onto an old laptop :)

stoivo:
@randy3k  still looking at this?

randy3k:
Hmmm, I could work on this if @asfaltboy agrees.

stoivo:
Ops, i though  I saw you opening this. 
Would be cool if we could close some of the long lasting PR

asfaltboy:
I keep postponing my duties, too much work during the week and weekends are busy. I'll try to address the concerns when able

randy3k:
No problem. Everyone has a real life.

core/interfaces/rebase.py
```diff
@@ -301,13 +302,21 @@ def _get_diverged_outside_rebase(self):
...
-        if not base_ref:
+        if not base_ref or reset_ref:
             project_data = sublime.active_window().project_data() or {}
             project_settings = project_data.get('settings', {})
-            base_ref = project_settings.get("rebase_default_base_ref", "master")
+            base_ref = project_settings.get("rebase_default_base_ref")
+
+            if not base_ref:
+                # use remote tracking branch as a sane default
+                remote_branch = self.get_active_remote_branch().name_with_remote
```
randy3k: `self.get_active_remote_branch()` could be `None` if there is no remote tracking, you probably only need `get_upstream_for_active_branch()`.

randy3k:
It would be more convinenent if `GsRebaseDefineBaseRefCommand` and `GsRebaseOnTopOfCommand` will preselect the chosen `base_ref`.

core/interfaces/rebase.py
```diff
@@ -301,13 +302,21 @@ def _get_diverged_outside_rebase(self):
...
-    def base_ref(self):
+    def base_ref(self, reset_ref=False):
         base_ref = self.view.settings().get("git_savvy.rebase.base_ref")
 
-        if not base_ref:
+        if not base_ref or reset_ref:
             project_data = sublime.active_window().project_data() or {}
             project_settings = project_data.get('settings', {})
-            base_ref = project_settings.get("rebase_default_base_ref", "master")
+            base_ref = project_settings.get("rebase_default_base_ref")
```
randy3k: This is irrelevant to this PR, but I would also like to discuss about this setting `rebase_default_base_ref`. It seems it is the only place we use a project setting while we are using `git config` to set GitHub integration repo and branch.

I just want to make sure there will be a consistent way for storing user settings. So which one shall we use if there is such a need? A: `git config` or B: sublime text project settings.

core/interfaces/rebase.py
randy3k: I have fixed it.

core/interfaces/rebase.py
asfaltboy: I personally don't think "sublime-specific settings" belong in `git config` ...

My project-settings files are synced to dropbox, and I'd like to avoid defining the override ref again for every new project when I switch machines. Then again, this feature kind of makes this a non-issue.

However, now that you mentioned the Github settings, maybe it could be good to store them in project settings too, unless it makes sense to use these outside of GitSavvy?

asfaltboy:
@randy3k thanks a lot for the review, and the fixes that follow!

Unless @divmain, or anyone with a slow/old/windows machine, wants to profile the rebase interface with this PR, we can go ahead and merge it into develop.

core/interfaces/rebase.py
randy3k: The `sublime-project` file might be shared within a team (though it is no likely), the `git config` is not. It might be one of the reasons why historically we were using `git config` rather than project settings.

Though, I agree that those information should not be stored at `git config` neither.

PR #533, opened 2016-10-28
Fix: diff history against active branch in Branch dashboard

stoivo:
GsCompareCommit have been changed to WindowCommand


asfaltboy:
@stoivo this is meant to fix issue 500 right? If so, please update description so it autocloses it.

stoivo:
@asfaltboy, no this only fixes `H` not the `M` key

asfaltboy:
O alright, my bad, this one's ok though!

Issue #524, opened 2016-10-27
Rebase view: editing a merge commit fails


asfaltboy:
I often edit commits in the rebase view by pressing <kbd>e</kbd>). However, merge commits fail to be edited, and the following error is displayed:

```
`/usr/local/bin/git cherry-pick --allow-empty --allow-empty-message --no-commit 7035574424b1e8a668820ece6d401d1e731ee4e0` failed with following output:

error: Commit 7035574424b1e8a668820ece6d401d1e731ee4e0 is a merge but no -m option was given.
fatal: cherry-pick failed
```

Not sure if we simply use `-m` option here.

related to #426

randy3k:
The same bug also occurs when moving/ squshing (revoking `rewrite_active_branch`) if there is a merge commit.

The current implementation of `rewrite_active_branch` does not handle merge commits correctly.

randy3k:
Just gave it a thought, it is difficult to fix. We cannot just use the `-m` option, as it will convert the merge commit to a normal single-parent commit...

asfaltboy:
If I read [`git-rebase` code here][1] and [also here][2] correctly, we should fork on whether this is a merge commit and use either `cherry-pick `or `git-merge` (passing in a list of parents and the new commit message) accordingly.

[1]: https://github.com/git/git/blob/master/git-rebase--interactive.sh#L1160-L1180
[2]: https://github.com/git/git/blob/master/git-rebase--interactive.sh#L386-L387

randy3k:
Sounds like right. It is basically a rewrite of `rebase -i -p`. 

related to #297, #452

divmain:
Merge commits have historically been really difficult to manage properly in the rebase dashboard.  Would love to see this addressed, but want to avoid adding the functionality at the cost of greatly increased implementation complexity.

asfaltboy:
I'm not sure we can (or even want) to allow every action for a merge commit, but editing commit message should be possible, I do it myself a lot.

But I can't imagine how, for example, one could move a merge commit up/down .. unless the whole sub-tree is moved with it. 

In any case, I think merge commits are pretty simple to preserve, we simply need to detect it's a merge (get multiple parents) and redo the merge, instead of the cherry pick. Here's a list of relevant actions and what I suggest we do with merge commits:

- [ ] <kbd>q</kbd> squash commit with previous -- _remove  merge commit_
- [ ] <kbd>Q</kbd> squash all commits -- _remove merge commit_
[...]

Issue #452, opened 2016-07-13
Rebase moving without action


p3lim:
When I want to move commits around I only have the option to move them up/down one at a time.
Sometimes I need to move certain commits to a certain place, or squash with a certain commit, and moving them one position at a time can yield conflicts.

Adding an option to select a commit (or multiple) for any action before commiting (basically queuing) them would be nice.

For displaying this, you could use a different color for the ref, and/or indent them further, just thoughts.

divmain:
Can you give me an example of a time when moving a commit several places would not result in a merge conflict, but moving one spot at a time would result in a merge conflict?  I assume it'd be something you changed and then changed back in the context of your branch?

We've thrown around a queuing mechanism before, and while it sounds like a useful feature, I'm not really sure how it would function in the context of the dashboard.  It might be worth building something closer to `git rebase -i`, or maybe even providing a shortcut to use `git rebase -i` itself.

p3lim:
When there are hunks close to each other would be a good example.

divmain:
Ah that's an irritatingly solid example.  OK.  How do you feel about farming this work out to `git rebase -i`?  That may be hard to integrate well with GitSavvy/Sublime - Git makes certain assumptions about how that workflow should happen (namely, you're doing it on the terminal and it will open a text editor in your terminal, full stop).

We could try to create some sort of "stage changes" mode, where you move commits around, drop commits, etc, but they won't take effect until you accept all of the changes (with an option to abort).  It would add significant complexity to the rebase dashboard, which is a downside, and would also diverge from the real-time behavior, which is also a downside (unless we dump realtime behavior altogether).

In any case, this will be a largish task.  If you have UX ideas, or anything that might make for a good middle-ground, I'm all ears!

p3lim:
> (namely, you're doing it on the terminal and it will open a text editor in your terminal, full stop)

Setting the gitconfig flag `core.editor` to `subl -nw` works fine, it's what I've used on Windows forever, and it is something you can do on a non-global basis, as long as you have SublimeText as part of the path (or just use the full path).

> We could try to create some sort of "stage changes" mode, where you move commits around, drop commits, etc, but they won't take effect until you accept all of the changes (with an option to abort).

Sounds perfect to me, the dashboard has the (UX) potential of normal `rebase -i` operation.

I would help on the matter, but I've fallen behind on the packages internals, and don't have much time to work on such a large task in the foreseeable future.

divmain:
[...]

PR #534, opened 2016-10-29
Feature: upgrade rebase dashboard

### LINKED ISSUE #297: Queue up actions in Rebase dashboard

### LINKED ISSUE #426: Rebasing should support keeping merge commits
Today if we rebase (e.g over another branch) we lose the merge commits as this default git behaviour. To overcome this, I use a [command line argument `-p` to preserve merge commits](http://explainshell.com/explain?cmd=git+rebase+-p).

I would also go the extra step and make it a setting so that people can switch to the behaviour they prefer, as I assume without merge commits the history is cleaner and thus less prone to duplicate commits appearing in the tree (when merging big trees?).

### LINKED ISSUE #560: Rebase: status incorrect after failure
Trying to perform a change that fails, e.g because it causes a conflict, results in a SUCCESS message even though the change failed.

Reproduction steps:
1. Commit a change
2. Commit another change in closely related line.
3. Try moving the second commit "UP" by hitting <kbd>u</kbd>
4. "STATUS" should not change, but it shows e.g:
    > Successfully moved 6435b68 up. Undo available.
5. Nothing else (hash/order) seems to change.

[error-log.json.zip](https://github.com/divmain/GitSavvy/files/578721/error-log.json.zip)

randy3k:
screen shot 2016-10-31 at 12 47 43 am

- [x] Feature: Preserve merges rebase: use <kbd>m</kbd> to activate preserve merges mode in rebase dashboard. Fix #426 #524
- [x] Feature: Squash to any previous commits
- [x] Enhacement: Fixup command can handle merges
- [x] Feature: move commits to arbitrary location in rebase dashboard.  Fix  #297 #452 #561 
- [x] Fix: update rebase log only when success, Fix #560

---
This change is [Reviewable](https://reviewable.io/reviews/divmain/gitsavvy/534)

### DIFF
--- a/core/interfaces/rebase.py
+++ b/core/interfaces/rebase.py
@@ -1012,40 +1012,14 @@ class GsRebaseNavigateCommitsCommand(GsNavigate):
     If a commit has conflicts, navigate to the next (or previous) file.
     """
 
-    offset_map = {
-        "meta.git-savvy.rebase-graph.entry": 7,
-        "meta.git-savvy.rebase-graph.conflict": 16,
-    }
-
-    def run(self, edit, forward=True):
-        sel = self.view.sel()
-        if not sel:
-            return
-
-        current_position = sel[0].a
-
-        available_regions = self.get_available_regions()
-
-        new_position = (self.forward(current_position, available_regions)
-                        if forward
-                        else self.backward(current_position, available_regions))
-
-        if new_position is None:
-            return
-
-        offset = 7  # default to commit offset, conflict is always after
-        next_context = self.view.scope_name(new_position)
-        for scope in next_context.split():
-            if scope in self.offset_map:
-                offset = self.offset_map[scope]
-
-        # Position the cursor at next/previous commit or conflict filename
-        sel.clear()
-        new_position += offset
-        sel.add(sublime.Region(new_position, new_position))
+    offset = 0
 
     def get_available_regions(self):
-        regions = [region for selector in self.offset_map.keys()
-                   for region in self.view.find_by_selector(selector)]
-        return sorted([
-            line for region in regions for line in self.view.lines(region)])
+        commit_selector = "meta.git-savvy.rebase-graph.entry support.type.git-savvy.rebase.commit_hash"
+        conflict_selector = "meta.git-savvy.rebase-graph.conflict keyword.other.name.git-savvy.rebase-conflict"
+
+        regions = self.view.find_by_selector(conflict_selector)
+        if len(regions) == 0:
+            regions = self.view.find_by_selector(commit_selector)
+
+        return regions
--- a/core/interfaces/rebase.py
+++ b/core/interfaces/rebase.py
@@ -8,6 +8,7 @@ from ..commands import GsNavigate
 from ..constants import MERGE_CONFLICT_PORCELAIN_STATUSES
 from ..git_command import GitCommand
 from ..ui_mixins.quick_panel import PanelActionMixin
+from ..exceptions import GitSavvyError
 
 
 def filter_quick_panel(fn):
@@ -55,7 +56,7 @@ class RebaseInterface(ui.Interface, GitCommand):
 
     template = """\
 
-      REBASE:  {active_branch} --> {base_ref} ({base_commit})
+      REBASE:  {active_branch} --> {base_ref} ({base_commit}){preserve_merges}
       STATUS:  {status}
 
         ┬ ({base_commit})
@@ -76,10 +77,10 @@ class RebaseInterface(ui.Interface, GitCommand):
 
       [q] squash commit with previous           [f] define base ref for dashboard
       [Q] squash all commits                    [r] rebase branch on top of...
-      [e] edit commit message                   [c] continue rebase
-      [p] drop commit                           [{skip_commit_key}] skip commit during rebase
-      [d] move commit down (after next)         [A] abort rebase
-      [u] move commit up (before previous)
+      [e] edit commit message                   [m] toggle preserve merge mode
+      [p] drop commit                           [c] continue rebase
+      [d] move commit down (after next)         [{skip_commit_key}] skip commit during rebase
+      [u] move commit up (before previous)      [A] abort rebase
       [w] show commit
 
       [?]         toggle this help menu
@@ -106,7 +107,6 @@ class RebaseInterface(ui.Interface, GitCommand):
 
     separator = "\n    │\n"
     commit = "  {caret} {status}  {commit_hash}  {commit_summary}{conflicts}"
-    conflict = "    │           conflict: {path}"
 
     _base_commit = None
     _active_conflicts = None
@@ -166,7 +166,7 @@ class RebaseInterface(ui.Interface, GitCommand):
         try:
             cursor_entry = log[cursor]
         except IndexError:
-            return "Ready."
+            return "Not yet rebased." if self.not_yet_rebased() else "Ready."
 
         if cursor == log_len - 1:
             return "Successfully {}. Undo available.".format(cursor_entry["description"])
@@ -185,6 +185,13 @@ class RebaseInterface(ui.Interface, GitCommand):
     def render_super_key(self):
         return util.super_key
 
+    @ui.partial("preserve_merges")
+    def render_preserve_merge(self):
+        if self.preserve_merges():
+            return " (Preserving merges)"
+        else:
+            return ""
+
     @ui.partial("help")
     def render_help(self):
         help_hidden = self.view.settings().get("git_savvy.help_hidden")
@@ -197,8 +204,16 @@ class RebaseInterface(ui.Interface, GitCommand):
                 conflicts_bindings=self.render_conflicts_bindings(),
                 skip_commit_key='k' if not vintageous_friendly else 'K')
 
+    def preserve_merges(self):
+        savvy_settings = sublime.load_settings("GitSavvy.sublime-settings")
+        default = savvy_settings.get("rebase_preserve_merges")
+        return not self.in_rebase_apply() and \
+            (self.in_rebase_merge() or
+             self.view.settings().get("git_savvy.rebase.preserve_merges", default))
+
     def get_diverged_commits_info(self, start, end):
-        self.entries = self.log(start_end=(start, end), reverse=True)
+        preserve = self.preserve_merges()
+        self.entries = self.log_rebase(start, end, preserve)
         return (self._get_diverged_in_rebase()
                 if self._in_rebase else
                 self._get_diverged_outside_rebase())
@@ -206,21 +221,39 @@ class RebaseInterface(ui.Interface, GitCommand):
     def _get_diverged_in_rebase(self):
         self._active_conflicts = None
         conflict_commit = self.rebase_conflict_at()
-        rewritten = dict(self.rebase_rewritten())
+        rewritten = self.rebase_rewritten()
         commits_info = []
 
         for entry in self.entries:
+            conflicts = ""
             was_rewritten = entry.long_hash in rewritten
             new_hash = rewritten[entry.long_hash][:7] if was_rewritten else None
-            is_conflict = entry.long_hash == conflict_commit
-
-            if is_conflict:
-                self._active_conflicts = self._get_conflicts_in_rebase()
-                conflicts = (
-                    "" if not self._active_conflicts else
-                    "\n" + "\n".join("    │           ! {}".format(conflict.path)
-                                     for conflict in self._active_conflicts)
-                    )
+            is_merge = self.commit_is_merge(entry.long_hash)
+            if self.in_rebase_merge() and is_merge:
+                is_conflict = conflict_commit in self.commits_of_merge(entry.long_hash)
+                if is_conflict:
+                    conflict_logs = self.log_merge(entry.long_hash)
+                    for conflict_idx, c in enumerate(conflict_logs):
+                        conflicts = conflicts + "\n    │    {}  {}  {}".format(
+                            self.SUCCESS if c.long_hash in rewritten else
+                            self.CONFLICT if c.long_hash == conflict_commit else
+                            self.UNKNOWN,
+                            c.short_hash,
+                            c.summary)
+                        if c.long_hash == conflict_commit:
+                            self._active_conflicts = self._get_conflicts_in_rebase()
+                            if self._active_conflicts:
+                                conflicts = conflicts + "\n" + "\n".join(
+                                    "    │           ! {}".format(conflict.path)
+                                    for conflict in self._active_conflicts)
+            else:
+                is_conflict = entry.long_hash == conflict_commit
+                if is_conflict:
+                    self._active_conflicts = self._get_conflicts_in_rebase()
+                    if self._active_conflicts:
+                        conflicts = conflicts + "\n" + "\n".join(
+                            "    │           ! {}".format(conflict.path)
+                            for conflict in self._active_conflicts)
 
             commits_info.append({
                 "caret": self.CARET if is_conflict else " ",
@@ -291,6 +324,10 @@ class RebaseInterface(ui.Interface, GitCommand):
         base_ref = self.base_ref()
         self._base_commit = self.git("merge-base", "HEAD", base_ref).strip()
         return self._base_commit
+
+    def not_yet_rebased(self):
+        return self.git("merge-base", self.base_commit(), "HEAD").strip() != \
+                self.git("rev-parse", self.base_ref()).strip()
 
     def get_branch_ref(self, branch_name):
         stdout = self.git("show-ref", "refs/heads/" + branch_name)
@@ -412,8 +449,26 @@ class RewriteBase(TextCommand, GitCommand):
     Base class for all commit manipulation actions.
     """
 
+    # use to remap parents of merges
+    commit_parent_map = {}
+
     def run(self, edit):
         self.interface = ui.get_interface(self.view.id())
+
+        if self.interface.not_yet_rebased():
+            sublime.message_dialog(
+                "Unable to manipulate commits, rebase first."
+            )
+            return
+
+        _, _, changed_files = self.interface.get_branch_state()
+
+        if len(changed_files):
+            sublime.message_dialog(
+                "Unable to manipulate commits while repo is in unclean state."
+            )
+            return
+
         sublime.set_timeout_async(self.run_async, 0)
 
     def get_selected_short_hash(self):
@@ -441,12 +496,6 @@ class RewriteBase(TextCommand, GitCommand):
         branch_name, ref, changed_files = self.interface.get_branch_state()
         success = True
 
-        if len(changed_files):
-            sublime.message_dialog(
-                "Unable to perform rebase actions while repo is in unclean state."
-            )
-            return
-
         try:
             self.rewrite_active_branch(
                 base_commit=base_commit,
@@ -461,6 +510,10 @@ class RewriteBase(TextCommand, GitCommand):
             self.interface.complete_action(branch_name, ref, success, description)
 
         util.view.refresh_gitsavvy(self.view, refresh_sidebar=True)
+
+    def commit_parents(self, commit):
+        return [self.commit_parent_map[p] if p in self.commit_parent_map else p
+                for p in super().commit_parents(commit)]
 
 
 class GsRebaseSquashCommand(RewriteBase):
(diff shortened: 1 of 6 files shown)


randy3k:
Please review.

asfaltboy:
Sorry for the delay, real life stuff need my attention, will review once get a chance.

randy3k:
Take your time. Anyone has their own thing to take care.

docs/rebase.md
```diff
@@ -83,6 +95,10 @@ When running this command, you will be prompted for the branch on top of which y
 
 If any merge conflicts occur while rebasing, the state of the operation will be reflected in the dashboard, with successful rebased commits showing a check mark, and the commit with the current merge conflict showing an X as well as a list of files with conflicts.
 
+#### Toggle perserve merges mode (`m`)
+
+In default, merges will be lost when rebasing. If perserve merges mode is on, commits within a merge are consolidated in a single item. If any conflicts occur while rebasing a merge commit, the commits within the merge will be shown and the conflicting commit will be marked.
```
asfaltboy: I'm not sure I like the idea of consolidating all commits in a single item. I want to avoid implicitly squashing any commits if possible.

docs/rebase.md
```diff
@@ -23,7 +23,7 @@ Commits are shown in order of application, where the base commit (or inferred ro
 
 #### Squash commit with next (`q`)
 
-The selected commit will be squashed with the commit below it.  The two commit messages will be combined into the one below.
+A list of commits will be displayed, the commit under cursor will be squashed with the selected commit.  The two commit messages will be combined into the one.
```
asfaltboy: I kind of liked the previous 1 key squash. Would it make sense to make this a separate command/shortcut?

docs/rebase.md
randy3k: The commits within a merge are preserved, they are not actually squashed to together. It should be better phrased: 

```text
No commits will be implicitly squashed, commits within a merge are just *displayed* as a single item to avoid commits accidentally moving in or moving out of a merge.
```

core/interfaces/rebase.py
```diff
@@ -475,34 +532,45 @@ def run_async(self):
...
+            sublime.status_message("Unable to squash merges.")
+            return
+
+        self.squash_idx = squash_idx
+        self.squash_entry = squash_entry
+        reversed_logs = list(reversed(self.interface.entries[0:squash_idx]))
+
+        # GsLogBase is a WindowCommand
+        self.window = self.view.window()
+        self.display_commits(self.render_commits(reversed_logs))
```
randy3k: AFAIK, `git_mixins.history` contains only git related code, all UI related codes go to `core.command`.

@divmain Any idea where to put UI related common codes?

docs/rebase.md
randy3k: How about

- <kbd>q</kbd> stays what it as
- <kbd>Q</kbd> squashes into a selected commit
- <kbd>P</kbd> to squash all commits (the original key for this is <kbd>Q</kbd>) 
- move up and down keys have similar structure.

asfaltboy: <kbd>q</kbd> and <kbd>Q</kbd> always seemed strange to me, any reason we can't use <kbd>s</kbd> and <kbd>S</kbd> instead?

docs/rebase.md
asfaltboy: Oh! I see what you mean now. Hmm, any chance this can be toggled so that I can (if I want) to e.g edit a commit message that is on a merged branch?

randy3k: Your could edit the merge message of a merge commit in the current code.  
I argue that if you really need to edit a commit of a merged branch, you should use the non-preserving mode.

Manipulating commits within a merge is tricky, particularly if there are nested merges.
Although, editing commit messages within a merge can be made possible, but it will be a tedious job.

asfaltboy: I see, so without the preserving mode I will see all the commits "behind a merge", but I'd lose the merge commit. I guess I'd have to keep using "rebase -i -p" for these rare cases.

randy3k:
Actually, the doc of `git rebase -p` says

> This uses the --interactive machinery internally, but combining it with the --interactive option explicitly is generally not a good idea unless you know what you are doing (see BUGS below).

docs/rebase.md
randy3k: **REPOST** my mobile app posted in a wrong location.

Actually, the doc of `git rebase -p` says

> This uses the --interactive machinery internally, but combining it with the --interactive option explicitly is generally not a good idea unless you know what you are doing (see BUGS below).

randy3k:
Depends on #539 #540 #541

divmain:
Sorry I've taken time to review this - real life has interrupted for the last couple of weeks.  I definitely want to give my input before this is merged, so hold on for a bit.

asfaltboy:
Reviewed 1 of 5 files at r1, 4 of 5 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[GitSavvy.sublime-settings, line 205 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWi5RdU32xOFKRXZeBc:-KWi5RdU32xOFKRXZeBd:bdjrpxa) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/GitSavvy.sublime-settings#L205)):*
> ```
>         When set to `true`, rebase dashboard uses preserve-merges mode when opened.
>     */
>     "rebase_preserve_merges": false,
> ```

Toggle setting is a great idea, I would prefer it to be true by default, since it is the less destructive of the two options. The other noticeable change would be that any merges into current branch would appear "collapsed". @divmain you ok with this being `true` by default?

---

*[core/commands/fixup.py, line 46 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWqrYSAv04osUx8IspW:-KWqrYSAv04osUx8IspX:b-3pacfn) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/core/commands/fixup.py#L46)):*
> ```Python
>                 commit_chain[orig_commit_indx + 1].datetime = commit_chain[orig_commit_indx].datetime
>                 commit_chain[orig_commit_indx + 1].author = commit_chain[orig_commit_indx].author
>                 commit_chain[orig_commit_indx + 1].modified = True
> ```

I realize this is basically copy-paste from `core/git_mixins/rewrite.py`, but, it would be nice if we refactor this for readability by e.g naming `original_commit = commit_chain[orig_commit_indx]` and `next_commit = commit_chain[orig_commit_indx + 1]` , etc.

---

*[core/git_mixins/rewrite.py, line 38 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWiJVfxAgZMBzXqO2Js:-KWiJVfxAgZMBzXqO2Jt:b7c8ao0) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/core/git_mixins/rewrite.py#L38)):*
> ```Python
>             RewriteTemplate(orig_hash=entry.long_hash,
>                             msg=entry.raw_body,
>                             do_commit=True,
> ```

Unnecessary change?

---

*[core/git_mixins/rewrite.py, line 241 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWiFQeLM7W9bIEgoFdU:-KWiFQeM5Dl5ozjN6Kyg:b-cotirv) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/core/git_mixins/rewrite.py#L241)):*
> ```Python
>                 with open(path, "r") as f:
>                     entries = f.read().strip().split("\n")
>                     return (entry.split(" ") for entry in entries)
> ```

Should this function return a generator? The other cases always return a dictionary, doesn't the calling function expect only a dict?

---

*[core/interfaces/rebase.py, line 353 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWiHME_IQfJtyu4E95u:-KWiHME_IQfJtyu4E95v:bx0tuvw) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/core/interfaces/rebase.py#L353)):*
> ```Python
>         log = self.view.settings().get("git_savvy.rebase_log") or []
>         cursor = self.view.settings().get("git_savvy.rebase_log_cursor") or (len(log) - 1)
>         if success:
> ```

Will logging only a successful rebase still allow to undo any "partially successful" rebase, were some kind of issue to occur?

---

*[core/interfaces/rebase.py, line 464 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWiHzOJEDqBKExE-Pv6:-KWiHzOK5QdbUGsR2AMk:blj8ae8) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/core/interfaces/rebase.py#L464)):*
> ```Python
> 
>         if self.interface.is_not_rebased() or \
>                 (not self.interface.preserve_merges() and self.interface.contain_merges()):
> ```

We should probably mention in the docs why rebase is not allowed in some cases, and how this can be fixed. E.g "Manipulating commits is not allowed when the current branch is not connected to the base-ref. Please select another base-ref or rebase first." or something in those lines.

In addition, I had a few cases where I had to close+reopen the rebase interface in order to make the base-ref trigger the state to "Ready". Not sure if there's a bug here, or just some missing refresh state (call `render_status` ?)

---

*[core/interfaces/rebase.py, line 494 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWquwlMyNmb_Lfv180F:-KWquwlMyNmb_Lfv180G:bvqknfm) ([raw file](https://github.com/divmain/gitsavvy/blob/2d41dec240e07434502206407373a9425a3c3631/core/interfaces/rebase.py#L494)):*
> ```Python
>         # not be commited again.  However, the second commit (the selected) must include
>         # the diff from the first, and all the meta-data for the squashed commit must
>         # match the first.
> ```

Do we really need to remove the comment? Maybe shorten / simplify it, instead of removing?

---

*[core/interfaces/rebase.py, line 506 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWiJH_ASYp4c1OdC3IN:-KWiJH_ASYp4c1OdC3IO:b7c8ao0) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/core/interfaces/rebase.py#L506)):*
> ```Python
>     def make_changes(self, commit_chain, description, base_commit=None):
>         branch_name, ref, changed_files = self.interface.get_branch_state()
>         base_commit = base_commit or self.interface.base_commit()
> ```

Unnecessary change?

---

*[core/interfaces/rebase.py, line 605 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWqu5L3RGRPn4ZyrGw1:-KWqu5L3RGRPn4ZyrGw2:b7c8ao0) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/core/interfaces/rebase.py#L605)):*
> ```Python
>         self.make_changes(
>             commit_chain,
>             "squashed all commits")
> ```

Unnecessary change?

---

*[core/interfaces/rebase.py, line 1025 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWqtNFVsSDWkQx1TGe8:-KWqtNFVsSDWkQx1TGe9:b-v0fekh) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/core/interfaces/rebase.py#L1025)):*
> ```Python
>             regions = self.view.find_by_selector(selectors)
> 
>         return regions
> ```

Love this refactoring. Thanks!

---

*[docs/rebase.md, line 30 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWiAKusSk45US90VIPK:-KWiAKuti94qHYg15a-7:br5mgy2) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/docs/rebase.md#L30)):*
> ```Markdown
> **Note:** This command cannot be performed while a rebase is ongoing.
> 
> #### Squash commit with a previous commit (`Q`)
> ```

Should this be "Squash commit with another commit" ? I mean does it have to be a predating commit to the one selected?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/divmain/gitsavvy/534)*

stoivo:
👍  for @asfaltboy comments.

randy3k:
I may not have time to modify the commits until next week.

---

Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[core/commands/fixup.py, line 46 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWqrYSAv04osUx8IspW:-KWrzMavqeWkWapARd32:b-gqs144) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/core/commands/fixup.py#L46)):*
<details><summary><i>Previously, asfaltboy (Pavel Savchenko) wrote…</i></summary>
> I realize this is basically copy-paste from `core/git_mixins/rewrite.py`, but, it would be nice if we refactor this for readability by e.g naming `original_commit = commit_chain[orig_commit_indx]` and `next_commit = commit_chain[orig_commit_indx + 1]` , etc.
</details>
will change it.

---

*[core/git_mixins/rewrite.py, line 38 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWiJVfxAgZMBzXqO2Js:-KWrzW2qYbzjGgFUOu3w:biak3vx) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/core/git_mixins/rewrite.py#L38)):*
<details><summary><i>Previously, asfaltboy (Pavel Savchenko) wrote…</i></summary>
> Unnecessary change?
</details>
I guess there is no need to expose the class `ChangeTemplate` in the Mixin. The renaming may not be necessary. I could also undo it.

---

*[core/git_mixins/rewrite.py, line 241 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWiFQeLM7W9bIEgoFdU:-KWrxplXlg8NJ4VXVAPV:b1mch50) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/core/git_mixins/rewrite.py#L241)):*
<details><summary><i>Previously, asfaltboy (Pavel Savchenko) wrote…</i></summary>
> Should this function return a generator? The other cases always return a dictionary, doesn't the calling function expect only a dict?
</details>
you are right.

---

*[core/interfaces/rebase.py, line 353 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWiHME_IQfJtyu4E95u:-KWrxxP80M019V-X771L:b-a8zy9q) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/core/interfaces/rebase.py#L353)):*
<details><summary><i>Previously, asfaltboy (Pavel Savchenko) wrote…</i></summary>
> Will logging only a successful rebase still allow to undo any "partially successful" rebase, were some kind of issue to occur?
</details>
Currently, all commit manipulation is all or nothing, "partially successful" should not be an issue. It may be an issue later if we allow resolving conflicts in commit manipulation.

---

*[core/interfaces/rebase.py, line 464 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWiHzOJEDqBKExE-Pv6:-KWrzx3f3FUCedMuw1b7:b-aqc3x5) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/core/interfaces/rebase.py#L464)):*
<details><summary><i>Previously, asfaltboy (Pavel Savchenko) wrote…</i></summary>
> We should probably mention in the docs why rebase is not allowed in some cases, and how this can be fixed. E.g "Manipulating commits is not allowed when the current branch is not connected to the base-ref. Please select another base-ref or rebase first." or something in those lines.
> 
> In addition, I had a few cases where I had to close+reopen the rebase interface in order to make the base-ref trigger the state to "Ready". Not sure if there's a bug here, or just some missing refresh state (call `render_status` ?)
</details>
agree.

the "Ready" issue needs further investigation.

---

*[core/interfaces/rebase.py, line 494 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWquwlMyNmb_Lfv180F:-KWryelHr_6ri3K8Iu-o:b-txpote) ([raw file](https://github.com/divmain/gitsavvy/blob/2d41dec240e07434502206407373a9425a3c3631/core/interfaces/rebase.py#L494)):*
<details><summary><i>Previously, asfaltboy (Pavel Savchenko) wrote…</i></summary>
> Do we really need to remove the comment? Maybe shorten / simplify it, instead of removing?
</details>
will undo it

---

*[core/interfaces/rebase.py, line 506 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWiJH_ASYp4c1OdC3IN:-KWs-7J_ejhLs98EY5eq:b-s4hxqy) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/core/interfaces/rebase.py#L506)):*
<details><summary><i>Previously, asfaltboy (Pavel Savchenko) wrote…</i></summary>
> Unnecessary change?
</details>
yes, I removed the line when refactorizing and then realized that the line is needed.

---

*[core/interfaces/rebase.py, line 605 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWqu5L3RGRPn4ZyrGw1:-KWryqS9XMz5APjdseem:b-jrta9n) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/core/interfaces/rebase.py#L605)):*
<details><summary><i>Previously, asfaltboy (Pavel Savchenko) wrote…</i></summary>
> Unnecessary change?
</details>
Do you mean the newlines? I could undo them.

---

*[docs/rebase.md, line 30 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWiAKusSk45US90VIPK:-KWrwWhcym8IrzL-wHOh:bh5xmb4) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/docs/rebase.md#L30)):*
<details><summary><i>Previously, asfaltboy (Pavel Savchenko) wrote…</i></summary>
> Should this be "Squash commit with another commit" ? I mean does it have to be a predating commit to the one selected?
</details>
I have no preferences. But I find that  `squash commit with a previous commit` is a natural extension of the key `q` as from `u` to `U`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-:-KWs-OVSyb6aPMYmDi3i:b-i7y17i)*

asfaltboy:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[core/git_mixins/rewrite.py, line 38 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWiJVfxAgZMBzXqO2Js:-KWs155N07zs2aVgadoK:ba46kq4) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/core/git_mixins/rewrite.py#L38)):*
<details><summary><i>Previously, randy3k (Randy Lai) wrote…</i></summary>
> I guess there is no need to expose the class `ChangeTemplate` in the Mixin. The renaming may not be necessary. I could also undo it.
</details>
Sorry, I only meant the swapping of the 2 lines `do_commit=True` and `msg=entry.raw_body`. Removing the redundant `self.ChangeTemplate` class attribute is totally OK!

---

*[core/interfaces/rebase.py, line 353 at r3](https://reviewable.io:443/reviews/divmain/gitsavvy/534#-KWiHME_IQfJtyu4E95u:-KWs3GrSf0w8tnJ6R4Vg:b-s091i1) ([raw file](https://github.com/divmain/gitsavvy/blob/301cef06068a82936586bb27c98fa598bc711e06/core/interfaces/rebase.py#L353)):*
<details><summary><i>Previously, randy3k (Randy Lai) wrote…</i></summary>
> Currently, all commit manipulation is all or nothing, "partially successful" should not be an issue. It may be an issue later if we allow resolving conflicts in commit manipulation.
</details>
Sounds good, but can you add this note as a comment here for clarity?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/divmain/gitsavvy/534)*

randy3k:
@asfaltboy's comment should have been addressed. One remaining issue:
Shall we set `rebase_preserve_merges` default to `true`? It definitely changes the rebase behaviour, but as mentioned as @asfaltboy, it is less destructive to preserve merges.

randy3k:
@divmain How do you feel about this? I guess this is about time to release a new version of GitSavvy. I would like to include this PR with the new tag.

divmain:
This is really fantastic work @randy3k.  Thanks a ton, and thanks for being so patient.  Merging! 🎉

divmain:
@randy3k, for future reference, if you use PRs to automatically close issues, you need to list each one individually.  I.e. `Fixes #426. Fixes #524.` etc.

PR #535, opened 2016-10-29
Fix: Send list to merge

### LINKED ISSUE #500: [M] fetch and merge into active branch
```
`/usr/bin/git merge o r i g i n / d e v e l o p` failed with following output:

merge: o r i g i n / d e v e l o p - not something we can merge
```

peek 2016-09-12 11-43

stoivo:
Close #500

### DIFF
--- a/core/interfaces/branch.py
+++ b/core/interfaces/branch.py
@@ -557,7 +557,7 @@ class GsBranchesFetchAndMergeCommand(TextCommand, GitCommand):
             if remote_region and remote_region[0].contains(selection):
                 sublime.status_message("Fetching from `{}`...".format(remote_name))
                 self.fetch(remote=remote_name)
-                self.merge("{}/{}".format(remote_name, branch_name))
+                self.merge(list("{}/{}".format(remote_name, branch_name)))
                 sublime.status_message("Fetch and merge complete.")
                 util.view.refresh_gitsavvy(self.view)
                 return
--- a/core/interfaces/branch.py
+++ b/core/interfaces/branch.py
@@ -502,18 +502,23 @@ class GsBranchesMergeSelectedCommand(TextCommand, GitCommand):
 
     def run_async(self):
         self.interface = ui.get_interface(self.view.id())
-        current_branch_name = self.get_current_branch_name()
 
+        branches = self.get_selected_branches()
+        branches_strings = self.create_branches_strs(branches)
+        self.merge(branches_strings)
+        util.view.refresh_gitsavvy(self.view)
+
+    def get_selected_branches(self):
+        current_branch_name = self.get_current_branch_name()
         branches = set()
         for sel in self.view.sel():
-            for name in self.get_selected_branches(sel):
+            for name in self.get_selected_branch(sel):
                 if name and not name == current_branch_name:
                     branches.add(name)
 
-        self.merge(branches) if not len(branches) == 0 else None
-        util.view.refresh_gitsavvy(self.view)
+        return branches
 
-    def get_selected_branches(self, sel):
+    def get_selected_branch(self, sel):
         lines = util.view.get_lines_from_regions(self.view, [sel])
         return (self.get_selected_branch_name(sel, line) for line in lines)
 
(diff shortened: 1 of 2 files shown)


randy3k:
It is still not working if multiple branches are highlighted.

It seems that [this](
https://github.com/divmain/GitSavvy/blob/e9bd976a8e20b4bbec29602468557d2dcc89b245/core/git_mixins/branches.py#L84-L84) is wrong, should it be

```
    def merge(self, branch_names):
        """
        Merge `branch_names` into active branch.
        """

        self.git("merge", *branch_names)
```

core/interfaces/branch.py
```diff
@@ -524,43 +534,36 @@ def get_selected_branch_name(self, selection, line):
...
         self.interface = ui.get_interface(self.view.id())
-        selection, line = self.interface.get_selection_line()
-        if not line:
-            return
 
-        segments = line.strip("▸ ").split(" ")
-        branch_name = segments[1]
+        branches = self.get_selected_branches()
+        branches_strings = set()
+        for branch in branches:
```
asfaltboy: This for loop sort of repeats 508 doesn't it? Can we factor these out into a helper method perhaps? I.e a sort of `get_branch_names_with_remote` that can be used in both places? Calling `self.fetch` can be toggled with a boolean argument

core/interfaces/branch.py
```diff
@@ -524,43 +534,36 @@ def get_selected_branch_name(self, selection, line):
...
-                self.fetch(remote=remote_name)
-                self.merge("{}/{}".format(remote_name, branch_name))
+            else:
+                sublime.status_message("Fetching from `{}`...".format(branch[0]))
+                self.fetch(remote=branch[0])
                 sublime.status_message("Fetch and merge complete.")
-                util.view.refresh_gitsavvy(self.view)
-                return
+                branches_strings.add("{}/{}".format(branch[0], branch[1]))
+        self.merge(branches_strings) if not len(branches_strings) == 0 else None
```
asfaltboy: This still won't work with multiple branches until [the fix mentioned by @randy3k is applied][1].

[1]: https://github.com/divmain/GitSavvy/pull/535#issuecomment-257183051

stoivo:
Sorry for taking so long. Does this look better?

asfaltboy:
Looks great, can you also push a fix to [the `BranchesMixin.merge()` method][1]?

[1]:  https://github.com/divmain/GitSavvy/blob/e9bd976a8e20b4bbec29602468557d2dcc89b245/core/git_mixins/branches.py#L79-L84

core/interfaces/branch.py
```diff
@@ -502,18 +502,23 @@ def run(self, edit):
 
     def run_async(self):
         self.interface = ui.get_interface(self.view.id())
-        current_branch_name = self.get_current_branch_name()
 
+        branches = self.get_selected_branches()
+        branches_strings = self.create_branches_strs(branches)
+        self.merge(branches_strings) if not len(branches_strings) == 0 else None
```
asfaltboy: Once [`BranchesMixin.merge()`][1] is fixed (*required*) this should be simplified to `self.merge(branches_strings)` without the ternary non-empty confirmation.

[1]: https://github.com/divmain/GitSavvy/blob/e9bd976a8e20b4bbec29602468557d2dcc89b245/core/git_mixins/branches.py#L79-L84

core/interfaces/branch.py
```diff
@@ -502,18 +502,23 @@ def run(self, edit):
...
+            for name in self.get_selected_branch(sel):
                 if name and not name == current_branch_name:
                     branches.add(name)
 
-        self.merge(branches) if not len(branches) == 0 else None
-        util.view.refresh_gitsavvy(self.view)
+        return branches
 
-    def get_selected_branches(self, sel):
+    def get_selected_branch(self, sel):
```
asfaltboy: Personally, I'd consolidate `get_selected_branch` and `get_selected_branch_name` under `get_selected_branches`. In my opinion since they call each other and not used anywhere else, they can be combined into a single method that does only 1 thing (get all selected branch names, with remote optionally included).

core/interfaces/branch.py
```diff
@@ -524,43 +529,46 @@ def get_selected_branch_name(self, selection, line):
...
 
-        segments = line.strip("▸ ").split(" ")
-        branch_name = segments[1]
+        branches = self.get_selected_branches()
+        for branch in branches:
+            if branch[0] is None:
+                # is remote is not set it is a local branch and can't be fetched
+                continue
+            sublime.status_message("Fetching from `{}`...".format(branch[0]))
+            self.fetch(remote=branch[0])
```
asfaltboy: We're fetching all branches from remote here correct? So this may fetch multiple times from same branch without this being necessary (slowing down the operation). Try instead something like:

```python
remotes_to_fetch = set(filter(None, (b[0] for b in branches)))
for remote in remotes_to_fetch:
    sublime.status_message("Fetching from `{}`...".format(remote))
    self.fetch(remote=remote)
    sublime.status_message("Fetch from {} complete.".format(remote))
```

And the row `sublime.status_message("Fetch and merge complete.")` can be moved after the call to `self.merge`.

core/interfaces/branch.py
stoivo: Would you like to merge these three methods into one?

randy3k: I guess `get_selected_branch` is also useful in some other places. Perhaps it could moved to `BranchInterface`?

asfaltboy:
I kind of like that. And since we're in an interface already, we can have a shortcut property such as `self.interface`. but the again, do we have any other view that allows selecting branches other than branch interface?

core/interfaces/branch.py
stoivo: I like that idea

asfaltboy: I kind of like that. And since we're in an interface already, we can have a shortcut property such as `self.interface`. but the again, do we have any other view that allows selecting branches other than branch interface?

Personally, I just don't like to read a "jumping flow" like that. Which is basically one code unit broken up and sorted reversed. The total row count is not that significant as well.
But, anyway this is totally personal preference. Not worth discussing any more 😄 
(p.s comment moved to proper review context)

asfaltboy:
Anyway, this looks good. I leave the 1-3 method thing up to you.

randy3k:
I can refactor the function later. This PR LGTM.

PR #538, opened 2016-11-01
Fix: bring back log and log graph --all

### LINKED ISSUE #523: all option for "git log" and "git graph"
First off, major kudos for GitSavvy!!! I've now switched from the Git plugin.
Have you considered adding an **All** option to display **git log** and **git graph** for all branches?

randy3k:
Fix #523


divmain:
LGTM!

PR #541, opened 2016-11-01
Enhacement: show "no branch, rebasing" in status when rebasing

randy3k:
screen shot 2016-10-31 at 10 08 55 pm

### DIFF
--- a/core/git_mixins/active_branch.py
+++ b/core/git_mixins/active_branch.py
@@ -95,6 +95,10 @@ class ActiveBranchMixin():
         return status, secondary
 
     def get_branch_status_short(self):
+
+        if self.in_rebase():
+            return "(no branch, rebasing {})".format(self.rebase_branch_name())
+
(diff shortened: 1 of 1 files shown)


core/git_mixins/active_branch.py
```diff
@@ -95,22 +95,26 @@ def get_branch_status(self, delim=None):
         return status, secondary
 
     def get_branch_status_short(self):
-        detached, initial, branch, remote, clean, ahead, behind, gone = \
-            self._get_branch_status_components()
 
-        dirty = "" if clean else "*"
+        if self.in_rebase():
+            return "(no branch, rebasing {})".format(self.rebase_branch_name())
+        else:
```
divmain: Since the `if` block above provides an early return, no need for all the rest of this to reside within an `else`.

randy3k:
Removed the `else` statement.

stoivo:
Look good

PR #542, opened 2016-11-01
Fix: Avoid exception when window is None

asfaltboy:
Some open files (maybe without project), would otherwise occasionally raise the following exception in console:

```python
  File "/Users/pavel/Library/Application Support/Sublime Text 3/Packages/GitSavvy/common/util/view.py", line 72, in refresh_gitsavvy
    view.window().run_command("refresh_folder_list")
AttributeError: 'NoneType' object has no attribute 'run_command'
```

### DIFF
--- a/common/util/view.py
+++ b/common/util/view.py
@@ -69,7 +69,8 @@ def refresh_gitsavvy(view):
         view.run_command("gs_branches_diff_commit_history_refresh")
 
     view.run_command("gs_update_status_bar")
-    view.window().run_command("refresh_folder_list")
+    if view.window():
+        view.window().run_command("refresh_folder_list")
 
 
 def handle_closed_view(view):


stoivo:
Do you know when this happens?

randy3k:
I confirm this, but have no clue why it happens..

asfaltboy:
The [API docs][1] mention:

    Methods		Return Value	Description
    window()	Window			Returns a reference to the window containing the view.

We're calling `refresh_gitsavvy` in quite a few places (bug?). I can only guess that while we were queuing up some async command, which causes "refresh" to be executed, the view was closed/changed. So, while the command may still hold a reference to the view it was executed on, the window does not exist anymore, or rather the window now holds a different view so the reference to the view is severed.

[1]: https://www.sublimetext.com/docs/3/api_reference.html#sublime.View

stoivo:
Make sence

divmain:
Seems reasonable!

Issue #152, opened 2015-03-28
Git blame line walker


yyyc514:
Feel free to rename.  References #102.

When viewing a file in git blame I'd like to hover over a group of lines and then go back in time to each prior version of the file and do a git blame there while preserving my cursor position.  So if the git blame said the prior commit was 'abc' then I hit say '<' (back) and then I'm looking at the blame from abc^ (the parent commit of ABC)... so I can find out who is truly responsible for a change even if there is a lot of churn in a single file.

This would of course also survive renames, etc... since that information is included in git blame.

yyyc514:
A very rough first pass: 
https://github.com/yyyc514/GitSavvy/tree/feature/blame_walker

Features:

- "<" steps back to find the previous change to a given line
- refocuses the view on the old version of the line
- stepping back works even across file renames
- yells at you when you can't go back any further
- yells at you if you haven't picked an exact line

Questions:

- What keystroke should we use here?
- Should there be help to explain keystrokes to people or should they need to read the docs?
- What about stepping forward?

Since blame is a historical view there really isn't any real concept for stepping forward (from a given actual `blame` with no prior context), but we could keep track of your trail as you step back in time and then have another key to step forward.  The you could hit <> to flip between two revisions of the file... the  newer and older.

Thoughts?

divmain:
For early feedback, either is fine since I can just hit "Compare" on the feature branch.  Keeping track of the trail sounds good - that info should be stored in the view settings to persist Sublime restarts.

For views like this, the UX to present contextual help can be complicated.  So I've preferred to take the user to relevant documentation when they do `GitSavvy: help`.  I'm open to other ideas, but user docs are probably the place to start.

`<` and `>` are fine for walking back and forth.

benmosher:
:+1:

stevage:
This sounds awesome - how can I try it?

stoivo:
maybe we can use some of this https://github.com/georgebrock/git-browse

rchl:
Long time without progress. It would be really nice to have this feature. Do you need some help with implementing it?

stoivo:
If you have the time to look at this we would be very thankful.

rchl:
I wonder what original patch author is up to though. Wouldn't want to reinvent the wheel.

divmain:
[...]

Issue #102, opened 2015-03-19
Show current file feature


maxim:
Git package had a very convenient feature: "show current file". It's just like GitSavvy's "log current file", but instead of taking you to full commit diff, it just opens current file at chosen revision. It helps you see how a particular file looked in the past, and manually bring back some code. I don't see an equivalent in GitSavvy, would be awesome to have it. That's pretty much the only thing I'm missing from Git.

divmain:
It shouldn't be difficult to implement as described.  I've also considered a view where you can walk back in time for a particular file.

For example, you open the palette, type something like `git: show file at...`, select a revision, and the view pops up.  This sounds like basically what you've described.  However, at that point, you could then hit maybe <kbd>Super</kbd>+<kbd>Left</kbd>, and the view would change to display the revision immediately before the visible one.

Thoughts @maxim?

maxim:
@divmain This walking sounds super convenient (which is what I love about GitSavvy, you just keep taking things a step or two further). Imho, it's such a clean addition that it doesn't even have to be done in 1st iteration of this feature, can be added later, but yes, definitely would instause it.

maxim:
I would give default key combo a bit more thought however. <kbd>Super</kbd>+<kbd>Left</kbd> and some other similar bindings are often taken, but say my <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>Left</kbd> is available.

divmain:
Hmm, I think <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>Left</kbd> may effect the screen orientation in Windows 7 and later.  How about:

- <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>Left</kbd> - walk back in time (Windows)
- <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>Right</kbd> - walk forward in time (Windows)
- <kbd>Super</kbd>+<kbd>Shift</kbd>+<kbd>Left</kbd> - walk back in time (Linux/OS X)
- <kbd>Super</kbd>+<kbd>Shift</kbd>+<kbd>Right</kbd> - walk forward in time (Linux/OS X)

maxim:
[...]

PR #544, opened 2016-11-01
Enhancement: make show_paginated_panel more flexible

randy3k:
`show_paginated_panel` is now more flexible to handle items.

In many cases, `on_done(index)` is useless for paginated panel. It will be more useful to do `on_done(item)`. It is particularly important when `items` is a generator, as there is no way to pre-store the items,  the function `format_item` can be specified to format generated items on the fly.

### DIFF
--- a/common/quick_panel.py
+++ b/common/quick_panel.py
@@ -1,13 +1,34 @@
+import itertools
 import sublime
 from . import util
 
 
-def show_paginated_panel(items, on_done, limit=6000, **kargs):
+def show_paginated_panel(items, on_done, flags=None, selected_index=None, on_highlight=None,
+                         limit=6000, next_message=None):
     """
-    Display items in quick panel with pagination, and execute on_done(index)
-    when item is selected. `items` can be either a list or a generator.
+    Display items in quick panel with pagination, and execute on_done
+    when item is selected.
+
+    items: can be either a list or a generator.
+    on_done: a callback will take one argument
+    limit: the number of items per page
+    next_message: a message of next page, default is ">>> NEXT PAGE >>>"
+
+    If the elements are tuples of the form `(value1, value2)`,
+    `value1` would be displayed via quick panel and `value2` will be passed to
+    `on_done` when the item is selected. Furthermore, if the quick panel is
+    cancelled, `None` will be passed to `on_done`.
+
     """
-    pp = PaginatedPanel(items, on_done, limit, **kargs)
+
+    pp = PaginatedPanel(
+            items,
+            on_done,
+            flags=flags,
+            selected_index=selected_index,
+            on_highlight=on_highlight,
+            limit=limit,
+            next_message=next_message)
     pp.show()
 
 
@@ -17,7 +38,7 @@ def show_log_panel(entries, on_done, limit=6000, selected_index=None):
     when item is selected. `entries` can be either a list or a generator of LogEnty.
 
     """
-    lp = LogPanel(entries, on_done, limit, selected_index)
+    lp = LogPanel(entries, on_done, limit=limit, selected_index=selected_index)
     lp.show()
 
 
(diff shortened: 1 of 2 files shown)


common/quick_panel.py
```diff
@@ -2,12 +2,23 @@
 from . import util
 
 
-def show_paginated_panel(items, on_done, limit=6000, **kargs):
+def show_paginated_panel(items, on_done, flags=None, selected_index=None, on_highlight=None,
+                         limit=6000, format_item=None, on_done_item=False, next_message=None):
```
divmain: Such a high arity seems suboptimal, and suggests that this function and class are doing too much.  Is there any way to defer some of the responsibility to the consumer, and expose the necessary bits there?

One idea would be to return the `PaginatedPanel` from this function, and allow the consumer to act on it within a callback.  Another would be to accept callback functions that are invoked with other functions (e.g. for panel continuation).

randy3k:
I love the fact `show_paginated_panel` is compatible with`show_quick_panel`, we can just change `show_quick_panel` to `show_paginated_panel` if it is needed.
Therefore returning `PaginatedPanel` is not `my choice. 

To address your concern, I have removed some of the parameters of `show_paginated_panel` . To maintain the same level of flexibility, it understands two types of inputs.

1. compatible with `show_quick_panel`
```python
show_paginated_panel(["foo", "bar"], on_done)
```
`foo` -> `on_done(0)`, `bar` -> `on_done(1)` and `on_done(-1)` if cancalled.

2. allow different return type
```python
show_paginated_panel([("foo", "apple"), ("bar", "orange")], on_done)
```
`foo` -> `on_done("apple")`, `bar` -> `on_done("orange")` and `on_done(None)` if cancalled.

Note: do not confuse with
```
window.show_quick_panel([["foo", "apple"], ["bar", "orange"]], on_done)
```
which displays "apple" and "oranage" as footnote.

common/quick_panel.py
```diff
@@ -81,45 +103,43 @@ def on_selection(self, index):
...
     def __init__(self, items, on_done, limit=6000, selected_index=None):
-        self.limit = limit
-        self.items = (entry for entry in items)
-        self.on_done = on_done
-        self.limit = limit
-        self.flags = sublime.MONOSPACE_FONT | sublime.KEEP_OPEN_ON_FOCUS_LOST
-        self.selected_index = selected_index
-        self.on_highlight = self.on_entry_highlight
+        flags = sublime.MONOSPACE_FONT | sublime.KEEP_OPEN_ON_FOCUS_LOST
+        super().__init__(items, on_done, flags=flags, selected_index=selected_index, limit=limit)
```
asfaltboy: Another way to resolve the mentioned "too many kwargs" code-smell, in case of classes that are to be inherited from, is to move these "options" to class attributes. For example for flags we would do this:

```python
class PaginatedPanel(object):
    flags = None


class LogPanel(PaginatedPanel):
    flags = sublime.MONOSPACE_FONT | sublime.KEEP_OPEN_ON_FOCUS_LOST
```

And then the `__init__` can be removed from `LogPanel`. Afterwards, `PaginatedPanel` can be cleaned up to include parameters that are really required for generally initializing the panel. It's `__init__` method can also accept `**kwargs`, like you have initially, to allow override the options (class attributes) from these as well:

```python
class PaginatedPanel(object):
    flags = None
    # ... more options
    
    def __init__(self, items, on_done, **kwargs):
        self.items = items
        self.on_done = on_done

        for option in ['flags', 'selected_index', 'on_highlight',
        			   'limit', 'next_message', ]:
            if option in kwargs:
                setattr(self, option, kwargs[option])

```

common/quick_panel.py
```diff
@@ -81,45 +96,43 @@ def on_selection(self, index):
...
                     entry.short_hash + " " + entry.summary,
                     entry.author + ", " + util.dates.fuzzy(entry.datetime)
-                ] for entry in batch]
+                ]
 
-    @property
-    def next_message(self):
-        return [">>> NEXT {} COMMITS >>>".format(self.limit),
+    def next_message(self, limit):
+        return [">>> NEXT {} COMMITS >>>".format(limit),
```
randy3k: This is a special class of `PaginatedPanel`, where the `display_list` is always
```
[
    ["abcd111 commit1", "author, data and time"],
    ...,
    ["abcd222 commit2", "author, data and time"],
    [">>> NEXT X COMMITS >>>", "Skip this ..."]
]
```

randy3k:
I have included the changes suggested by @asfaltboy

asfaltboy:
Looks good to me 👍  LGTM

I mean I see more that can be factored, but we gotta stop somewhere!

core/commands/reset.py
```diff
@@ -88,43 +88,35 @@ def run(self):
...
+                if not entry:
+                    continue
+                short_hash, long_hash, summary, reflog_name, reflog_selector, author, datetime = \
+                    entry.split("\n")
+                yield ([reflog_selector + " " + reflog_name,
                         short_hash + " " + summary,
-                        author + ", " + util.dates.fuzzy(datetime)
-                    ]
+                        author + ", " + util.dates.fuzzy(datetime)],
+                       long_hash)
```
asfaltboy: In the future we may want to add a `RefLogEntry` to `git_mixins.history` (similar to `LogEntry`) and pass these around instead of strings. Then `LogPanel` can easily be subclassed in `RefLogPanel` for even cleaner separation.

stoivo:
Can we use this to solve #494?

randy3k:
No, this PR is not related to the graph view at all.

asfaltboy:
Arguably, one could write a generic paginating iterator from that base class, changing the `show()` and removing the `on_highlight` functionality. A paginated graph view could then use that as base.

randy3k:
I would like a general solution to address #102 #152  and #494  simultaneously. Perhaps a class `PaginatedView` which allows step forward and step backward.

stoivo:
Sounds cool and like the approve.

PR #547, opened 2016-11-03
Feature: use arrow key to navigate on dashboards

randy3k:
Arrow key navigation is enabled on dashboards allowing users to use arrow keys to navigate between branches, files, commits etc.

However, it is not enabled in `diff_view`, `inline_diff_view` and `show_commit_view` . These views are showing commits, may be still useful to have arrow keys.

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -381,6 +381,26 @@
         ]
     },
     {
+        "keys": ["down"],
+        "command": "gs_status_navigate_file",
+        "args": { "forward": true },
+        "context": [
+            { "key": "setting.command_mode", "operator": "equal", "operand": false },
+            { "key": "setting.git_savvy.arrow_keys_navigation", "operator": "equal", "operand": true },
+            { "key": "setting.git_savvy.status_view", "operator": "equal", "operand": true }
+        ]
+    },
+    {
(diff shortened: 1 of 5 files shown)


asfaltboy:
I have a feeling there's a reason @divmain left the arrow as default caret movement on most interfaces. Perhaps to allow stuff like copy paste of rows/words anywhere in the interface without requiring mouse ?

stoivo:
Agree @asfaltboy

randy3k:
If the others are not into it, I could make it configurable with the default setting being disabled.

stoivo:
You could make it configurable.  

I still think I would keep the same flow, always `,` and `.` to navigate, everywhere.  Though I don't mind, I will continue to use `,` and `.`.

randy3k:
Sometimes I accidentally used `,` and `.` to navigate quick panel and I find no ways to map `,` and `.` to arrows in quick panel. Therefore I decided to go the other route, i.e., always use arrow keys instead. Of course, `,` and `.` are much handy and faster to type.

randy3k:
It's now configurable with the settings `arrow_keys_navigation`.

randy3k:
Can I get 👍 for merging this?

PR #516, opened 2016-10-23
Fix: fix navigation offset and cursor location when view starts

randy3k:
Now, the cursor will be initialized correctly for all tabs.

PR #549, opened 2016-11-03
Fix: only trigger gs_diff_navigate on initialization

randy3k:
Refreshing the page should not execute `gs_diff_navigate`.

### DIFF
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -69,6 +69,7 @@ class GsDiffCommand(WindowCommand, GitCommand):
         self.window.focus_view(diff_view)
         diff_view.sel().clear()
         diff_view.run_command("gs_diff_refresh")
+        diff_view.run_command("gs_diff_navigate")
 
 
 class GsDiffRefreshCommand(TextCommand, GitCommand):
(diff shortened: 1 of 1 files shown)


asfaltboy:
I would argue current behavior can be safer. If the list of items in an interface changes, then the selection may change to another item, the user might not notice this and perform an unintended action.

Regardless of where the selection goes, I would love if we could flash the _updated_ caret location after a refresh. Is this something that can maybe be done with style switch? Actually, I just enabled the `"highlight_line": true` option in sublime and it's pretty clear where the caret is right now 😊 so this is not such a big issue for me.

randy3k:
- That line was introduced not long ago in my #516 which tries to fix a cursor initialization issue.

- There is an unwanted bug introduced by #516, when we press `w` or `s` to toggle various settings in the view, the cursor moves to the next hunk.

- If the list of items changes, the cursor always goes to the next hunk, it is not useful at all.

- Also, this is a diff view, users can stage or unstage hunks by [GsDiffStageOrResetHunkCommand](https://github.com/divmain/GitSavvy/blob/4ababe3e0559d1332c08a97bcb09ddecf2b57383/core/commands/diff.py#L144-L144), which aways fetch the information from the view, but not an internal list of items. Therefore, it should be safe to not running `gs_diff_navigate` when refreshing


Of the above reasons, I believe that it is correct to only running `gs_diff_navigate` when initialization.

asfaltboy:
OK, thank you for clarifying! I didn't notice this was only for the diff command, you are absolutely right about what is expected here. LGMT!

randy3k:
Great.

PR #550, opened 2016-11-03
Other: implement tests

### LINKED ISSUE #195: Implement and add plugin tests
We can do this using [one](https://github.com/randy3k/UnitTesting) of the [sublime](https://github.com/twolfson/sublime-plugin-tests) [testing](https://bitbucket.org/klorenz/sublimepluginunittestharness) [frameworks](https://github.com/guillermooo/AAAPT) out there (sorted by update date descending).

randy3k:
[Build Status](https://travis-ci.org/randy3k/GitSavvy) [Build status](https://ci.appveyor.com/project/randy3k/gitsavvy) [Coverage Status](https://coveralls.io/github/randy3k/GitSavvy?branch=tests)

close #195

### DIFF
--- a/tests/test_git/common.py
+++ b/tests/test_git/common.py
@@ -1,0 +1,10 @@
+class AssertionsMixin:
+
+    def assert_git_status(self, status):
+        """
+        Assertion of the current git status. `status` should be a list of 4 intergers:
+        The lengths of the staged, unstaged, untracked and conflicted entries.
+        """
+        self.assertEqual(
+            [len(x) for x in self.sort_status_entries(self.get_status())],
+            status)
--- a/tests/test_git/common.py
+++ b/tests/test_git/common.py
@@ -1,3 +1,13 @@
+import sublime
+import subprocess
+import os
+from unittesting.helpers import TempDirectoryTestCase
+
+
+def tidy_path(path):
+    return os.path.realpath(os.path.normcase(path))
+
+
 class AssertionsMixin:
 
     def assert_git_status(self, status):
@@ -8,3 +18,32 @@ class AssertionsMixin:
         self.assertEqual(
             [len(x) for x in self.sort_status_entries(self.get_status())],
             status)
+
+
+class GitRepoTestCase(TempDirectoryTestCase, AssertionsMixin):
+    """
+    TempDirectoryTestCase is a subclass of DeferrableTestCase which creates and opens a temp
+    directory before running the test case and close the window when the test case finishes running.
+    https://github.com/randy3k/UnitTesting/blob/master/unittesting/helpers.py
+    """
+    # add readme file as the initial commit
+    initialize = True
+
+    @classmethod
+    def setUpClass(cls):
+        # since TempDirectoryTestCase.setUpClass is a generator
+        yield from super(GitRepoTestCase, cls).setUpClass()
+        assert tidy_path(cls._temp_dir) == tidy_path(sublime.active_window().folders()[0])
+        assert tidy_path(cls._temp_dir) == tidy_path(cls.window.folders()[0])
+        subprocess.check_call(["git", "init"], cwd=cls._temp_dir)
+        if cls.initialize:
+            cls.add_readme()
+
+    @classmethod
+    def add_readme(cls):
+        readme = os.path.join(cls._temp_dir, "README.md")
+        f = open(readme, "w")
+        f.write("README")
+        f.close()
+        subprocess.check_call(["git", "add", "README.md"], cwd=cls._temp_dir)
+        subprocess.check_call(["git", "commit", "-m", "Add README.md"], cwd=cls._temp_dir)
--- a/tests/test_git/test_git_init.py
+++ b/tests/test_git/test_git_init.py
@@ -1,0 +1,47 @@
+import sublime
+import os
+import subprocess
+from unittesting.helpers import TempDirectoryTestCase
+from .common import AssertionsMixin
+from GitSavvy.core import git_command
+
+
+def tidy_path(path):
+    return os.path.realpath(os.path.normcase(path))
+
+
+class TestInitialization(TempDirectoryTestCase, AssertionsMixin, git_command.GitCommand):
+    """
+    TempDirectoryTestCase is a subclass of DeferrableTestCase which creates and opens a temp
+    directory before running the test case and close the window when the test case finishes running.
+    https://github.com/randy3k/UnitTesting/blob/master/unittesting/helpers.py
+    """
+
+    def test_01_init(self):
+        self.assertEqual(
+            tidy_path(self._temp_dir),
+            tidy_path(sublime.active_window().folders()[0]))
+        subprocess.check_call(["git", "init"], cwd=self._temp_dir)
+        self.assertEqual(
+            tidy_path(self._temp_dir),
+            tidy_path(self.window.folders()[0]))
+
+    def test_02_add_first_file(self):
+        readme = os.path.join(self.repo_path, "README.md")
+        f = open(readme, "w")
+        f.write("README")
+        f.close()
+        self.git("add", "README.md")
+        self.git("commit", "-m", "Init")
+
+    def test_03_is_master(self):
+        branch = self.get_current_branch_name()
+        self.assertEqual(branch, "master")
+
+    def test_04_untrack_file(self):
+        foo = os.path.join(self.repo_path, "foo")
+        with open(foo, "w") as f:
+            f.write("foo")
+        self.assert_git_status([0, 0, 1, 0])
+        self.stage_file(foo)
+        self.assert_git_status([1, 0, 0, 0])
--- a/tests/test_git/test_git_init.py
+++ b/tests/test_git/test_git_init.py
@@ -1,44 +1,18 @@
-import sublime
 import os
-import subprocess
-from unittesting.helpers import TempDirectoryTestCase
-from .common import AssertionsMixin
+from .common import GitRepoTestCase
 from GitSavvy.core import git_command
 
 
-def tidy_path(path):
-    return os.path.realpath(os.path.normcase(path))
+class TestInitialization(GitRepoTestCase, git_command.GitCommand):
 
-
-class TestInitialization(TempDirectoryTestCase, AssertionsMixin, git_command.GitCommand):
-    """
-    TempDirectoryTestCase is a subclass of DeferrableTestCase which creates and opens a temp
-    directory before running the test case and close the window when the test case finishes running.
-    https://github.com/randy3k/UnitTesting/blob/master/unittesting/helpers.py
-    """
-
-    def test_01_init(self):
-        self.assertEqual(
-            tidy_path(self._temp_dir),
-            tidy_path(sublime.active_window().folders()[0]))
-        subprocess.check_call(["git", "init"], cwd=self._temp_dir)
-        self.assertEqual(
-            tidy_path(self._temp_dir),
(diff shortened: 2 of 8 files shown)


randy3k:
The basic testing skeleton is done. 

The tests are written using [DeferrableTestCase](
https://github.com/randy3k/UnitTesting/tree/dc810ee334bb031710b859478faaf50293880995/unittesting/core/st3), it allows the runtime yield to UI until a [condition](https://github.com/randy3k/GitSavvy/blob/2ed494c04017c0cf71df83ad31ebdb127db9fcd4/tests/test_git/helper.py#L44-L44) is met or [pause](https://github.com/randy3k/GitSavvy/blob/2ed494c04017c0cf71df83ad31ebdb127db9fcd4/tests/test_git/test_git_status.py#L28-L28) a certain time after the async calls are begin execuating.

It will be our next tasks to write some tests.

@divmain Please enable travis, appveyor and coveralls for the repo. The above badges are linking to my fork.

stoivo:
What is appveyor.yml?

Looks cool

randy3k:
Appveyor is a CI service for Windows. https://ci.appveyor.com/project/randy3k/gitsavvy

stoivo:
Should we have appveyor and travis?

randy3k:
Travis only supports linux and mac. With appveyor, we provide full coverage for all users. Why not use both?

stoivo:
I think that is cool. I just did not think of it.

randy3k:
Just checked that it works with #556 .

tests/test_git/common.py
```diff
@@ -0,0 +1,60 @@
...
+        app_path = executable_path[:executable_path.rfind(".app/") + 5]
+        executable_path = app_path + "Contents/SharedSupport/bin/subl"
+    subprocess.Popen([executable_path] + list(args))
+    if sublime.platform() == "windows":
+        def fix_focus():
+            window = sublime.active_window()
+            view = window.active_view()
+            window.run_command('focus_neighboring_group')
+            window.focus_view(view)
+        sublime.set_timeout(fix_focus, 300)
```
asfaltboy: This feels like an implementation detail of `DeferrableTestCase`, can it be moved into a helper function in unittesting perhaps?

tests/test_git/common.py
```diff
@@ -0,0 +1,60 @@
...
+        subl(["-n", cls._temp_dir])
+
+        def condition():
+            for d in sublime.active_window().folders():
+                # on Windows, `cls._temp_dir` is lowered cased,
+                # `os.path.normcase` is needed for comparison.
+                if cls._temp_dir == os.path.normcase(d):
+                    return True
+
+        yield condition
```
asfaltboy: Could you kindly explain what is going on here? It looks like we're spawning a new sublime window and then iterating over windows to find the window with the path we gave it as an argument, but I may be wrong.

Why are we yielding a callable in the `setUpClass`? Perhaps this can be separate method with a clear functionality that can be called when setting up the testcase class?

tests/test_git/common.py
```diff
@@ -0,0 +1,60 @@
...
+
+        print(subprocess.check_output(["git", "--version"]).decode("ascii").strip())
+        subprocess.check_call(["git", "init"], cwd=cls._temp_dir)
+
+    @classmethod
+    def tearDownClass(cls):
+        # need at least one window in order to keep sublime running
+        if len(sublime.windows()) > 1:
+            cls.window.run_command("close")
+            yield 1000
```
asfaltboy: Again I find this hard to read. Why are we yielding an integer here but a callable in setUpClass? Does this integer cause some kind of sleep ?

tests/test_git/test_git_init.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def test_01_window_path(self):
+        self.assertEqual(
+            tidy_path(self._temp_dir),
+            tidy_path(sublime.active_window().folders()[0]))
+        self.assertEqual(
+            tidy_path(self._temp_dir),
+            tidy_path(self.window.folders()[0]))
+
+    def test_02_add_file(self):
```
asfaltboy: I assume the numbering means you want test 01 to execute before test 02? This means these [tests are not isolated][1] and is probably not a good idea as dependencies between unit tests usually points at a problem.

[1]: https://pragprog.com/magazines/2012-01/unit-tests-are-first

tests/test_git/test_git_init.py
```diff
@@ -0,0 +1,47 @@
...
+        readme = os.path.join(self.repo_path, "README.md")
+        f = open(readme, "w")
+        f.write("README")
+        f.close()
+        self.git("add", "README.md")
+        yield 100
```
asfaltboy: Here it is again, are we yielding to wait? Please explain why 100, and will it always work with this value?

tests/test_git/test_git_init.py
```diff
@@ -0,0 +1,47 @@
...
+        self.assertEqual(branch, "master")
+
+    def test_04_untrack_file(self):
+        foo = os.path.join(self.repo_path, "foo")
+        with open(foo, "w") as f:
+            f.write("foo")
+
+        self.assertEqual(
+            [len(x) for x in self.sort_status_entries(self.get_status())],
+            [0, 0, 1, 0])
```
asfaltboy: I'm not sure I understand what this list of integers mean. Can we make some kind of CONST table to make this more readable?

tests/test_git/common.py
randy3k: Deferrable test cases are executed by the [DeferringTextTestRunner][1] and it expects not only regular test functions, but also generators. If the test function is a generator, it does the following

- if the yielded object is a callable, the runner will evaluate the callable and check its returned value. If the result is `True`, the runner continues the generator, if not, the runner will wait until the condition is met.

- If the yielded object is an integer, say `x`, then it will [continue][3] the generator after `x` ms.

- Otherwise, the `yield` statement will always wait for 10 ms.
  
[1]: https://github.com/randy3k/UnitTesting/blob/dc810ee334bb031710b859478faaf50293880995/unittesting/core/st3/runner.py#L7
[2]: https://github.com/randy3k/UnitTesting/blob/dc810ee334bb031710b859478faaf50293880995/unittesting/core/st3/runner.py#L49
[3]: https://github.com/randy3k/UnitTesting/blob/dc810ee334bb031710b859478faaf50293880995/unittesting/core/st3/runner.py#L57

randy3k:
I will definitely write some docs about `DeferrableTestCase`.

tests/test_git/test_git_init.py
randy3k: A lot of git operations are sequential, I don't know if it is a better idea to put all the assertions in a single function than what I have now.

tests/test_git/test_git_init.py
randy3k: It definitely needs more explanations. They are the lengths of the [staged, unstaged, untracked and conflicted entries](https://github.com/divmain/GitSavvy/blob/390256711eb4bd0cefeb387f2f0938927608a693/core/git_mixins/status.py#L99
).

tests/test_git/test_git_init.py
asfaltboy: Each operation "should" be tested separately. For operations that need the app to be at a certain state we could use a fixture of sorts, e.g a collection of directories with git repositories at specific states to be copy-pasted on each related set of tests (e.g in `setUpClass`).

This brings me to another point I would like to consider. I would prefer the tests (and their dependencies) to be separate from the plugin, so that downloading a plugin from package-control will not include any of our internal development files. Now, I'm not sure Package Control supports this degree of separation, so we may need to change our build process to [manual release][1].

[1]: https://packagecontrol.io/docs/creating_package_files

stoivo: I also think we should test each operation separately so if I commit something that breaks the commit function I would break every test we have.

asfaltboy:
@randy3k I can't wait to get this merged,  I want to add at least a basic assertion of all principal command functionality, for easier detection of misses like I just had on #558

randy3k:
I guess it is ready to be merged. But the tests may need to be rewritten. The examples which I provided are just for illustration purpose. We also need @divmain to enable the CI services.

stoivo:
I created an issue at  Package Control Channel repo https://github.com/wbond/package_control_channel/issues/5942.

I somehow think that we can go for a second option. Create a banche called release where we squash all commit from last release and remove test folder. Then tell Package Control to look at that branch. Still keep all development on  master.

randy3k:
I guess we just need to tag from the release branch, it is not necessary to tell PC to look at that the release branch.

stoivo:
oh yes thats correct 👍

divmain:
@asfaltboy, when you get a chance, can you mark your review as "Approved" (if you think its ready)?  I'm taking a look also.

docs/README.md
```diff
@@ -159,3 +159,8 @@ For more information see [custom commands documentation](custom.md)
 - [flow: feature/release/hotfix publish](flow.md#flow-featurereleasehotfix-publish)
 - [flow: feature/release track](flow.md#flow-featurerelease-track)
 - [flow: feature pull](flow.md#flow-feature-pull)
+
+
+## Package Testing
+
+Check the implementation details of the [tests](testing.md).
```
divmain: This should be moved into `/CONTRIBUTE.md`, since this is not pertinent for GitSavvy users.

tests/test_git/common.py
```diff
@@ -0,0 +1,73 @@
...
+import subprocess
+import os
+import shutil
+import tempfile
+from unittesting import DeferrableTestCase
+
+
+def subl(*args):
+    """
+    Ideally, it would be moved into UnitTesting as a helper function.
```
divmain: I'm not sure what this doc string is communicating.

tests/test_git/common.py
divmain: Seconded.

tests/test_git/test_git_init.py
divmain: This `git` command as run here should be executed synchronously.  That being the case, I'm surprised that its necessary to wait after the `add` occurs.

randy3k: you are right.

divmain: Btw, I'm loving this pattern of using generators to control the flow of tests!

tests/test_git/common.py
randy3k: moved.

docs/README.md
randy3k: moved.

randy3k:
As for the style of testing,  
- the unit tests for `core/git_mixins` like `git status` and `git add` should be easily done
- functional tests for `core/commands` and `core/interface` are a bit complicated. Especially when the quick panel is involved. It is unclear to me how we could interact with the quick panel. (Can the quick panel be controlled programmatically at all?)

randy3k:
If there are no objections, I would like to push this to `GitSavvy/dev` or `GitSavvy/test` to see how it will go.

divmain:
@randy3k sorry to leave this idle for so long.  I have a strong preference for CircleCI but, honestly, you've done all the work here.  So I consider this ready to merge.  I'll setup Travis as soon as that happens, if you think its ready.

randy3k:
Currently, circleci only has limited support over multiple os. I thing it would be easier to stick with Travis. This should be ready to be merged. You may need to also enable appveyor for the windows platform and coverage.io for the test coverage.

divmain:
Sounds great.  Thanks again @randy3k!

PR #553, opened 2016-11-04
Fix: display the "git binary not found" error message only once

### LINKED ISSUE #551: GitSavvy makes Sublime unusable when git is not installed
I installed sublime on a new machine. I copied AppData/Roaming/Sublime Text 3. When I try to do anything, GitSavvy absolutely blasts me with messages that git can't be found. It has made sublime utterly unusable. I can't even disable the package, because trying to open the command window triggers another message.

FOR THE LOVE OF GOD, ADD AN OPTION FOR "SHUT THE F*** UP! I KNOW I DON'T HAVE GIT INSTALLED YET!"

asfaltboy:
This avoids blocking the user from using sublime by spamming her/him with endless messages about git executable not found. It displays the alert dialog once and continues to log further failures as exceptions.

This allows GitSavvy to start functioning once git is on path (or the settings are updated to point to it's explicit location).

Fixes #551

### DIFF
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -29,7 +29,7 @@ from .exceptions import GitSavvyError
 
 
 git_path = None
-
+error_message_displayed = False
 
 UTF8_PARSE_ERROR_MSG = (
     "GitSavvy was unable to parse Git output as UTF-8. Would "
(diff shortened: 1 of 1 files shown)


randy3k:
LGTM

PR #554, opened 2016-11-04
Fix: do not merge current branch

randy3k:
`get_selected_branch` returns a tuple `(remote, branch_name)`, we need to compare the second element.

### DIFF
--- a/core/interfaces/branch.py
+++ b/core/interfaces/branch.py
@@ -513,7 +513,7 @@ class GsBranchesMergeSelectedCommand(TextCommand, GitCommand):
         branches = set()
         for sel in self.view.sel():
             for name in self.get_selected_branch(sel):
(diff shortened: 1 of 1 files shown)


asfaltboy:
LGTM

core/interfaces/branch.py
```diff
@@ -513,7 +513,7 @@ def get_selected_branches(self):
         branches = set()
         for sel in self.view.sel():
             for name in self.get_selected_branch(sel):
-                if name and not name == current_branch_name:
+                if name and not name[1] == current_branch_name:
```
stoivo: Should we check that name[0] is None since I might want to merge remotea/master in to master

asfaltboy:
Oh! @stoivo good catch! Can you submit a quick PR for this?

stoivo:
I will

PR #555, opened 2016-11-04
Move method from GsBranchesMergeSelectedCommand to Branch interface

stoivo:
#554

### DIFF
--- a/core/interfaces/branch.py
+++ b/core/interfaces/branch.py
@@ -159,6 +159,41 @@ class BranchInterface(ui.Interface, GitCommand):
 
         return output_tmpl, render_fns
 
+    def get_selected_branches(self):
+        current_branch_name = self.get_current_branch_name()
+        branches = set()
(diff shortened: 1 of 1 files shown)


core/interfaces/branch.py
```diff
@@ -159,6 +159,41 @@ def render(remote_name=remote_name, branches=branches):
...
+        branches = set()
+        for sel in self.view.sel():
+            for line in util.view.get_lines_from_regions(self.view, [sel]):
+                branch = self.get_selected_branch_name(sel, line)
+                if branch and not (branch[0] is None and branch[1] == current_branch_name):
+                    branches.add(branch)
+
+        return branches
+
+    def get_selected_branch_name(self, selection, line):
```
randy3k: it may be better to use a underscore prefix `get_selected_branch_name` -> `_get_selected_branch_name` to indicate that it is an internal function.

PR #558, opened 2016-11-06
Feature: refactor quick panel for (mostly) static options

asfaltboy:
This  PR adds two reusable mixins that replace much repetitive and sometimes inconsistent code. 
Hopefully this way the commands has less cruft and easier to read.

In  this refactor I only touched panels with "static options". The next step would be to replace the remaining panels that prompt for branch/remote. By default most of the updated commands run in a synchronous manner (in the current thread), since the panel options are available immediately.

### DIFF
--- a/core/commands/log_graph.py
+++ b/core/commands/log_graph.py
@@ -2,7 +2,7 @@ import sublime
 from sublime_plugin import WindowCommand, TextCommand
 import re
 from ..git_command import GitCommand
-from .log import GsLogActionCommand
+from .log import GsLogActionCommand, GsLogCommand
 from .navigate import GsNavigate
 from ...common import util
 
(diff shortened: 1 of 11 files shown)


stoivo:
This looks really great.

randy3k:
Look great. Thanks.

randy3k:
It seems that `git: log current file` and `git: graph current file` are broken.

asfaltboy:
@randy3k thanks! Silly mistake with loop local variables. I fixuped the specific GsLogCommand commit, hence had to force push all the last 5 ones.

core/commands/log_graph.py
```diff
@@ -152,81 +152,59 @@ def get_graph_args(self):
...
+            self.actions.extend([
                 ["diff_commit", "Diff commit"],
-                ["diff_commit_cache", "Diff commit (cached)"]
-            ]
+                ["diff_commit_cache", "Diff commit (cached)"],
+            ])
 
-        self._file_path = self.file_path
         if self._file_path:
             self.actions.insert(1, ["show_file_at_commit", "Show file at commit"])
```
randy3k: For some reason, it is still not functioning. This option does no show up in `git: graph current file`

randy3k: I figured out the reason. We need to evaluate `self.file_path` before the `graph_view` is created.

asfaltboy: Sort of, I just forgot to pass the evaluated `self.file_path` property to the parent class, so it was using default `None` .. should be fixed now

randy3k:
Great, it fixes it.

PR #569, opened 2016-11-23
Fix: various fixes for the push commands

### LINKED ISSUE #566: STATUS dashboard: "Push current branch"
I think the action might not be what it's meant to be. I found it surprising when it pushed more than just my "current" branch, which I took to mean the very branch I was on, not all the other ones that are configured to track a remote branch with the same name.

Example: I'm on `master`, tracking remote `master`, and I also have a local branch `develop` tracking remote `develop`. When I'm on the status dashboard, hitting P (push current branch) also tries to push `develop`, which may or may not be the intention (it certainly wasn't mine).

Having viewed the logs via "GitSavvy: enable logging", then "GitSavvy: view recorded log", the command is simply `git push`. I think it should be `git push <remote> master`.

`git push` is definitely useful, but there is already a corresponding action on the branches dashboard: `[P] push all branches to remote`, which has a much clearer message.

randy3k:
- `GsPushCommand` only push current branch, fix #566 
- change the keybind from [P] to [p] in Status dashboard so that it is consistent with the branch dashboard.
- extend `GsPushToBranchNameCommand` default arguments to accept a prespecified branch name.
- push current branch to a remote branch with the same name if `prompt_for_tracking_branch` is false
- fix `GsPushToBranchCommand` when the name of the remote branch is not the same as the local

### DIFF
--- a/core/commands/push.py
+++ b/core/commands/push.py
@@ -140,9 +140,10 @@ class GsPushToBranchCommand(WindowCommand, PushBase):
         # If the user pressed `esc` or otherwise cancelled.
         if branch_index == -1:
             return
-
+        current_local_branch = self.get_current_branch_name()
         selected_branch = self.branches_on_selected_remote[branch_index].split("/", 1)[1]
-        sublime.set_timeout_async(lambda: self.do_push(self.selected_remote, selected_branch))
+        sublime.set_timeout_async(
+            lambda: self.do_push(self.selected_remote, selected_branch, local_branch=current_local_branch))
 
 
 class GsPushToBranchNameCommand(WindowCommand, PushBase):
--- a/core/commands/push.py
+++ b/core/commands/push.py
@@ -42,7 +42,7 @@ class PushBase(GitCommand):
 class GsPushCommand(WindowCommand, PushBase):
 
     """
-    Perform a normal `git push`.
+    Push current branch.
     """
 
     def run(self, force=False):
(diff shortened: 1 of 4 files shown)


stoivo:
LGTM

docs/remotes.md
```diff
@@ -18,7 +18,7 @@ When running this command, you will be prompted first for the remote you want to
 
 ## `git: push`
 
-This command performs a simple `git push`.  In recent Git versions, this will result in a push to the tracking branch if it exists, else a remote branch with the same name as the local.
+This command will push current branch to the tracking branch. If the tracking branch is not set, you will be asked to configure it.
```
asfaltboy: Is there any case where a user would like to use push the branch to <remote>/<branch> without setting a tracking branch? Perhaps if a mistake is made when configuring tracking? In my opinion it's a really good default; and we can point to the `[t] configure tracking` action in the BRANCHES dashboard.

core/commands/push.py
```diff
@@ -127,9 +131,10 @@ def on_select_branch(self, branch_index):
         # If the user pressed `esc` or otherwise cancelled.
         if branch_index == -1:
             return
-
+        current_local_branch = self.get_current_branch_name()
         selected_branch = self.branches_on_selected_remote[branch_index].split("/", 1)[1]
-        sublime.set_timeout_async(lambda: self.do_push(self.selected_remote, selected_branch))
+        sublime.set_timeout_async(
+            lambda: self.do_push(self.selected_remote, selected_branch, local_branch=current_local_branch))
```
asfaltboy: I'm not sure about the reason for this change? Can you give a little bit more context? Does this attempt to resolve #570 ?

docs/remotes.md
randy3k: The docs is updated to reflect the recent behavior. The original docs is oudated and the command requires a remote tracking branch even without this PR.

core/commands/push.py
randy3k: No. It does not resolve #570. But it fixes a similar issue mentioned in #570, i.e, when the local branch and the remote tracking branch are of different name.

docs/remotes.md
randy3k: I have updated the code, if `prompt_for_tracking_branch` is set to `false`, a user will be prompted for a remote, and then it pushes the current branch to a remote branch with the same name.

randy3k:
Hi guys, I have updated the code substantially, please review again. The description of this PR should explain the modifications.

stoivo:
I think it looks good @randy3k, maybe @asfaltboy  will see something else

core/commands/push.py
```diff
@@ -53,8 +53,21 @@ def run(self, force=False):
...
+            upstream = self.get_upstream_for_active_branch()
+            current_local_branch = self.get_current_branch_name()
+            remote, branch = upstream.split("/", 1)
+            sublime.set_timeout_async(
+                lambda: self.do_push(remote, branch, local_branch=current_local_branch, force=force))
         else:
-            sublime.set_timeout_async(lambda: self.do_push(None, None, force=force))
+            # if `prompt_for_tracking_branch` is false, ask for a remote and perform
+            # push current branch to a remote branch with the same name
+            current_local_branch = self.get_current_branch_name()
```
divmain: I know this has already been merged, but ideally this sort of operation would occur on the non-UI thread (`run_async` pattern).

randy3k: I will fix it in #571.

PR #573, opened 2016-11-24
Other: refactor GsBranches commands to use interface.get_selected_branches()

randy3k:
A number of functions in `core/interfaces/branch.py` share the same common code to reveal the selected branch. This PR refactors them so that they all use the common function `get_selected_branch`.


randy3k:
I will test this for one or two days before merging this.

PR #575, opened 2016-11-24
Fix: handle vintageous and arrow keys in commit compare view

randy3k:
As title.


randy3k:
I agree. It is a hot fix for the current issue. We need @divmain's response to #550 to starting working on the tests.

PR #577, opened 2016-11-25
Feature: show current file history

randy3k:
Showing current file history, it is essentially a shortcut of `git: log current file` -> `For current branch` -> `show file at commit`.

### DIFF
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -225,6 +225,10 @@
         "args": { "current_file": true }
     },
     {
+        "caption": "git: show current file at commit",
+        "command": "gs_show_current_file"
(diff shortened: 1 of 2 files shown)


stoivo:
LGTM!

Default.sublime-commands
```diff
@@ -225,6 +225,10 @@
         "args": { "current_file": true }
     },
     {
+        "caption": "git: show current file",
```
asfaltboy: Perhaps the command can be more obvious: `git: show current file at commit` ?

randy3k:
Renamed the command as `git: show file at commit`.

PR #580, opened 2016-12-07
Feature: add command to show log of selected branch

asfaltboy:
Includes a shortcut in the branches dashboard.

I find this valuable as it saves me 1-2 addl. keypresses.

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -1031,6 +1031,14 @@
             { "key": "setting.git_savvy.branch_view", "operator": "equal", "operand": true }
         ]
     },
+    {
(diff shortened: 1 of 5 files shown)


randy3k:
I would rather modify `GsLogByBranchCommand` and call it with the selected branch than to call `GsLogBase` directly.

randy3k:
We should actually rename `GsLogBase` as `LogMixins` and remove the `WindowCommand` and `GitCommand` inheritance. Those should go to the actual commands `GsLogCurrentBranchCommand`, `GsLogAllBranchesCommand` etc.

randy3k:
it looks good to me.

PR #585, opened 2016-12-09
Other: refactor GsResetBase, RefLogEntry and RefLogMixin to handle ref logs

randy3k:
- refactor GsResetBase to ResetMixin
- added RefLogMixin (mirror of LogMixin)
- added RefLogEntry (mirror of LogEntry)

after it is merged, we can provide command `git: reflog` similar to `git: log`.

### DIFF
--- a/core/commands/reflog.py
+++ b/core/commands/reflog.py
@@ -1,0 +1,37 @@
+import sublime
+from ..ui_mixins.quick_panel import show_paginated_panel
+from ...common import util
+
+
+class RefLogMixin(object):
+
+    _limit = 6000
+
+    def run(self):
+        sublime.set_timeout_async(self.run_async)
+
+    def run_async(self):
+        show_paginated_panel(self.log_generator(), self.do_action, limit=self._limit)
+
+    def log_generator(self):
+        skip = 0
+        while True:
+            logs = self.reflog(limit=self._limit, skip=skip)
+            if not logs:
+                break
+            for l in logs:
+                yield (["{} {}".format(l.reflog_selector, l.reflog_name),
+                        "{} {}".format(l.short_hash, l.summary),
+                        "{}, {}".format(l.author, util.dates.fuzzy(l.datetime))],
+                       l.long_hash)
+            skip = skip + self._limit
+
+    def do_action(self, commit_hash):
+        if hasattr(self, 'window'):
+            window = self.window
+        else:
+            window = self.view.window()
+        window.run_command("gs_log_action", {
+            "commit_hash": commit_hash,
+            "file_path": self._file_path
+        })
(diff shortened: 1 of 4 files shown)


randy3k:
Let's wait for a day or two.

asfaltboy:
I'll check it out too. By the way, how do you feel about freezing the current release content, i.e feature PRs should merge into `develop` which will mark the "next next" version, while fixes can continue to merge into `master`. This way the content of the next release (that will come from `master`) will be "frozen"?

core/commands/reflog.py
```diff
@@ -0,0 +1,37 @@
...
+        show_paginated_panel(self.log_generator(), self.do_action, limit=self._limit)
+
+    def log_generator(self):
+        skip = 0
+        while True:
+            logs = self.reflog(limit=self._limit, skip=skip)
+            if not logs:
+                break
+            for l in logs:
+                yield ([l.reflog_selector + " " + l.reflog_name,
```
asfaltboy: I would prefer we used string formatting here; [string concat is inefficient][1] and, arguably, harder to read.

[1]: https://wiki.python.org/moin/PythonSpeed/PerformanceTips#String_Concatenation

stoivo:
@asfaltboy, I asume you want to do that so we can shit bug fixes without shipping refactures that we have done?

Do you have a plan who we can solve this and the problem with shiping tests?

asfaltboy:
Yes, basically the more features/refactoring we add, the more risk we have of breaking things, especially without tests. I tend to limit the number of changes in shipped release, to keep it manageable. And IMHO we reached critical change count in current unreleased branch `master`

asfaltboy:
What I suggest is from next PR onwards we merge into a separate "features" or "next" branch rather than "master"

randy3k:
Good idea. How about following git flow?

asfaltboy:
Yes, basically as described in [jeff 's blog post][1], we add a develop branch, against which we'll do PRs most of the time. Here are 4 steps that are the essence of git-flow:

1. _Master_ branch will always be at the latest released version commit, and will only receive merged after a release, including a hotfix release.
2. At a pre-defined scheduled interval, let's say every month, we will create a new _release_ branch (with semantic versioning); this branch will end up being released at another predefined interval, say 2 weeks.
3. During those 2 weeks, the release branch will not be targeted for feature merges, but only for bug fixes, while feature merges will target the next release by merging into _develop_. This is also known as "feature freeze". Every commit in the release branch should also be cherry-picked into develop.
4. When the scheduled release date arrives, we finish the release by tagging and merging it back into _master_.

Once we agree on the flow and schedule we can use e.g a recurring calendar event, to remind us to create a release branch, finish release, etc...

I am going to email @divmain with this and other points awaiting his attention.

gitflow


[1]: http://jeffkreeftmeijer.com/2010/why-arent-you-using-git-flow/

divmain:
I'm responding to your thoughts on Git Flow via email :)

Merging this.

PR #586, opened 2016-12-09
Fix: cherry-pick branch variable has changed

randy3k:
The bug was introduced in #580 as the variable `_selected_branch` was renamed as `_branch`.


asfaltboy:
You are absolutely right, refactoring is risky, thanks!

randy3k:
No problem

PR #597, opened 2017-02-10
Stash unstaged files only

mreq:
The `GsStatusCreateStashCommand` and `GsStatusCreateStashWithUntrackedCommand` commands are (according to the comments) meant to create stash from unstaged changes - for that, the `-k` flag is needed.

Stashing only the unstaged changes does indeed make more sense.

### DIFF
--- a/core/git_mixins/stash.py
+++ b/core/git_mixins/stash.py
@@ -35,7 +35,7 @@ class StashMixin():
         """
         Create stash with provided description from working files.
         """
-        self.git("stash", "save", "-u" if include_untracked else None, description)
+        self.git("stash", "save", "-k", "-u" if include_untracked else None, description)
 
     def drop_stash(self, id):
         """


stoivo:
I'm not sure I have a strong opinion here. I think we should check we should check with  @divmain, @asfaltboy @randy3k 

I would use the full name of the option, not the one character flag.

```
       save [-p|--patch] [-k|--[no-]keep-index] [-u|--include-untracked] [-a|--all] [-q|--quiet] [<message>]
           Save your local modifications to a new stash, and run git reset --hard to revert them. The
           <message> part is optional and gives the description along with the stashed state. For quickly
           making a snapshot, you can omit both "save" and <message>, but giving only <message> does not
           trigger this action to prevent a misspelled subcommand from making an unwanted stash.

           If the --keep-index option is used, all changes already added to the index are left intact.

           If the --include-untracked option is used, all untracked files are also stashed and then cleaned
           up with git clean, leaving the working directory in a very clean state. If the --all option is
           used instead then the ignored files are stashed and cleaned in addition to the untracked files.

           With --patch, you can interactively select hunks from the diff between HEAD and the working tree
           to be stashed. The stash entry is constructed such that its index state is the same as the index
           state of your repository, and its worktree contains only the changes you selected interactively.
           The selected changes are then rolled back from your worktree. See the "Interactive Mode" section
           of git-add(1) to learn how to operate the --patch mode.

           The --patch option implies --keep-index. You can use --no-keep-index to override this.
```

from `$ man git-stash`

stoivo:
@mreq, Thanks for you PR 🎉

Issue #594, opened 2017-01-22
Endless loop: "GitSavvy was unable to parse Git output as UTF-8"


deathaxe:
GitSavvy still suffers from the endless loop, if it fails to parse git output as UTF-8. This issue is already descriped in #300. By pressing the message dialog's 'cancel' button a panel with the following message is opened, but the dialog box reapears as soon as the open DIFF view is activated again. There is no chance to close it at all. I need to kill Sublime Text and delete the recently open *.sublime-workspace file. Otherwise the DIFF view is restored and the endless loop starts again on next execution of Sublime Text.

panel text
```
'utf-8' codec can't decode byte 0xfc in position 139: invalid start byte
```

You must not expect utf-8 as encoding of any output as it doesn't need to be true for `git show` command especially if git is configured to store files in a certain encoding other then utf-8.

I can't say who is responsible but I use git for Windows 2.11.0 and GitExtensions 2.49 with default settings to do most of my work with versioning. One of them sets git's file encoding to operating system default value, which is Windows-1252 in my case.

We could discuss whether it is a good idea to store files with an enconding different from utf-8 or not but infact git allows it, so everyone needs to support it.

I am just starting to use git tools like GitSavvy directly from Sublime Text to hopefully replace external tools some day. 

## Steps to reproduce

1. Configure git to store files as Windows-1252 encoding as it is done automatically with Git for Windows on. (`i18n.filesencoding=Windows-1252`)
2. Create a repository and commit a file with none ascii characters like (äöß)
3. Modify the file in worktree
4. Call git: status
5. Select a modified file
6. press `l`  key to run inline diff
7. You are in endless loop

## System Information

- Sublime Text 3126
- GitSavvy 2.13.0 or commit e6c00ef4da1d49ea86a9f323f3541bd201838ed4
- Windows 10 x64 de-DE

stoivo:
Thank you very much for a good reproduction steps. I have had the same issue myself but never found the case of it. I will test the steps later today.

stoivo:
I have been testing yesterday and today. I have tested on Mac OSX and on Windows 10 with default installation of git from https://git-scm.com/. 

I have been setting that setting with  `git config i18n.filesencoding Windows-1252` on the repo I have tested with. It still work like it should. Is possible if you can create a repo so I can pull it down and reproduce? I think that would help us a lot.

deathaxe:
Created a repository with Windows-1252 codepage set

[...]

PR #603, opened 2017-02-21
First try to read with utf-8 then latin-1

stoivo:
Related to #594

### DIFF
--- a/core/commands/inline_diff.py
+++ b/core/commands/inline_diff.py
@@ -225,8 +225,12 @@ class GsInlineDiffRefreshCommand(TextCommand, GitCommand):
         as a string.
         """
         file_path = os.path.join(self.repo_path, file_path)
-        with open(file_path, "rt", encoding="utf-8") as f:
-            return f.read()
+        try:
+            with open(file_path, "rt", encoding="utf-8") as f:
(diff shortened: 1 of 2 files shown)


divmain:
This seems pretty reasonable.  Going to merge, and we can ask for feedback from those who reported the issue.

bizoo:
Argh, why latin-1 ?
Please look at the last comment of #594, the encoding must be the current encoding of the view, not an arbitrary one.

stoivo:
I would use the view encoding if it was an option, but in this case it is not a open file in a buffer. It is stdout from a command. I can't see how I can use this to find the encoding for the file. 
@bizoo, do you know how I can get it correct?

stoivo:
@bizoo, https://github.com/divmain/GitSavvy/pull/603/files#diff-1d123c94cf2d78df2c8edace6fa6ee97R228 
Here I would open the file in a buffer, let sublime determine the encoding, then read the buffer and close it. I don't see what we are gaining on that. 

I want to say thanks you for commenting, @bizoo. I don't like working with encoding problems, so I am looking the the quickest solution. I very open for solution which make sure I never need to look back at this part of the code.

PR #605, opened 2017-02-24
Add: Docs for `Signing your commits with GPG`

stoivo:
Related to #36


divmain:
This is great.  Thanks @stoivo!

PR #606, opened 2017-02-25
Render inline diff with correct encoding

stoivo:
So the diff renders correctly now. While there is still a problem when we want to apply the diff. I think I know how to solve it. I just want to know If this works.

### DIFF
--- a/core/commands/inline_diff.py
+++ b/core/commands/inline_diff.py
@@ -151,7 +151,7 @@ class GsInlineDiffRefreshCommand(TextCommand, GitCommand):
         else:
             indexed_object = self.get_indexed_file_object(file_path)
             indexed_object_contents = self.get_object_contents(indexed_object)
-            working_tree_file_contents = self.get_file_contents(file_path)
+            working_tree_file_contents = self.get_file_contents_binary(file_path)
             working_tree_file_object = self.get_object_from_string(working_tree_file_contents)
 
             # Display the changes introduced between index and working dir.
@@ -219,18 +219,25 @@ class GsInlineDiffRefreshCommand(TextCommand, GitCommand):
         """
         return self.git("show", "--no-color", object_hash)
 
+    def get_file_contents_binary(self, file_path):
+        """
+        Given an absolute file path, return the binary contents of that file
+        as a string.
+        """
+        file_path = os.path.join(self.repo_path, file_path)
+        with open(file_path, "rb") as f:
+            return f.read()
+
     def get_file_contents(self, file_path):
         """
         Given an absolute file path, return the text contents of that file
         as a string.
         """
-        file_path = os.path.join(self.repo_path, file_path)
+        binary = self.get_file_contents_binary(file_path)
         try:
-            with open(file_path, "rt", encoding="utf-8") as f:
-                return f.read()
+            return binary.decode('utf-8')
         except UnicodeDecodeError as unicode_err:
-            with open(file_path, "rt", encoding="latin-1") as f:
-                return f.read()
+            return binary.decode('latin-1')
 
     def get_object_from_string(self, string):
         """
@@ -238,7 +245,7 @@ class GsInlineDiffRefreshCommand(TextCommand, GitCommand):
         stored in the current repo, and return an object-hash that can be
         used to diff against.
         """
-        stdout = self.git("hash-object", "-w", "--stdin", stdin=string)
+        stdout = self.git("hash-object", "-w", "--stdin", stdin=string, encode=False)
         return stdout.split("\n")[0]
 
     def get_inline_diff_contents(self, original_contents, diff):
--- a/core/commands/inline_diff.py
+++ b/core/commands/inline_diff.py
@@ -226,7 +226,10 @@ class GsInlineDiffRefreshCommand(TextCommand, GitCommand):
         """
         file_path = os.path.join(self.repo_path, file_path)
         with open(file_path, "rb") as f:
-            return f.read()
+            binary = f.read()
+            binary = binary.replace(b"\r\n", b"\n")
+            binary = binary.replace(b"\r", b"")
+            return binary
 
     def get_file_contents(self, file_path):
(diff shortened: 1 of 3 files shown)


core/commands/inline_diff.py
```diff
@@ -219,26 +219,33 @@ def get_object_contents(self, object_hash):
...
-        file_path = os.path.join(self.repo_path, file_path)
+        binary = self.get_file_contents_binary(file_path)
         try:
-            with open(file_path, "rt", encoding="utf-8") as f:
-                return f.read()
+            return binary.decode('utf-8')
         except UnicodeDecodeError as unicode_err:
-            with open(file_path, "rt", encoding="latin-1") as f:
-                return f.read()
+            return binary.decode('latin-1')
```
deathaxe: I am wondering if calling `binary.decode()` without argument would enable automatic encoding detection? Why not let the `decode` method decide what encoder to use?

asfaltboy: Not so easy unfortunately, [the `decode()` method of bytes object defaults to "utf-8"][1]

[1]: https://docs.python.org/3/library/stdtypes.html#bytes.decode

core/git_command.py
```diff
@@ -124,7 +125,7 @@ def raise_error(msg):
                                  cwd=working_dir or self.repo_path,
                                  env=environ,
                                  startupinfo=startupinfo)
-            stdout, stderr = p.communicate(stdin.encode(encoding="UTF-8") if stdin else None)
+            stdout, stderr = p.communicate((stdin.encode(encoding="UTF-8") if encode else stdin) if stdin else None)
```
deathaxe: Pushing binary file content to git works fine for me here.

One question: Why is utf-8 enforced in text mode? This is the default encoding of git, but as you push file content here, it might be more reliable to use the value from git config `i18n.filesencoding` if available to decide how to encode stdin and fall back to utf-8 if setting is not present?

core/commands/inline_diff.py
deathaxe: Is it the same with `codecs.decode()`? I ask because this is the last fallback in GitGutter if any other decoding failed. Wouldn't make sense if it defaults to utf-8, too.

deathaxe:
Just to state it: In general, the current solution works for me. The inline diff now shows correct text in all situations.

core/commands/inline_diff.py
asfaltboy: It seems [like it is][1], I have not seen GitGutter's code so I can't comment on what it does

[1]: https://docs.python.org/3/library/codecs.html#codec-base-classes

core/git_command.py
stoivo: I don't think we can rely on `i18n.filesencoding` since  one project can have multiple encodings without a problem in git. I think we have couple of this project ourself.
I will implement it to use the same encoding to stage a hunk as showing the diff.

stoivo:
Yay, good to hear.

core/commands/inline_diff.py
```diff
@@ -219,26 +219,33 @@ def get_object_contents(self, object_hash):
...
         return self.git("show", "--no-color", object_hash)
 
+    def get_file_contents_binary(self, file_path):
+        """
+        Given an absolute file path, return the binary contents of that file
+        as a string.
+        """
+        file_path = os.path.join(self.repo_path, file_path)
+        with open(file_path, "rb") as f:
+            return f.read()
```
deathaxe: Inline diff shows the whole file as changed because of different line feeds. On Windows the `core.autocrlf` setting is set `true` which causes files in the working tree to have Windows style `\r\n` line endings while the file in the index still has `\n`. By passing the binary content to `hash-object` the translation is no longer done and the file is "commited" as is. I changed the following line just to see if this is the case ... and it is. With those lines, everything is fine for me. Not sure, if there any edge cases to care about, when replacing.

```python
            binary = f.read()
            binary = binary.replace(b"\r\n", b"\n")
            binary = binary.replace(b"\r", b"")
            return binary
```

core/commands/inline_diff.py
```diff
@@ -219,26 +219,33 @@ def get_object_contents(self, object_hash):
...
+            return binary.decode('latin-1')
 
     def get_object_from_string(self, string):
         """
         Given a string, pipe the contents of that string to git and have it
         stored in the current repo, and return an object-hash that can be
         used to diff against.
         """
-        stdout = self.git("hash-object", "-w", "--stdin", stdin=string)
+        stdout = self.git("hash-object", "-w", "--stdin", stdin=string, encode=False)
```
deathaxe: A question about `hash-object` in common: Each call creates a new blob file in git's object database. This might be fine, if the file is commited anyway. But how do you ensure no longer files are deleted? From my current perspective you create dangling objects this way and polute the database. Even a simple `git gc` does not remove them.

stoivo: I don't know, I know that reflow is saved for 90 days by default so I would assume the same here ?

stoivo:
Yo,  give it a try now. This might be the final work on this encoding issues for a while 🎉 :)

core/commands/inline_diff.py
deathaxe: Hope so. But why do you create such a commit at all? If you want to diff the working file directly from disk, you can call git diff with the relative local file path and commit hash to compare against.

Creating a fake commit makes only sense, if you want to compare the view content (which may differ from saved file on disk) againt the index.

deathaxe:
Works like a charm.

core/commands/inline_diff.py
divmain: These should be cleaned up by `gc` whenever their creation date falls before the `prune` date.  You can force that by passing an explicit date or `all`.

As to why it is here, this is used in a couple of contexts for content that is not currently committed to the repo anywhere.  I can't recall why it was needed here, but even in large repos, I have not noticed any performance impact from the orphaned git blobs.

Issue #610, opened 2017-03-01
Feature Request: Add a quick way to "git show HEAD"


SujayKrishna:
I know that we can do a `git: log` then choose a commit and then choose `Show commit` to get to this. But `git log` is sometimes not fast enough and so a shorthand for `git show X` would be good.

divmain:
Hi @SujayKrishna!

This is unlikely to become a standard feature - everybody has their own niche commands, and it would be untenable to support all of them explicitly.

However, we have supported a couple of methods to add your own simple features.  One of those ways is via [custom commands](https://github.com/divmain/GitSavvy/blob/master/docs/custom.md).

Another option might be something like adding the following to a `.sublime-commands` file in your `User` package, which would open an actual commit view:
```json
    {
        "caption": "git: show HEAD",
        "command": "gs_show_commit",
        "args": {
            "commit_hash": "HEAD"
        }
    },
```

Does that adequately address your needs?

stoivo:
Custom command could look like 
```
    {
        "caption": "git: show X",
        "command": "gs_custom",
        "args": {
            "args": ["show", "{PROMPT_ARG}"],
            "prompt_msg": "sha"
        }
    }
```

SujayKrishna:
Cool! Makes sense!

Thank you!

SujayKrishna:
@stoivo The custom command doesn't end up showing a ST3 tab though. 
I could see the output only by using `output_to_panel: true`.

Is there a flag I could use to open it as a new ST3 tab?

stoivo:
I added it now. Will be on master in a day or two.

SujayKrishna:
Thanks @stoivo

PR #611, opened 2017-03-01
Enhancement: Add setting to custom command to open in new buffer

stoivo:
Solves #610

### DIFF
--- a/core/commands/custom.py
+++ b/core/commands/custom.py
@@ -47,9 +47,11 @@ class GsCustomCommand(WindowCommand, GitCommand):
 
     def run_async(self,
                   output_to_panel=False,
+                  output_to_buffer=False,
                   args=None,
                   start_msg="Starting custom command...",
                   complete_msg="Completed custom command.",
+                  syntax=None,
                   run_in_thread=False,
                   custom_argument=None,
                   custom_environ=None):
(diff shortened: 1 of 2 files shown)


randy3k:
If the result of a git command is outputed to the buffer, I guess it would be helpful to have the option to set a specific syntax to the output. For example, it would be a `diff` or `show_commit` `sublime-syntax`.

stoivo:
Yes, can the setting be the path to the sublime-syntax?

How would I write spec for this to coveralls test not to fail?

divmain:
@stoivo don't worry about code coverage for now.  It's handy to see how a PR changes the metric, but until we invest serious time into test coverage, it should't be a blocker to merging.  Ping me on Gitter when this PR is ready, and I'll tweak the coveralls config to avoid this issue.

As to the merit of this PR, I'm on the fence.  Since folks are asking for it, I guess its okay.  But the problem may then become that people will expect full functionality that they'd have in a real GitSavvy diff view or commit view.  That would involve a much more significant effort.

_However,_ this is a small change to address a specific request, so its probably fine.

asfaltboy:
@divmain I feel the same way! [Your recommendation][1] in #610 was spot on to provide the full functionality we have in the shot command, including diff syntax, etc.
I do however appreciate the work that went into this PR and feel we should enhance custom commands whenever they are inadequate, though we should probably also update documentation.

[1]: https://github.com/divmain/GitSavvy/issues/610#issuecomment-283261040

stoivo:
I added syntax option. I think I would draw the line at interactive custom commands.

stoivo:
Thanks, I saw that you merged master into this. I rebased on top of master and pushed again, to eliminate merge commits.

asfaltboy:
Good catch! There should be a setting to make this button rebase instead of merge, i.e isaacs/github#88

PR #615, opened 2017-03-04
Improve Stashs

### LINKED ISSUE #600: Erratic behavior of stash commands in git: status
Disabled stash commands should be marked as such in the help. Pressing [t][a] when that command is actually not available does an [a] - stage all files. And there's no indication that pressing [t][a] will do something else entirely. I also think that pressing [t][a] should be a noop when the command doesn't make sense - and not disregard the initial [t]. Ideally, I think, pressing [t] should change the help text to show that "you're in stash mode", for example, by replacing the entire help text with just the help text of the stash commands for the duration of the command (without triggering a status refresh, that would be annoying).

### LINKED ISSUE #609: Enhancement: Ask confirmation for "discard stash" action
Today I lost some valuable time trying to figure out how to get back a dropped stash, and so I think this would definitely come in handy for most people.

Since `s` and `d` are so close its quite easy to accidentally discard a stash which makes this even more painful.

stoivo:
Fix: Trigger stash command outside of stash section
fixes #600

Fix: Drop a stash is destructive
fix #609

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -295,7 +295,6 @@
         "context": [
             { "key": "setting.command_mode", "operator": "equal", "operand": false },
             { "key": "setting.git_savvy.status_view", "operator": "equal", "operand": true },
-            { "key": "selector", "operator": "equal", "operand": "git-savvy.status meta.git-savvy.status.stashes meta.git-savvy.status.saved_stash", "match_all": true }
         ]
     },
     {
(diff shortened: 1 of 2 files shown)


stoivo:
As reasoned in #600 when hitting `t` + `a` and did not move the cursion to a stash, it would only stage all files. Doing the action is was the keyboard shortcut `a`.

divmain:
This was a very precise change, I expected it to be a bit uglier.  Nice work!

PR #631, opened 2017-04-06
Fix Issue 624

### LINKED ISSUE #624: Getting error when using 'diff file inline' for diff already opened
Starting on release 2.14.1, if I press 'l' on a file, I get its diff. If I go back to the status view, without closing the diff, and press 'l' again, nothing happens and I get the following error in the console:

```
  File "core.commands.inline_diff in /Users/user/Library/Application Support/Sublime Text 3/Installed Packages/GitSavvy.sublime-package", line 37, in <lambda>
  File "core.commands.inline_diff in /Users/user/Library/Application Support/Sublime Text 3/Installed Packages/GitSavvy.sublime-package", line 70, in run_async
  File "common.util.file in /Users/user/Library/Application Support/Sublime Text 3/Installed Packages/GitSavvy.sublime-package", line 55, in get_file_contents_binary
  File "./python3.3/posixpath.py", line 80, in join
AttributeError: 'NoneType' object has no attribute 'startswith'
```

jipumarino:
`self.file_path` and `self.repo_path` are None when view is cached.

Fixes #624

### DIFF
--- a/core/commands/inline_diff.py
+++ b/core/commands/inline_diff.py
@@ -67,7 +67,7 @@ class GsInlineDiffCommand(WindowCommand, GitCommand):
 
             inline_diff_views[view_key] = diff_view
 
-        file_binary = util.file.get_file_contents_binary(self.repo_path, self.file_path)
+        file_binary = util.file.get_file_contents_binary(settings["git_savvy.repo_path"], settings["git_savvy.file_path"])
         try:
             file_binary.decode()
         except UnicodeDecodeError as unicode_err:


jipumarino:
Thank you @stoivo, this was my first time using the "edit" feature from github, I didn't realize I was writing a commit message instead of the pull request title until it was done; I'll appreciate it if you change it.

I'm not sure yours is correct though, this feature was already implemented, I was just fixing an error introduced in https://github.com/divmain/GitSavvy/commit/77471fd120712b6ed0cbbf744308f070c7111261#diff-1d123c94cf2d78df2c8edace6fa6ee97R70

stoivo:
I just glared at the code. I don't understand why my path 77461fd broke this.

jipumarino:
Because of what I commented before. When the diff view is already open both `self.repo_path` and `self.file_path` are set to None at that point (not sure why that happens though).

When it's not open, those properties get set in this loop: https://github.com/divmain/GitSavvy/commit/77471fd120712b6ed0cbbf744308f070c7111261#diff-1d123c94cf2d78df2c8edace6fa6ee97L65

I tried replicating that loop in the other if branch, but it didn't work. This was the most straightforward solution.

stoivo:
Thanks again

PR #632, opened 2017-04-11
Sort branches by recent commited

yetithefoot:
**Problem:** I have a lot of branches in and extensively use branch switching. It's hard to navigate,  I have to search branch by name and then checkout.

**Proposed solution:** Show recent used branches on top of list in BRANCHES dashboard.

Since Git 2.7.0+ its [possible](https://www.kernel.org/pub/software/scm/git/docs/git-branch.html) to list branches with sort param. This PR adds option `sort_by_recent_in_branch_dashboard` to show recent branches on top.


yetithefoot:
@stoivo got it.
Git provides more options for sorting branches list. I thought about extracting this to settings but started with recently committed. If someone requests different sort options, I will be happy to contribute.

yetithefoot:
@stoivo you said that it should be one commit. It's possible to squash these commits during merge PR. It's OK but I'm just wondering why the policy is like that?

Also, it's a good practice to create `CONTRIBUTING.md` ([Github will process this file in special way](https://github.com/blog/1184-contributing-guidelines)) file to avoid this kind of issues..

stoivo:
What about having the option be a string and not set by default?
Add link to all options.

```python
    sort_by = savvy_settings.get("sort_by_in_branch_dashboard")
    stdout = self.git("branch", "-a", "-vv", "--no-abbrev", "--no-color", "--sort=-{}".format(sort_by) if sort_by else None)
```

yetithefoot:
@stoivo [here](https://github.com/git/git/blob/076c05393a047247ea723896289b48d6549ed7d0/ref-filter.c#L180) is a list of all available keys to sort.
In my opinion, there are only two meaningful keys: `committerdate` and `refname`. I can't imagine cases for other `keys`. I believe this is overflexibility to allow all of these fields.

core/git_mixins/branches.py
```diff
@@ -22,7 +22,9 @@ def get_branches(self):
         """
         Return a list of all local and remote branches.
         """
-        stdout = self.git("branch", "-a", "-vv", "--no-abbrev", "--no-color")
+        savvy_settings = sublime.load_settings("GitSavvy.sublime-settings")
+        sort_by_recent = savvy_settings.get("sort_by_recent_in_branch_dashboard")
+        stdout = self.git("branch", "-a", "-vv", "--no-abbrev", "--no-color", "--sort=-committerdate" if sort_by_recent else None)
         return (branch
```
randy3k: If the order of the branch list is only intended for the dashboard, the `sublime.load_settings` should not be called here. Instead, the argument of the call should be extended. And the setting call be fetch at [here](https://github.com/divmain/GitSavvy/blob/8cfeb94360d71e0b3f7b40ed9f130cfe10b23a90/core/interfaces/branch.py#L82-L82).

If your original idea is to sort the branch list in any conditions, I guess it will be a bit too ambitious.

```python
    def get_branches(self, sortby=None):
        """
        Return a list of all local and remote branches.
        """
        stdout = self.git(
            "branch", "-a", "-vv", "--no-abbrev", "--no-color",
            "--sort=-{}".formate(sortby) if sortby else None
            )
        return (branch
                for branch in (self._parse_branch_line(self, line) for line in stdout.split("\n"))
                if branch)
```

yetithefoot: @randy3k thanx for review. Pushed fixes to avoid global branches sorting.

core/git_mixins/branches.py
```diff
@@ -18,11 +18,11 @@
 
 class BranchesMixin():
 
-    def get_branches(self):
+    def get_branches(self, sort_by_recent=None):
```
randy3k: Should it be `True/False` value instead?

yetithefoot: Yep

PR #635, opened 2017-04-17
Improve: Remove some clutter from debug.log

stoivo:
It's not super nice but I prefer to have setting as arguments instead of state on an object.

### DIFF
--- a/core/commands/status_bar.py
+++ b/core/commands/status_bar.py
@@ -58,7 +58,7 @@ class GsUpdateStatusBarCommand(TextCommand, GitCommand):
             self.view.erase_status("gitsavvy-repo-status")
             return
 
-        with disable_logging():
+        with debug.disable_logging():
             short_status = self.get_branch_status_short()
         self.view.set_status("gitsavvy-repo-status", short_status)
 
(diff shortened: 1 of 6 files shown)


core/git_command.py
```diff
@@ -72,7 +72,8 @@ def git(self, *args,
             decode=True,
             encode=True,
             stdin_encoding="UTF-8",
-            custom_environ=None):
+            custom_environ=None,
+            skip_adding_to_log=False):
```
randy3k: it seems to me that it overkills. Maybe something like this (untested)?

```python
@contextmanager
def disable_logging():
    util.debug.stop_logging()
    try:
        yield
    finally:
        util.debug.start_logging()


with disable_logging():
    short_status = self.get_branch_status_short()
```

stoivo: start_logging() is resetting the log.
but it looks better

stoivo:
@randy3k, I did't know of contextmanager. This is so much cleaner :tada:

core/commands/status_bar.py
```diff
@@ -57,7 +58,8 @@ def run_async(self):
             self.view.erase_status("gitsavvy-repo-status")
             return
 
-        short_status = self.get_branch_status_short()
+        with disable_logging():
```
randy3k: debug.disable_logging?

stoivo: Yes, you are so right.

stoivo:
Thanks.

PR #636, opened 2017-04-18
Improve: Warn the user when we have permission error

### LINKED ISSUE #634: Inline diffs showing up empty on ST 3129
Hello!

Just tried to inline diff and then the basic diff via gitSavvy on multiple projects - the inline diff showed an empty tab, while the basic diff worked.

[Here's](https://pastebin.com/FVd9bjj6) a log file of me doing the following actions
1. Enabling the log on status screen
2. Pressing <kbd>L</kbd> for inline diff, then closing that tab when it comes up empty
3. Pressing <kbd>E</kbd> for basic diff on same file, then closing that tab 
4. Disabling log

stoivo:
fix #634

Looks like this.
screen shot 2017-04-18 at 09 44 17

### DIFF
--- a/common/theme_generator.py
+++ b/common/theme_generator.py
@@ -75,8 +75,12 @@ class ThemeGenerator():
                 out_f.write(STYLES_HEADER.encode("utf-8"))
                 out_f.write(ElementTree.tostring(self.plist, encoding="utf-8"))
         except PermissionError as e:
-            sublime.ok_cancel_dialog("GitSavvy was disk error: \n{}".format(e))
+            sublime.ok_cancel_dialog("GitSavvy could not write file to path: \n{}".format(e))
             raise e
+        except OSError as e:
+            sublime.ok_cancel_dialog("GitSavvy encountered a disk error: \n{}".format(e))
+            raise e
+
         return path_in_packages
 
     def apply_new_theme(self, name, target_view):
--- a/common/theme_generator.py
+++ b/common/theme_generator.py
@@ -70,10 +70,13 @@ class ThemeGenerator():
 
         full_path = os.path.join(sublime.packages_path(), path_in_packages)
 
-        with open(full_path, "wb") as out_f:
-            out_f.write(STYLES_HEADER.encode("utf-8"))
-            out_f.write(ElementTree.tostring(self.plist, encoding="utf-8"))
-
+        try:
+            with open(full_path, "wb") as out_f:
+                out_f.write(STYLES_HEADER.encode("utf-8"))
+                out_f.write(ElementTree.tostring(self.plist, encoding="utf-8"))
+        except PermissionError as e:
+            sublime.ok_cancel_dialog("GitSavvy was disk error: \n{}".format(e))
+            raise e
         return path_in_packages
 
     def apply_new_theme(self, name, target_view):


common/theme_generator.py
```diff
@@ -70,10 +70,13 @@ def get_new_theme_path(self, name):
 
         full_path = os.path.join(sublime.packages_path(), path_in_packages)
 
-        with open(full_path, "wb") as out_f:
-            out_f.write(STYLES_HEADER.encode("utf-8"))
-            out_f.write(ElementTree.tostring(self.plist, encoding="utf-8"))
-
+        try:
+            with open(full_path, "wb") as out_f:
```
asfaltboy: Searching the code, there are a 16 other calls to `with open()`, perhaps we should write a wrapping context manager function to handle open and catch `OSError` exceptions for all of them?

Most of the other file actions are on repo files, so it differs from #634 but could have a similar impact of "failing strangely".

common/theme_generator.py
```diff
@@ -70,10 +70,13 @@ def get_new_theme_path(self, name):
...
+                out_f.write(STYLES_HEADER.encode("utf-8"))
+                out_f.write(ElementTree.tostring(self.plist, encoding="utf-8"))
+        except PermissionError as e:
+            sublime.ok_cancel_dialog("GitSavvy was disk error: \n{}".format(e))
```
stoivo: Thanks

common/theme_generator.py
```diff
@@ -70,10 +70,13 @@ def get_new_theme_path(self, name):
...
 
-        with open(full_path, "wb") as out_f:
-            out_f.write(STYLES_HEADER.encode("utf-8"))
-            out_f.write(ElementTree.tostring(self.plist, encoding="utf-8"))
-
+        try:
+            with open(full_path, "wb") as out_f:
+                out_f.write(STYLES_HEADER.encode("utf-8"))
+                out_f.write(ElementTree.tostring(self.plist, encoding="utf-8"))
+        except PermissionError as e:
```
stoivo: Thanks

common/theme_generator.py
stoivo: Yeah, probably a good idea.

I not that good in python so possibly you you know good way to solve this.
If it was ruby we could do this

```python
# current
def get_new_theme_path(self, name):
    try:
        with open(full_path, "wb") as out_f:
            out_f.write(STYLES_HEADER.encode("utf-8"))
            out_f.write(ElementTree.tostring(self.plist, encoding="utf-8"))
    except PermissionError as e:
        sublime.ok_cancel_dialog("GitSavvy could not write file to path: \n{}".format(e))
        raise e
    except OSError as e:
        sublime.ok_cancel_dialog("GitSavvy encountered a disk error: \n{}".format(e))
        raise e

# What it could look like
def get_new_theme_path(self, name):
    common.file.open 
        with open(full_path, "wb") as out_f:
            out_f.write(STYLES_HEADER.encode("utf-8"))
            out_f.write(ElementTree.tostring(self.plist, encoding="utf-8"))

# in common.file
def open 
    try:
        yield
    except PermissionError as e:
        sublime.ok_cancel_dialog("GitSavvy could not write file to path: \n{}".format(e))
        raise e
    except OSError as e:
        sublime.ok_cancel_dialog("GitSavvy encountered a disk error: \n{}".format(e))
        raise e

```

stoivo: @asfaltboy, do you have a suggestion?

asfaltboy: My apologies, but on second look, it would be better to resolve the current issue as-is and later, introduce the full change a separate pull request.

Regardless, in Python, the `contextlib.contextmanager` decorator makes this sort of thing trivial (please note that I have not tested this code):

```python

import sublime


@contextmanager
def safe_open(filename, mode, *args, **kwargs):
    try:
        with open(filename, mode, *args, **kwargs) as file:
            yield file
    except PermissionError as e:
        sublime.ok_cancel_dialog("GitSavvy could not access file: \n{}".format(e))
        raise e
    except OSError as e:
        sublime.ok_cancel_dialog("GitSavvy encountered an OS error: \n{}".format(e))
        raise e
```

This creates a new context manager that can be anywhere open can:

```python
def get_new_theme_path(self, name):

    # ... other code

    with safe_open(full_path, "wb") as out_f:
        out_f.write(STYLES_HEADER.encode("utf-8"))
        out_f.write(ElementTree.tostring(self.plist, encoding="utf-8"))
```

divmain:
@stoivo, feel free to merge as soon as CI passes.

PR #640, opened 2017-04-28
Warn the user if GS encounter a disk error

stoivo:
followup to #636

I will use this branch to test that this work as expected. Would be cool it you could too :)

Some places I wrote some code which tries to open with utf-8 and if it fails try latin-1, and we don't want to have a `ok window` open.


asfaltboy:
I've been using the branch for a while with no side effects

PR #643, opened 2017-05-02
do not check if HEAD is rebased

### LINKED ISSUE #619: Squashing two commits should not depend on the base ref
Squashing two consecutive commits is always technically possible (there is no chance of a merge conflict), but GitSavvy will not do it if the base ref for the rebase indicates a merge conflict.

Similarly, squashing non-consecutive commits (let's call them A and B) should not depend on the base ref in the rebase screen (it should only depend on the commits starting after A and leading up to B).

ps I love GitSavvy

randy3k:
It is not necessary to check if HEAD is rebased. Manuplicating commits should be always possible (if there are no merges).

close #619

### DIFF
--- a/core/interfaces/rebase.py
+++ b/core/interfaces/rebase.py
@@ -470,7 +470,7 @@ class RewriteBase(TextCommand, GitCommand):
         if not self.interface.preserve_merges() and self.interface.contain_merges():
 
             sublime.message_dialog(
-                "Unble to manipulate merge commits unless in preserve merges mode."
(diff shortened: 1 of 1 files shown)


stoivo:
I don't really use this part of gitsavvy, so I'm not sure how it should be.

randy3k:
Most of the related code in this PR was written by @divmain, I guess he will be the best one to review this.

divmain:
Sorry for the long wait.  These changes look mostly cosmetic, and fit in well with well-used parts of the codebase.  Merging.

PR #648, opened 2017-05-15
Check that file exist before opening

### LINKED ISSUE #646: Rebase conflicts cause endless dialog fail
So this started happening, and I'm not able to close dialogs (i.e they open infinitely).

screen shot 2017-05-12 at 11 43 41

My GitSavvy was on commit 5631c53 and also continued when I checked out latest 8b3d313. So this is a recent regression.

I'm not sure which of the changes did it, I can try to bisect later if want.

stoivo:
close #646

### DIFF
--- a/core/git_mixins/rewrite.py
+++ b/core/git_mixins/rewrite.py
@@ -233,6 +233,8 @@ class RewriteMixin():
     def rebase_rewritten(self):
         if self.in_rebase_merge():
             path = os.path.join(self._rebase_merge_dir, "rewritten")
+            if not os.path.exists(path):
+                return dict()
             entries = []
             for sha in os.listdir(path):
                 with util.file.safe_open(os.path.join(path, sha), "r") as f:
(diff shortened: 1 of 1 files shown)


randy3k:
Let's also tag a new release after merging this~

stoivo:
A new release? should I make it

```
Changes since 2.14.1:

  Fix:
   - Show open inline diff if it already open form status dashboard
   - Push tags command is not blocking
   - added self.get_short_hash to honor core.abbrev settings (#642)

  Other:
   - Sort branches by recent commited (#632)
   - Check that file exist before opening (#648)

  Improvement:
   - Remove some clutter from debug.log (#635)
   - Warn the user when we have permission error (#636)
   - it is not necessary to check if HEAD is rebased
   - use message_dislog for show the messages
   - do not allow squashing merges

  Contributors:
   - Simon
   - Juan Pumarino
   - Pavel Savchenko
   - Vlad Tsepelev
   - Randy Lai
```
New commit a then a new tag

randy3k:
Sure!

PR #649, opened 2017-05-17
Release notes for 2.14.2

stoivo:
Since I can't push to master.


randy3k:
I remember to discuss the possibility of creating a `release` branch.

randy3k:
Anyway, let's merge this and release a new tag now.

stoivo:
2.14.2 is out. 
I think we should just keep it simple and ship the tests in the package. If the user is disk contains, we would probably not install GitSavvu in the first place. Less than 1 MB for all of the package files.   and 25 for .git.

PR #650, opened 2017-05-21
Enhance GitCommandFromTerminal w/ PULLREQ_EDITMSG

aristidesfl:
This is used by `hub pull-request` and is identical to a commit message

https://hub.github.com/

Related #402

PR #652, opened 2017-05-22
Fix: need to config username and email in appveyor

randy3k:
fix the appveyor failure at #650


stoivo:
Kind of strange that this made it all the way here.

PR #659, opened 2017-05-28
Improvement: refresh status bar and related fix

randy3k:
Do not refresh status bar using `GsInterfaceFocusEventListener`, it is handled by `GsStatusBarEventListener` already.

### DIFF
--- a/core/commands/status_bar.py
+++ b/core/commands/status_bar.py
@@ -35,7 +35,7 @@ class GsUpdateStatusBarCommand(TextCommand, GitCommand):
         global last_execution, update_status_bar_soon
         if sublime.load_settings("GitSavvy.sublime-settings").get("git_status_in_status_bar"):
 
-            millisec = int(round(time.time() * 100))
+            millisec = int(round(time.time() * 1000))
             # If we updated to less then 100 ms we don't need to update now but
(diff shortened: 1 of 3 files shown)


core/commands/status_bar.py
```diff
@@ -35,7 +35,7 @@ def run(self, edit):
         global last_execution, update_status_bar_soon
         if sublime.load_settings("GitSavvy.sublime-settings").get("git_status_in_status_bar"):
 
-            millisec = int(round(time.time() * 100))
+            millisec = int(round(time.time() * 1000))
             # If we updated to less then 100 ms we don't need to update now but
```
stoivo: Update the comment

randy3k: The comment is correct, the timeout is 100ms

stoivo: Ahh, and time.time() returns seconds. Thanks

stoivo:
Are there any specific reasons why you add `refresh_status_bar`, just to reduce the calls?

randy3k:
Both `GsStatusBarEventListener` and `GsInterfaceFocusEventListener` have a `on_activated` method to call `gs_update_status_bar`. This PR avoids duplicately calling to it.

PR #661, opened 2017-05-30
Blame walk

### LINKED ISSUE #102: Show current file feature

stoivo:
closed #102 #152 

First off, sorry for the big PR. Bit it is a big feature. 

I would like to have some review of the code of one other contributor.

Hello, @yyyc514, @benmosher, @stevage, @rchl, @reagle, @dlnsk, @jipumarino, @jipumarino, @jipumarino, @maxim, @zx8
Finally, I got around to implement this. If you are running GitSavvy from git pull down the branch a see if there is something we can improve further?

### DIFF
--- a/core/commands/blame.py
+++ b/core/commands/blame.py
@@ -351,7 +351,7 @@ class GsBlamePickCommitCommand(TextCommand, GitCommand):
         settings = self.view.settings()
         settings.set("git_savvy.commit_hash_old", settings.get("git_savvy.commit_hash"))
         lp = BlameCommitPanel(
-            self.commit_generator(),
+            self.commit_generator(follow=True),
             self.do_action,
             )
         lp.selected_commit(settings.get("git_savvy.commit_hash"))
--- a/core/commands/blame.py
+++ b/core/commands/blame.py
@@ -98,7 +98,7 @@ class GsBlameInitializeViewCommand(TextCommand, GitCommand):
 
     def get_content(self, ignore_whitespace=False, detect_move_or_copy=None, commit_hash=None):
         blame_porcelain = self.git(
-            "blame", "-p", '-w' if ignore_whitespace else None, detect_move_or_copy, commit_hash, self.file_path
+            "blame", "-p", '-w' if ignore_whitespace else None, detect_move_or_copy, commit_hash, "--", self.file_path
         )
         blame_porcelain = unicodedata.normalize('NFC', blame_porcelain)
         blamed_lines, commits = self.parse_blame(blame_porcelain.splitlines())
@@ -238,6 +238,8 @@ class GsBlameActionCommand(PanelActionMixin, TextCommand, GitCommand):
     default_actions = [
         ["open_commit", "Open Commit"],
         ["find_line_and_open", "Blame before selected commit"],
+        ["open", "Blame on one older commit", (), {'position': "older"}],
+        ["open", "Blame on one newer commit", (), {'position': "newer"}],
     ]
 
     @util.view.single_cursor_pt
@@ -297,5 +299,15 @@ class GsBlameActionCommand(PanelActionMixin, TextCommand, GitCommand):
             a = self.commit_before("older", commit_hash)
             self.view.settings().set("git_savvy.commit_hash", a)
 
+        self.view.run_command("gs_blame_initialize_view")
+
+    def open(self, position):
+        settings = self.view.settings()
+        commit_hash = settings.get("git_savvy.commit_hash")
+        if not commit_hash:
+            self.view.settings().set("git_savvy.commit_hash", self.newst_commit_for_file())
+        else:
+            previous_commit_hash = self.commit_before(position, settings.get("git_savvy.commit_hash"))
+            settings.set("git_savvy.commit_hash", previous_commit_hash)
         self.view.run_command("gs_blame_initialize_view")
 
--- a/core/commands/blame.py
+++ b/core/commands/blame.py
@@ -97,8 +97,23 @@ class GsBlameInitializeViewCommand(TextCommand, GitCommand):
             settings.erase("git_savvy.lineno")
 
     def get_content(self, ignore_whitespace=False, detect_move_or_copy=None, commit_hash=None):
+
+        if commit_hash:
+            # git blame does not follow file name changes like git log, therefor we
+            # need to look at the log first too see if the file has changed names since
+            # selected commit. I would not be surprised if this brakes in some special cases
+            # like rebased or multimerged commits
+            lines = self.git(
+                "log", "--oneline", "--follow", "--name-status", "--reverse",
+                "{}..HEAD".format(commit_hash),
+                "--", self.file_path
+            ).split("\n")
+            filename_at_commit = lines[1].split("\t")[1]
+        else:
+            filename_at_commit = self.file_path
+
         blame_porcelain = self.git(
-            "blame", "-p", '-w' if ignore_whitespace else None, detect_move_or_copy, commit_hash, "--", self.file_path
+            "blame", "-p", '-w' if ignore_whitespace else None, detect_move_or_copy, commit_hash, "--", filename_at_commit
         )
         blame_porcelain = unicodedata.normalize('NFC', blame_porcelain)
         blamed_lines, commits = self.parse_blame(blame_porcelain.splitlines())
--- a/core/commands/blame.py
+++ b/core/commands/blame.py
@@ -46,23 +46,17 @@ class GsBlameCommand(PanelActionMixin, WindowCommand, GitCommand):
         view.settings().set("git_savvy.blame_view", True)
         view.settings().set("git_savvy.repo_path", self.__repo_path)
         view.settings().set("git_savvy.file_path", self._file_path)
+        view.settings().set("git_savvy.lineno", self.coords[0] + 1)
         view.settings().set("git_savvy.commit_hash", self._commit_hash)
+        view.settings().set("git_savvy.ignore_whitespace", ignore_whitespace)
+        view.settings().set("git_savvy.detect_move_or_copy", option)
+
         view.settings().set("word_wrap", False)
         view.settings().set("line_numbers", False)
         view.settings().set('indent_guide_options', [])
-        view.set_name(
-            BLAME_TITLE.format(
-                self.get_rel_path(self._file_path),
-                " at {}".format(self._commit_hash) if self._commit_hash else ""
-            )
-        )
         view.set_scratch(True)
-        view.run_command("gs_blame_initialize_view", {
-            "coords": self.coords,
-            "ignore_whitespace": '-w' if ignore_whitespace else None,
-            "detect_move_or_copy": option,
-            "commit_hash": self._commit_hash
-        })
+
+        view.run_command("gs_blame_initialize_view")
 
     def pick_commit(self):
         self._branch = None
@@ -76,23 +70,35 @@ class GsBlameCommand(PanelActionMixin, WindowCommand, GitCommand):
 
 class GsBlameInitializeViewCommand(TextCommand, GitCommand):
 
-    def run(self, edit, coords=None, ignore_whitespace=None, detect_move_or_copy=None, commit_hash=None):
+    def run(self, edit):
+        settings = self.view.settings()
+        commit_hash = settings.get("git_savvy.commit_hash", None)
+        self.view.set_name(
+            BLAME_TITLE.format(
+                self.get_rel_path(self.file_path) if self.file_path else "unknown",
+                " at {}".format(commit_hash) if commit_hash else ""
+            )
+        )
         content = self.get_content(
-            ignore_whitespace=ignore_whitespace,
-            detect_move_or_copy=detect_move_or_copy,
+            ignore_whitespace=settings.get("git_savvy.ignore_whitespace", False),
+            detect_move_or_copy=settings.get("git_savvy.detect_move_or_copy", None),
             commit_hash=commit_hash
         )
-        self.view.sel().clear()
-        self.view.set_read_only(False)
-        self.view.replace(edit, sublime.Region(0, self.view.size()), content)
-        self.view.set_read_only(True)
 
-        if coords is not None:
-            self.scroll_to(coords)
+        self.view.run_command("gs_new_content_and_regions", {
+            "content": content,
+            "regions": {},
+            "nuke_cursors": False
+        })
 
-    def get_content(self, ignore_whitespace=None, detect_move_or_copy=None, commit_hash=None):
+        if settings.get("git_savvy.lineno", None) is not None:
+            self.scroll_to(settings.get("git_savvy.lineno"))
+            # Only scroll the first time
+            settings.erase("git_savvy.lineno")
+
+    def get_content(self, ignore_whitespace=False, detect_move_or_copy=None, commit_hash=None):
         blame_porcelain = self.git(
-            "blame", "-p", ignore_whitespace, detect_move_or_copy, commit_hash, self.file_path
+            "blame", "-p", '-w' if ignore_whitespace else None, detect_move_or_copy, commit_hash, self.file_path
         )
         blame_porcelain = unicodedata.normalize('NFC', blame_porcelain)
         blamed_lines, commits = self.parse_blame(blame_porcelain.splitlines())
@@ -216,8 +222,8 @@ class GsBlameInitializeViewCommand(TextCommand, GitCommand):
 
             yield output
 
-    def scroll_to(self, coords):
-        pattern = r".{{40}} \| {lineno: >4} ".format(lineno=coords[0] + 1)
+    def scroll_to(self, lineno):
+        pattern = r".{{40}} \| {lineno: >4} ".format(lineno=lineno)
         corresponding_region = self.view.find(pattern, 0)
         blame_view_pt = corresponding_region.b
 
--- a/core/commands/blame.py
+++ b/core/commands/blame.py
@@ -233,11 +233,19 @@ class GsBlameInitializeViewCommand(TextCommand, GitCommand):
         sublime.set_timeout_async(lambda: self.view.show_at_center(blame_view_pt), 0)
 
 
-class GsBlameOpenCommitCommand(TextCommand):
+class GsBlameActionCommand(PanelActionMixin, TextCommand, GitCommand):
+    selected_index = 0
+    default_actions = [
+        ["open_commit", "Open Commit"],
+    ]
 
     @util.view.single_cursor_pt
     def run(self, cursor_pt, edit):
-        hunk_start = util.view.get_instance_before_pt(self.view, cursor_pt, r"^\-+ \| \-+")
+        self.cursor_pt = cursor_pt
+        super().run()
+
+    def selected_commit_hash(self):
+        hunk_start = util.view.get_instance_before_pt(self.view, self.cursor_pt, r"^\-+ \| \-+")
         if hunk_start is None:
             short_hash_row = 1
         else:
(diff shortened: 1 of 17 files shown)


reagle:
I'm not confident I know how to use it but...

If I hit `enter` in a row I see options to move blame and, the one's I'm interested in, `open commit` and `show file at commit`.

open commit gives:

```
  File "/Users/reagle/Library/Application Support/Sublime Text 3/Packages/GitSavvy/core/ui_mixins/quick_panel.py", line 59, in on_action_selection
    func(*args, **kwargs)
  File "/Users/reagle/Library/Application Support/Sublime Text 3/Packages/GitSavvy/core/commands/blame.py", line 285, in open_commit
    self.view.window().run_command("gs_show_commit", {"commit_hash": short_hash})
NameError: global name 'short_hash' is not defined
```

show file at commit results in an untitled and blank tab with the following in the console:

```
`/usr/bin/git rev-parse --short` failed with following output:

fatal: Needed a single revision
```

reagle:
ac4215a and 215c08a fix my problem above. But the distinction between (a) "Show file at this commit" and (b) "Show file at this lines commit" is not clear.

(a) shows the file at the most recent commit? And (b) should be "line's".

stoivo:
(a) `Show file at commit` does open the file in the commit without the blame details. In case you want to read the file with syntax highlighting. Maybe `Show file without blame info`?

reagle:
I'm not sure what you mean about the blame details. Both options seem to open a version of the file from the past. (a) is HEAD and (b) is the commit corresponding to "this line's commit." 

Speaking of syntax highlighting, the resulting tabs that are opened have **no** syntax highlighting; they are opened as plain text for some reason....?

stoivo:
While you work on it, it is soo clear to you what everything means. It feels so obvious. Now after using it for a week, I understand how confusing it is.

So this pictures will explain it a bit better.
screen shot 2017-06-06 at 13 47 35


Syntax highlighting, Which language are you working with. It seams to work fine with .py extension.
mjvxeppfgi

reagle:
1. I still don't understand the first distinction, but I am a git novice. Does "show file without blame info" mean show the file at HEAD? What is "blame info"?
2. I suspect my highlighting problem is related to my use of [magicpython](https://github.com/MagicStack/MagicPython). (I have default python disabled). It opens other python files fine, just not here.

stoivo:
2. The reason why [magicpython](https://github.com/MagicStack/MagicPython) isn't working because it uses the old version of sublime syntax files. MagicPython is using `.tmLanguage`  while GitSavvy only checks for '.sublime-syntax`.
It is quite easy to fix for MagicPython if they want to. (You can create and issue)

1. HEAD is just a reference to where you are in the your git commits. When you checkout a branche, HEAD will now point to that branch. One branch is just a pointer to a specific commit, when  you commit on this branch you will update the branch and HEAD to point to at the new commit. This new commit then points back to the old commit.


1. git checkout fix
```
A - B - D 
      \  C
        ^ master
         ^ fix
```
HEAD -> C
master -> D
fix -> C


2. git commit -m"E"
```
A - B - D 
      \  C - E
        ^ master
             ^ fix
```
HEAD -> E
master -> D
fix -> E


3. git checkout master
```
A - B - D 
      \  C - E
        ^ master
             ^ fix
```
HEAD -> D
master -> D
fix -> E



4. git checkout C
```
A - B - D 
      \  C - E
        ^ master
             ^ fix
```
HEAD -> C
master -> D
fix -> E

At this stage, you are not on any branch. This is very useful if you want to look at how the code looked 2 weeks ago. If you make a commit when you have checked out a commit, it will be hard to find that commit again.

Hope That explain it a bit better. If you have other questions to git lets take it in the gitter chat. https://gitter.im/divmain/GitSavvy

reagle:
Okay, forgive my mentioning of HEAD, but what do you mean "without blame info." Could you show me what it is supposed to look like with blame info?

1st1:
@stoivo

> It is quite easy to fix for MagicPython if they want to. (You can create and issue)

Unfortunately is's not an "easy fix" for us. .sublime-syntax is a completely different format so it's not going to happen in the foreseeable future.

> MagicPython is using .tmLanguage while GitSavvy only checks for '.sublime-syntax`.

Maybe GitSavvy should check for `.tmLanguage` files too?

reagle:
In MagicStack/MagicPython/issues/86 they say they aren't likely to migrate away from .tmLanguage. Could you detect that in addition to .sublime-syntax?

stoivo:
> Could you show me what it is supposed to look like with blame info?

2017-06-07_08-01-08_commit 6524866 core commands blame py gitsavvy


I open a PR for sublime-syntax on MagicStack/MagicPython#86

reagle:
Okay. I would change:

1. "Show file without blame info" as "Show file at most recent commit"

I think "Show file at this line's commit" is okay, though it is really "at this row" or  "where the cursor is". For a while I thought "show file at this line's commit" meant that when it opened that version, it would automatically go to those line numbers, which would still be a neat feature.

reagle:
New thought for the second: "Show file at this chunk's commit"

stoivo:
@reagle, Thanks for naming commands with me. It is appreciated. 

I added one more feature. When to select show file. it jumps to the line you were at. And fixed some other bug when a file is renamed.

reagle:
Looks great and the goto line is indeed handy.

reagle:
BTW: I think the blame tab should be opened read-only, like the GRAPH tab.

core/ui_mixins/quick_panel.py
```diff
@@ -157,6 +157,7 @@ class PaginatedPanel:
     limit = 6000
     selected_index = None
     on_highlight = None
+    is_this_selected = None
```
randy3k: If I understand correctly, we need this because the index number of the selected item would be unknown if a generator is used. Instead of adding one more argument here, I suggest using the existing variable `selected_index`.

We could check if it is a callable. If it is, just do 

```
if selected_index(entry):
    bla lka bla
```

core/commands/blame.py
```diff
@@ -238,9 +274,132 @@ def run(self, cursor_pt, edit):
...
+        if commit_hash is None:
+            # Canceled panel
+            settings.set("git_savvy.commit_hash", settings.get("git_savvy.commit_hash_old"))
+            settings.erase("git_savvy.commit_hash_old")
+        else:
+            settings.set("git_savvy.commit_hash", commit_hash)
+        self.view.run_command("gs_blame_initialize_view")
+
+
+class BlameCommitPanel(PaginatedPanel):
```
randy3k: Would it make more sense to extend it from `LogPanel`?

core/git_mixins/history.py
```diff
@@ -63,12 +65,13 @@ def log(self, author=None, branch=None, file_path=None, start_end=None, cherry=N
 
         return entries
 
-    def commit_generator(self, limit = 6000):
+    def commit_generator(self, limit = 6000, follow=False):
```
randy3k: This is related to an earlier refactoring, should we actually pass the arguments `branch` and `file_path`?

core/commands/blame.py
```diff
@@ -238,9 +274,132 @@ def run(self, cursor_pt, edit):
...
+        if from_line:
+            commit_hash = self.selected_commit_hash() or 'HEAD'
+        else:
+            commit_hash = self.view.settings().get("git_savvy.commit_hash", "HEAD")
+
+        self.view.window().run_command("gs_show_file_at_commit", {
+            "commit_hash": commit_hash,
+            "filepath": self.file_path,
+            "lineno": self.find_lineno(),
+            "lang" : self.view.settings().get('git_savvy.original_syntax', None)
```
randy3k: Nice, any idea to patch the diff view? I don't think the same fix could be applied to diff view.

core/ui_mixins/quick_panel.py
stoivo: Yes, I squashed into 514b317, Blame option: Pick a new commit, if you are on a commit, set it as selected.

core/commands/blame.py
stoivo: Not really. The only thing in common is `format_item`.

randy3k: But one could argue that LogPanel and PaginatedPanel are only differed by format_item. The `on_selection` is almost identical.

randy3k: Also,  BlameCommitPanel and LogPanel are both history-related. I don't see why we shouldn't extend it from LogPanel.

core/git_mixins/history.py
stoivo: Yes. Generally, I prefer to pass several arguments then to set instance variables. On [log](https://github.com/divmain/GitSavvy/blob/53b11726f82733d186feaba0a3f0a90335b7ac27/core/git_mixins/history.py#L28-L28) it looks a little ugly, but it is very nice and explicit what's all the arguments.

randy3k: One less important point, this function returns a generator of log entries rather that commits. Maybe `log_generator` is a better name for this?

core/commands/blame.py
stoivo: I think is is already handled https://github.com/divmain/GitSavvy/blob/53b878a5891f15b90f8d796c34c16c44639d0f5e/core/commands/inline_diff.py#L40-L49

stoivo: Or which diff did you think of?

core/commands/blame.py
stoivo: Check [this](https://github.com/divmain/GitSavvy/pull/661/commits/0f1c2e76e3cb8f457c471cf0fac39b9a077dce18)

core/commands/blame.py
randy3k: The command `git: diff`.

core/git_mixins/history.py
```diff
@@ -63,12 +65,13 @@ def log(self, author=None, branch=None, file_path=None, start_end=None, cherry=N
 
         return entries
 
-    def commit_generator(self, limit = 6000):
+    def commit_generator(self, file_path, branch=None, limit=6000, follow=False):
```
randy3k: Is there any reason to switch the order of file_path and branch?

core/commands/log.py
```diff
@@ -21,12 +21,10 @@ class LogMixin(object):
     """
 
     def run(self, *args, file_path=None, branch=None):
-        self._file_path = file_path
-        self._branch = branch
-        sublime.set_timeout_async(self.run_async)
+        sublime.set_timeout_async(lambda: self.run_async(file_path, branch), 0)
```
randy3k: These changes are not necessary per se. But they do hurt.

core/git_mixins/history.py
randy3k: Perhaps something like this?
```python
    def commit_generator(self, limit=6000, **kwargs):
        # Generator for show_log_panel
        skip = 0
        while True:
            logs = self.log(limit=limit, skip=skip, **kwargs)
            if not logs:
                break
            for l in logs:
                yield l
            skip = skip + limit
       
```

core/commands/blame.py
stoivo: I can't think of any way to get that without opening the file and read the syntax, then close it. 👎

reagle:
BTW: "hunk" is already a term in git, so maybe we should use that instead of chunk.

core/git_mixins/history.py
stoivo: Yes

stoivo:
but I think of `hunk` as a park of the diff, and nothing else. I would stick with `chuck`

randy3k:
We also need a help pop up for blame view: https://github.com/divmain/GitSavvy/blob/master/popups/diff_view.html

Some more keybinds to navigate between commits should also be considered.

stoivo:
Should `,` and `.` go one commit forth and back based on the cursor or just jump one commit forward and backward compared to the commit?

randy3k:
I would say neither. Perhaps `,` and `.` to go up and down between chunks?

randy3k:
Closed it by accident

core/commands/blame.py
randy3k: We would open the file in the output panel to make it invisible to users.

randy3k: We could also capture the scopes of the file in the output panel to highlight the blame view.

randy3k: https://gist.github.com/mattst/71488757f2a2ca573c2d5e73af636d4c#file-getfileextensionsyntax-py

stoivo:
Still need to use creart the help pop up.

stoivo:
Popup:
Option 1.: 800x600
screen shot 2017-06-12 at 19 36 09
Option 2.: 1200x600
screen shot 2017-06-12 at 19 40 57

reagle:
When you say "blame on" do you mean that the commit is marked in some way? Is something done to that commit, or are you simply moving back in time?

core/git_mixins/history.py
stoivo: `log_generator` seem good since it if calls to log.

core/commands/blame.py
stoivo: Yes but this can be a separate PR.

stoivo:
You are time traveling.

reagle:
@stovio, some typos in help popup:

- redundant: "checkout out"
- in actions you use "chunk," later on "hung"
- s/coursor/cursor/

stoivo:
Thanks, option 1 or option 2?
Option 2 might cause issues is the user are on a small screen. 

TODO
- [ ] update https://github.com/divmain/GitSavvy/blob/88410f41f9dbb0661a3af06769d874239546fe9e/docs/history.md#git-blame-current-file

core/commands/log.py
```diff
@@ -21,13 +21,10 @@ class LogMixin(object):
...
-        self._file_path = file_path
-        self._branch = branch
-        sublime.set_timeout_async(self.run_async)
-
-    def run_async(self):
-        show_log_panel(self.commit_generator(), self.do_action)
+        sublime.set_timeout_async(lambda: self.run_async(file_path, branch), 0)
 
+    def run_async(self, file_path, branch):
+        show_log_panel(self.log_generator(file_path=file_path, branch=branch), self.do_action)
```
randy3k: need to also pass `file_path` to `do_action`.

stoivo: Why?

stoivo: Should we do it every place where we use show_log_panel

randy3k: I forgot why I said so, I remember a command was broken, may be `log current file`?

randy3k: Becasuse the new refactoring passes the variable `file_path` explicitly instead of using a instance variable.

randy3k: I am actually comfortable with the original code. However, if you would like to refactor it, the `file_path` variable has to be considered.

randy3k: I confirm that it broke `git: log current file`.

stoivo: Yes, it is broken. I will fix it.

stoivo:
Thanks. 
The difference  between     
Blame on one older commit -> Same as above but only jump one commit at a time. (Time travel)
Blame on one newer commit -> Same as above but only jump one commit at a time. (Time travel)
is the direction, one will take you further into the past while the other takes you closer to today's code. 


I think you missed some parts in
Blame before selected commit -> Find the chunk you are in and jump to the commit before that. This is handy because line numbers are likely to change over time. (Time travel)


What about this?

core/commands/log.py
stoivo: Now  I added `**kwards` where it is in use. Should I add `file_path=None` instead on those locations? I don't think we will have any other arguments passed around here any time soon so manybe it is better to only take `file_path`?
https://github.com/divmain/GitSavvy/pull/661/commits/c52f85c639c24eea7cadeb623c40ded76b5a29fa

reagle:
Oh, my HTML comments didn't show up (naturally, since this is markdown).

reagle:
What I don't understand is the "one commit at a time"; versus what? (I also removed "(Time travel)" since I don't think that helps...

```
In blame view you have a new set of commands

Use . to go to next chunk and , to go to previous. The tab title is the file name and commit hash.
If you hit Enter you will see a panel with all key bindings and actions.

Actions
Open Commit -> Open the commit of the chunk in which the cursor is located.
Blame before selected commit -> Find the chunk you are in and jump to the commit before that. This is handy because line numbers are likely to change over time. 

Blame on one older commit -> Same as above but only jump one commit at a time. 
Blame on one newer commit -> Same as above but only jump one commit at a time. 

Pick a new commit to blame -> Pick a new commit to checkout 
Show file at most recent commit -> Show the file with diff and syntax highlighting at the most recent commit

Show file at this chunk's commit -> Open the file at the version of the chunk in which the cursor is located.
```

stoivo:
`Blame before selected commit` might jump several commits.

core/commands/log.py
randy3k: I guess it is okay to use kwarg

asfaltboy:
@stoivo @randy3k  amazing work guys! I'm trying to add vintagious friendly mode to the blame command, at least for navigating hunks, and for some reason it's not working for me (see last commit 029e687). Any idea what I'm doing wrong?

core/commands/blame.py
```diff
@@ -71,8 +72,7 @@ def picked_commit(self, commit_hash):
         super().run()
 
 
-
-class GsBlameInitializeViewCommand(TextCommand, GitCommand):
+class GsBlameInitializeViewCommand(GsHandleVintageousCommand, TextCommand, GitCommand):
```
stoivo: I am sorry, thought of vintage support but apparently forgot. Sorry.
I think the issue is that either you need to call
`view.run_command("gs_handle_vintageous")` on line 65
Then you won't need to inherit it here.

This is who it's done in inline diff [core/commands/inline_diff.py#L84-L85](https://github.com/divmain/GitSavvy/blob/49ce7bdb83abc88485ef8a1582ad77a8e83e9ee8/core/commands/inline_diff.py#L84-L85)

stoivo:
@asfaltboy, this should do the trick.

stoivo:
Hi, I just want to tell you that I will leave for vacation on Wednesday next week, and will probably not reply or make any updated for the coming 5 weeks.

divmain:
@stoivo, what needs to be done for this to be merged?  If documentation is all that is left, we can document the remaining work in a separate issue, so that this can be merged prior to your vacation.  Also, have fun!

core/commands/log.py
stoivo: I have also noticed that reset commands are not working as they should.  Thanks for testing

randy3k: @asfaltboy you are absolutely right. 👍

stoivo:
@asfaltboy, I fixed the reset. 

Can you pull it and see if you can find any other bug?

asfaltboy:
Reset to branch works now, I'll keep using the branch for a while to see if anything else creeped in

divmain:
I pulled this down and it looks great to me.  How do you feel about merging into master and resolving any issues that pop up there?  This is a big branch, and my preference is to merge early in order to avoid merge conflicts and subtle errors from master/branch divergence.  @stoivo @asfaltboy?

asfaltboy:
@divmain agreed, good point!

stoivo:
I would like to merge

divmain:
Thanks a ton @stoivo!  This is merged :)

asfaltboy:
This also needs an update of [the relevant docs][1].

[1]: https://github.com/divmain/GitSavvy/blob/master/docs/history.md#action-in-blame-view

PR #663, opened 2017-05-31
Fix: Log all output even binaries

### LINKED ISSUE #662: GitSavvy: view recorded log throws an exception
When I mistakenly thought I encountered a bug, I tried to activate the log, but an exception was thrown when I ran _GitSavvy: view recorded log_ in the Command Palette. It's included below.
The last line in the exception contains the entire file I was working on, but I also attatched it in case it might help track down the bug (zipped, because Github wouldn't upload it otherwise).

**System info:**
* git 2.13.0
* OS X Yosemite10.10.5 (14F2109)
* ST Build 3132 (dev channel)

```
  File "/Applications/Sublime Text.app/Contents/MacOS/sublime_plugin.py", line 797, in run_
    return self.run()
  File "common.commands.debug in /Users/allan/Library/Application Support/Sublime Text 3/Installed Packages/GitSavvy.sublime-package", line 57, in run
  File "common.util.debug in /Users/allan/Library/Application Support/Sublime Text 3/Installed Packages/GitSavvy.sublime-package", line 35, in get_log
  File "./python3.3/json/__init__.py", line 240, in dumps
  File "./python3.3/json/encoder.py", line 193, in encode
  File "./python3.3/json/encoder.py", line 411, in _iterencode
  File "./python3.3/json/encoder.py", line 312, in _iterencode_list
  File "./python3.3/json/encoder.py", line 388, in _iterencode_dict
  File "./python3.3/json/encoder.py", line 422, in _iterencode
  File "./python3.3/json/encoder.py", line 173, in default
TypeError: b'import numpy as np\nimport pandas as pd\nfrom sklearn import decomposition\nimport multiprocessing\n\n\ndef color2igraphColor(color):\n    """Convert a color-list to an igraph color string with format \'rgb(#, #, #)\'.\n\n    Parameters\n    ----------\n    color : listlike\n        List with color values.\n\n    Returns\n    -------\n    TYPE\n        igraph compatible color string.\n    """\n    colorStr = \'rgb(\' + \', \'.join(map(str, [128, 64, 255])) + \')\'\n    return colorStr\n\n\ndef objDir2Df(obj):\n    """Given an object, return a DataFrame with all non-dunder parameters along with\n    theur type. If an exception are raised during determining the type, the except is\n    returned in place of the type. Object\n\n    Parameters\n    ----------\n    obj : Object\n        Object to get parameter info off.\n\n    Returns\n    -------\n    DataFrame\n        DataFrame with parameters and parameter types.\n    """\n    lst = list()\n    gen = (el for el in dir(obj) if not (el.startswith(\'__\') and el.endswith(\'__\')))\n    for el in gen:\n        try:\n            tup = [el, eval(\'type(pp.%s])\' % el)]\n        except Exception as e:\n            tup = [el, e]\n        try:\n            tup.append(eval(\'pp.%s.shape\' % el))\n        except Exception as e:\n            try:\n                tup.append(eval(\'len(pp.%s)\') % el)\n            except Exception as e:\n                tup.append(None)\n        lst.append(tup)\n    df = pd.DataFrame(lst, columns=[\'parameter\', \'partype\', \'shape\'])\n    return df\n\n\ndef nanEqual(a, b):\n    """Returs elementvise true for equal elements in inputs a and b, treading\n       NaN\'s as equal.\n       Inputs must both have the \'shape\'-attribute and be of equal shape.\n\n    Parameters\n    ----------\n    a : array\n        First array\n    b : array\n        Second array\n\n    Returns\n    -------\n    boolean array\n\n    Raises\n    ------\n    ValueError\n    The inputs aren\'t of equal shape.\n    """\n    if not (hasattr(a, \'shape\') or hasattr(b, \'shape\')):\n        raise ValueError("Inputs a nad b must both have the shape-attribute (like numpy arrays)")\n    if a.shape != b.shape:\n        raise ValueError("Inputs shapes must be identical. a.shape = {} and b.shape is {}"\n                         .format(a.shape, b.shape))\n    return np.all( (a == b) | (np.isnan(a) & np.isnan(b)) )  # noqa\n\n\ndef timedelta2unit(timedelta, unit):\n    """Convert a timedelta or iterable with Timedeltas to\n       a numeric value matching a given unit.\n\n    Parameters\n    ----------\n    timedelta : pd.Timedelta\n        Timedelta to convert. Can be an itetable of Timdeltas.\n    unit : str\n        Unit to convert timedeltas to. Valid arguments are {\'s\', \'h\', \'d\', \'y\'}.\n\n    Returns\n    -------\n    np.double\n        Numpy array of timedelta values matching given unit.\n\n    Raises\n    ------\n    ValueError\n    If input(s) lack(s) the \'total_seconds\' method.\n    """\n\n    def _inner(time, unit):\n        """Convert a pd.Timedelta (\'time\') to a numeric value matching \'unit\'self.\n\n        Parameters\n        ----------\n        time : pd.Timedelta\n            Timedelta to convert.\n        unit : str\n            String indicating unit, same as outer function.\n\n        Returns\n        -------\n        double\n            time converted to a double matching given \'unit\'\n\n        Raises\n        ------\n        ValueError\n            If input `time` is missing the attribute total_seconds.\n        """\n        if not hasattr(time, \'total_seconds\'):\n            raise ValueError((\'The input (or the elements of the iterable)\' +\n                              \'must have the attribute "total_seconds".\'))\n        ts = time.total_seconds()\n        if unit == \'s\':\n            return ts\n        if unit == \'h\':\n            return ts / 3600\n        if unit == \'d\':\n            return ts / (24 * 3600)\n        if unit == \'y\':\n            return ts / (365 * 24 * 3600)\n\n    if hasattr(timedelta, \'__iter__\'):\n        return np.fromiter((_inner(td, unit.lower()) for td in timedelta), np.double)\n    return _inner(timedelta, unit.lower())\n\n\ndef standardizeData(data, getStdMean=False):\n    """\n    Normalize the data by substracting the mean from each feature,\n    and dividing by the standard deviation.\n\n    Parameters\n    ----------\n    data : array\n        Data to be standardized columnwise.\n    getStdMean : bool, optional\n        return the standard deviation and mean for the data before the stnadardization,\n        along with the data iteself.\n\n    Returns\n    -------\n    array\n        The snandardized data, and depending on the flag getStdMean, also the std and\n        mean of the data.\n\n    """\n    mean = data.mean(axis=0)\n    std = data.std(axis=0)\n    # set values of where there\'s no data to 1.0, since we\'re dividing with the std\n    std[mean == 0] = 1.0\n    normData = (data - mean)/std\n    if getStdMean:\n        return (normData, std, mean)\n    return normData\n\n\ndef pcaFit(toPca, performStandardization=True, **kwargs):\n    """Standardize data, create a PCA object and fit the data.\n\n    Parameters\n    ----------\n    toPca : np.array\n        Data to perform PCA analysis on.\n    performStandardization : bool, optional\n        Standardize data if True, leave as be othewise.\n    **kwargs : dict, optional\n        Additional keyword arguments to PCA.\n\n    Returns\n    -------\n    PCA\n        Fitted PCA instance.\n    """\n    if performStandardization:\n        toPca, std, mean = standardizeData(toPca, getStdMean=True)\n    else:\n        mean, std = np.NaN, np.NaN\n    pca = decomposition.PCA(**kwargs)\n    pca.fit(toPca)\n    pca.norm_std = std\n    pca.norm_mean = mean\n    return pca\n\n\ndef icaFit(toIca, performStandardization=True, **kwargs):\n    """Standardize data, create a ICA object and fit the data.\n\n    Parameters\n    ----------\n    toIca : np.array\n        Data to perform ICA analysis on.\n    performStandardization : bool, optional\n        Standardize data if True, leave as be othewise.\n    random_state : np.random.RandomState, optional\n        Random state used for the analyses, default seed is int(time.time() * 1e6)\n    **kwargs : dict, optional\n        Additional keyword arguments to ICA.\n\n    Returns\n    -------\n    ICA\n        Fitted ICA instance.\n    """\n    if performStandardization:\n        toIca, std, mean = standardizeData(toIca, getStdMean=True)\n    else:\n        mean, std = np.NaN, np.NaN\n    ica = decomposition.FastICA(**kwargs)\n    ica.fit(toIca)\n    ica.norm_std = std\n    ica.norm_mean = mean\n    return ica\n\n\ndef mapAsync(func, funcargLst, n=None):\n    """Given a function and an iterable with tuple-packed arguments, evalueate the\n    the function in parallel using multiple processors.\n\n    Parameters\n    ----------\n    func : function\n        Function to evalueate.\n    funcargLst : list\n        List with tuples containing arguemnts for func.\n    n : int, optional\n        Number of processors to use. Default is 16 or number of processors - 1\n        (if there isn\'t 16 processors avaiable).\n\n    Returns\n    -------\n    list\n        List with result.\n    """\n    if n is None:\n        n = 16 if 16 < multiprocessing.cpu_count() else multiprocessing.cpu_count() - 1\n    with multiprocessing.Pool(n) as pool:\n        call = pool.starmap_async(func, funcargLst)  # parse args correctly\n        res = call.get()\n        pool.close()\n        pool.join()\n        return res\n\n\ndef lstDct2dct(lst):\n    """Merge a list of dicts into a single dict.\n    Raises warning if there\'s overlapping key values.\n\n    Parameters\n    ----------\n    lst : list\n        List with dictionaries.\n\n    Returns\n    -------\n    dict\n        Merged dictionary.\n    """\n    retDct = dict()\n    for dct in lst:\n        keyIntersection = set(dct.keys()).intersection(set(retDct.keys()))\n        if keyIntersection:\n            Warning(\'Keys are being overwritten: {}\'.format(keyIntersection))\n        retDct.update(dct)\n    return retDct\n\n\ndef randomSample(itr, n):\n    """Draw n samples from iterable, where each element can only be drawn once.\n\n    Parameters\n    ----------\n    itr : iterable\n        Iterable to draw from.\n    n : int\n        number of elements to draw.\n\n    Returns\n    -------\n    list\n        n elements from itr.\n\n    Raises\n    ------\n    ValueError\n    If the requested number of draws isn\'t smaller then the length if itr.\n    """\n    if len(itr) <= n:\n        raise ValueError("\'itr\' must be longer that n")\n    idx = np.arange(len(itr))\n    np.random.shuffle(idx)\n    ret = [itr[i] for i in idx[:n]]\n    return ret\n\n\ndef getFirstDayInTimeseries(ts):\n    """Given a Pandas time series, get the Timestamp for the start of the day for the\n    earliest entry in the time series.\n\n    Parameters\n    ----------\n    ts : Series[Tiemstamp]\n        Pandas Series with Timestamps (datetime might also work).\n\n    Returns\n    -------\n    Timestamp\n        Pandas Timestamp matching the data for the first entry in ts.\n    """\n    t0 = ts.min()\n    t0d = t0.date()\n    return pd.Timestamp(t0d)\n\n\ndef swapMatrixRows(m, r0, r1, inplace=False):\n    if inplace:\n        m[[r0, r1]] = m[[r1, r0]]\n        return None\n    # Don\'t modify matrix inplace\n    mt = m.copy()\n    mt[[r0, r1]] = mt[[r1, r0]]\n    return mt\n\n\ndef swapMatrixCols(m, c0, c1, inplace=False):\n    if inplace:\n        col0 = m[:, c0].copy()\n        m[:, c0] = m[:, c1]\n        m[:, c1] = col0\n        return None\n    # Don\'t modify the matrix inplace\n    mt = m.copy()\n    mt[:, c0] = m[:, c1]  # copy from original, unmodified matrix\n    mt[:, c1] = m[:, c0]  # copy from original, unmodified matrix\n    return mt\n\n' is not JSON serializable
```

[misc.py.zip](https://github.com/divmain/GitSavvy/files/1042046/misc.py.zip)

stoivo:
Some command in inline diff are workning with binaries so
when we try to log. Try to decode it as utf-8, otherwise
fail.

closes #662

### DIFF
--- a/common/util/debug.py
+++ b/common/util/debug.py
@@ -7,6 +7,7 @@ from contextlib import contextmanager
 
 _log = []
 enabled = False
+ENCODING_NOT_UTF8 = "{} was sent as binaries and we dont know the encoding, not utf-8"
 
 
 def start_logging():
@@ -41,15 +42,32 @@ def add_to_log(obj):
 
 
 def log_git(command, stdin, stdout, stderr, secounds):
-    add_to_log({
+    message = {
         "type": "git",
         "command": command,
         "stdin": stdin,
         "stdout": stdout,
         "stderr": stderr,
         "secounds": secounds
-        })
+        }
 
+    if stdin.__class__ == bytes:
+        message["stdin"] = try_to_decode(stdin, "stdin")
+
+    if stdout.__class__ == bytes:
+        message["stdout"] = try_to_decode(stdout, "stdout")
+
+    if stderr.__class__ == bytes:
+        message["stderr"] = try_to_decode(stderr, "stderr")
+
+    add_to_log(message)
+
+
+def try_to_decode(message, name):
+    try:
+        return message.decode(),
+    except UnicodeDecodeError:
+        return ENCODING_NOT_UTF8.format(name)
 
 def log_error(err):
     add_to_log({


stoivo:
A review?

randy3k:
I am not sure what was the issue and why it fixes the issue....perhaps @divmain may know more.

stoivo:
In this PR https://github.com/divmain/GitSavvy/pull/606 we do some command without decoding the output because we are sending it back in some other command. I made it pass the binary around therefor I would need to deal with the encoding and it would always keep the correct encoding. 

And I make a bug. When you would log this command which only worked with binaries it would save the messages. Here https://github.com/stoivo/GitSavvy/blob/409e5eefbf39f53c31d3d6253a6ea5246e7bc583/common/util/debug.py#L36 when we dump it to json if fails since json.dump don't work with binaries.

common/util/debug.py
```diff
@@ -7,6 +7,7 @@
 
 _log = []
 enabled = False
+encoding_not_utf8 = "{} was sent as binaries and we dont know the encoding, not utf-8"
```
randy3k: I guess it is a convention to use upper cases for such constant!?

stoivo: Does python make it immutable if it is uppercase?

asfaltboy: Strings are always immutable, and names can always be overridden, no matter the casing. It's only a convention for readability sake.

stoivo: That's different than ruby, which I have more experience with.

stoivo:
Merge?

Issue #664, opened 2017-06-12
Allow me to disable the "initialize repository" prompt


firelizzard18:
I do not ever use GitSavvy to create repos. I will not ever. I manually create or clone every repo. If I were to use GitSavvy for this, I would do it manually. The initialize repository prompt is obnoxious and serves absolutely no purpose for me.

Please give me the option of disabling this 'feature'!

stoivo:
I have it on my todo list, it will be in the next release.

Issue #589, opened 2017-01-04
Keeps asking me to initialise repository


FPtje:
There are folders in my sublime project that don't have git initialised. That should be fine, but GitSavvy (Windows) complains at every possible opportunity. When I open the Ctrl + P file menu and type something, a popup appears asking whether I want to initialise a git repository.

I don't. Stop asking me.

Also, these errors are printed in the sublime console:

```
Traceback (most recent call last):
  File "core.git_command in C:\Users\fpeij\AppData\Roaming\Sublime Text 3\Installed Packages\GitSavvy.sublime-package", line 133, in git
  File "core.git_command in C:\Users\fpeij\AppData\Roaming\Sublime Text 3\Installed Packages\GitSavvy.sublime-package", line 215, in repo_path
  File "core.git_command in C:\Users\fpeij\AppData\Roaming\Sublime Text 3\Installed Packages\GitSavvy.sublime-package", line 255, in _repo_path
  File "core.git_command in C:\Users\fpeij\AppData\Roaming\Sublime Text 3\Installed Packages\GitSavvy.sublime-package", line 156, in git
  File "core.git_command in C:\Users\fpeij\AppData\Roaming\Sublime Text 3\Installed Packages\GitSavvy.sublime-package", line 119, in raise_error
GitSavvy.core.git_command.GitSavvyError: `C:\Program Files\Git\cmd\git.EXE rev-parse --show-toplevel` failed with following output:

fatal: Not a git repository (or any of the parent directories): .git


During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "common.ui in C:\Users\fpeij\AppData\Roaming\Sublime Text 3\Installed Packages\GitSavvy.sublime-package", line 289, in run_async
  File "common.ui in C:\Users\fpeij\AppData\Roaming\Sublime Text 3\Installed Packages\GitSavvy.sublime-package", line 65, in __init__
  File "common.ui in C:\Users\fpeij\AppData\Roaming\Sublime Text 3\Installed Packages\GitSavvy.sublime-package", line 101, in render
  File "common.ui in C:\Users\fpeij\AppData\Roaming\Sublime Text 3\Installed Packages\GitSavvy.sublime-package", line 117, in _render_template
  File "common.ui in C:\Users\fpeij\AppData\Roaming\Sublime Text 3\Installed Packages\GitSavvy.sublime-package", line 155, in get_keyed_content
  File "./python3.3/collections/__init__.py", line 56, in __init__
[...]

PR #665, opened 2017-06-12
Adds a setting to disable git init prompts

p3lim:
As discussed in #589 and solves #664.

This could probably also have been implemented directly in [core/git_command.py](https://github.com/divmain/GitSavvy/blob/2b10772530437f7088f8fcebdba459f05efb2c7c/core/git_command.py#L101).

Feel free to squash or make any changes.


stoivo:
Greate, thanks.

Issue #671, opened 2017-06-15
Show diffstat in git graph/log status message


3v1n0:
Below the commit message, in the status window, it would be nice to see the diffstat of the commit, so that it can be easier to understand what has been changed.

Issue #488, opened 2016-08-04
More informative commit view


yetithefoot:
When im writing commit message I prefer to see changes that will be commited and branch info.

Now it looks like:
<img width="794" alt="screenshot 2016-08-04 17 51 19" src="https://cloud.githubusercontent.com/assets/699027/17406453/1870e844-5a6c-11e6-8da3-7aef85f4ecb9.png">

But view like this looks more informative:
<img width="698" alt="screenshot 2016-08-04 17 51 28" src="https://cloud.githubusercontent.com/assets/699027/17406460/2404cc3e-5a6c-11e6-90a9-90c02984d514.png">

Suggest extend commit view with extra information

divmain:
Would [this option](https://github.com/divmain/GitSavvy/blob/master/GitSavvy.sublime-settings#L48-L52) suffice?

aziz:
I think full diff is too much I also like to see a summary of things that are being commited.

yetithefoot:
@divmain Not really. As @aziz mentionned, its too much.
Probably smth like this would cover most of possible scenarios:
```
{
    "show_commit_diff": "full", // full changes
    "show_commit_diff": "filenames", // only filenames
    "show_commit_diff": "none" // nothing
}
```
In my cases filenames is enough.

divmain:
In #492, @tomngo asked for previous commit messages, since commit messages will often be similar.  Let's come up with all the possible bits of information we might want in the commit view, and then expose a config option to allow people to set this up granularly.

One way would be to just accept a raw `git` command.  You could pretty simply craft one to show previous commit subjects, or show a full diff, or show filenames that have changed.  We could also accept an array of these, for people that want lots of info and don't want to choose between them.

However, this would be more heavy-handed, and exposing more constrained but popular choices as boolean options might be better.

Thoughts?

tomngo:
Thanks! For convenience, I'd prefer the more constrained but popular choices. If there were a way to enhance by adding raw `git` commands to user preferences, I'd likely use that.

rchl:
I'd really like this improvement. One thing that wasn't mentioned is showing the branch we are on. I'd like that too.

rchl:
Ability to select previous commit messages would also sort-of solve #503.

asfaltboy:
@tomngo could you elaborate on what you mean by "'enhancing by adding raw `git` commands" ? 

[...]

PR #672, opened 2017-06-15
Diffstat

stoivo:
closing #671
screen shot 2017-06-15 at 14 18 28

### DIFF
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -134,7 +134,7 @@
         //   to use columns globally, --no-columns should be added here.
         // "branch": ["--no-columns"]
     },
-        
+
     /*
     Prints branches sorted by recent commit (works with git 2.7.0+)
     https://www.kernel.org/pub/software/scm/git/docs/git-branch.html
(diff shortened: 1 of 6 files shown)


randy3k:
👍 good job.

randy3k:
Actually, I would also like to see a similar summary when typing the commit message #488.

GitSavvy.sublime-settings
```diff
@@ -166,6 +166,18 @@
...
     /*
+        Diffstat look like
+        core/commands/blame.py     | 13 +++++--------
+        core/commands/log.py       |  9 +++------
+        core/git_mixins/history.py |  8 ++------
+        3 files changed, 10 insertions(+), 20 deletions(-)
+
+        Set to `false` if don't want to show this when view a commit.
+     */
+    "show_diffstat": true,
```
randy3k: may be more clear to call it `log_show_diffstat`?

stoivo: I think we can keep it all in one setting since if you are interested in diffstat somewhere you would probably like so see it elsewhere. 
Maybe we can have one setting for all views and one setting to display it in commit view

randy3k: Fair enough. #488 is a separated issue.

Issue #676, opened 2017-06-20
All '*' are translated into '●' in git graph


3v1n0:
This https://github.com/3v1n0/sublime-glib-snippets/commit/2b0ea67a51c70eadaaae16280e30f6baa941c277 is not well rendered in git graph:

![immagine](https://user-images.githubusercontent.com/345675/27340376-9c372c96-55da-11e7-833d-698513c63ea0.png)

I guess the regex for replacing it should be something like `/^\*/●/`

PR #677, opened 2017-06-20
Fix: graph, only replace the '*' before the commit sha

3v1n0:
The commit message might include asterisks everywhere, we just have to replace the first one.

Fixes Issue #676

### DIFF
--- a/core/commands/log_graph.py
+++ b/core/commands/log_graph.py
@@ -8,7 +8,7 @@ from ...common import util
 
 COMMIT_NODE_CHAR = "●"
 COMMIT_NODE_CHAR_OPTIONS = "●*"
-COMMIT_LINE = re.compile("[%s][ /_\|\-.]*([a-z0-9]{3,})" % COMMIT_NODE_CHAR_OPTIONS)
+COMMIT_LINE = re.compile("^[%s][ /_\|\-.]*([a-z0-9]{3,})" % COMMIT_NODE_CHAR_OPTIONS)
 
 
 class LogGraphMixin(object):
@@ -61,7 +61,7 @@ class GsLogGraphRefreshCommand(TextCommand, GitCommand):
 
         args = self.view.settings().get("git_savvy.git_graph_args")
         graph_content += self.git(*args)
-        graph_content = graph_content.replace("*", COMMIT_NODE_CHAR)
+        graph_content = re.sub('^\*', COMMIT_NODE_CHAR, graph_content, flags=re.MULTILINE)
 
         self.view.run_command("gs_replace_view_text", {"text": graph_content, "nuke_cursors": True})
         self.view.run_command("gs_log_graph_more_info")
--- a/core/commands/log_graph.py
+++ b/core/commands/log_graph.py
@@ -61,7 +61,7 @@ class GsLogGraphRefreshCommand(TextCommand, GitCommand):
 
         args = self.view.settings().get("git_savvy.git_graph_args")
         graph_content += self.git(*args)
-        graph_content = re.sub('^\*', COMMIT_NODE_CHAR, graph_content, flags=re.MULTILINE)
+        graph_content = re.sub('(^[\| ]*)\*', r'\1'+COMMIT_NODE_CHAR, graph_content, flags=re.MULTILINE)
 
         self.view.run_command("gs_replace_view_text", {"text": graph_content, "nuke_cursors": True})
         self.view.run_command("gs_log_graph_more_info")
--- a/core/commands/log_graph.py
+++ b/core/commands/log_graph.py
@@ -8,7 +8,7 @@ from ...common import util
 
 COMMIT_NODE_CHAR = "●"
 COMMIT_NODE_CHAR_OPTIONS = "●*"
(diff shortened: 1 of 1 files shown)


core/commands/log_graph.py
```diff
@@ -8,7 +8,7 @@
 
 COMMIT_NODE_CHAR = "●"
 COMMIT_NODE_CHAR_OPTIONS = "●*"
-COMMIT_LINE = re.compile("[%s][ /_\|\-.]*([a-z0-9]{3,})" % COMMIT_NODE_CHAR_OPTIONS)
+COMMIT_LINE = re.compile("^[%s][ /_\|\-.]*([a-z0-9]{3,})" % COMMIT_NODE_CHAR_OPTIONS)
```
asfaltboy: Won't this make [running an action][1] fail on any commits that don't start with `*`/`●` ? E.g when viewing a commit in a branch, it often starts with `|` such as:

```
●   a573d30 Merge pull request #672 from stoivo/diffstat (4 days ago) <Simon>
|\  
| ● 1d22b92 Settings: New setting `show_diffstat` defaults to true (5 days ago) <Simon>
| ● 56723e7 Enhancement: Show diffstat when showing a commit (5 days ago) <Simon>
|/  
● 1680baa Adds a setting to disable git init prompts (#665) (8 days ago) <Adrian L Lange>
```

[1]: https://github.com/3v1n0/GitSavvy/blob/a06d40d761f3c363db6e92b03bc6935bbe2fdeb8/core/commands/log_graph.py#L196

core/commands/log_graph.py
```diff
@@ -61,7 +61,7 @@ def run(self, edit):
 
         args = self.view.settings().get("git_savvy.git_graph_args")
         graph_content += self.git(*args)
-        graph_content = graph_content.replace("*", COMMIT_NODE_CHAR)
+        graph_content = re.sub('^\*', COMMIT_NODE_CHAR, graph_content, flags=re.MULTILINE)
```
asfaltboy: Same as in line 11, this may require a more complex regex.

stoivo:
Thanks, we <3 new contributes. We would like to mention you in the release message if that's ok by you?

core/commands/log_graph.py
```diff
@@ -8,7 +8,7 @@
 
 COMMIT_NODE_CHAR = "●"
 COMMIT_NODE_CHAR_OPTIONS = "●*"
-COMMIT_LINE = re.compile("^[%s][ /_\|\-.]*([a-z0-9]{3,})" % COMMIT_NODE_CHAR_OPTIONS)
+COMMIT_LINE = re.compile("^[\| ]*[%s][ /_\|\-.]*(?P<commit_hash>[a-z0-9]{3,})" % COMMIT_NODE_CHAR_OPTIONS)
```
stoivo: I think you should use 
GRAPH_CHAR_OPTIONS = " /_\|\-."
```
COMMIT_LINE = re.compile("^[%s]*[%s][%s]*(?P<commit_hash>[a-z0-9]{3,})" % COMMIT_LINE COMMIT_NODE_CHAR_OPTIONS COMMIT_LINE)
```
with --no-color
You can get all kind of strange graphs with multimerge.

3v1n0:
Ok, I've updated the regexes, including some cleanups too, however I think the `COMMIT_LINE` one can be simplified, since there's no case where the commit hash is after a \, /, . or -.

Thus `^[\| ]*[%s] (?P<commit_hash>[a-z0-9]{3,})" % COMMIT_NODE_CHAR_OPTIONS` is enough

Also an hash can't contain g-z keys, and should be at least 5 chars... Which will make it

`"^[\| ]*[%s] (?P<commit_hash>[a-f0-9]{5,40})" % COMMIT_NODE_CHAR_OPTIONS`

core/commands/log_graph.py
3v1n0: Oh, this (reusing the graph char sets) was my idea too... But i didn't find any case like that... But then it makes sense.

stoivo:
It looks good.
I will pull it down a see if it works like expected🖒⚅

stoivo:
Damn it❗️ I can make it 💥  :)

I will merge it.
What I did find is a bug in git.
I zipped the repo and attached it. 
[loooool.zip](https://github.com/divmain/GitSavvy/files/1097771/loooool.zip)

When you run `git log --all --graph --oneline --no-color`
You will get 
```
$ git log --all --graph --oneline --no-color
* 1948764 (HEAD -> master) f
| *-.   3fe5938 (master_merge) Merge branches 'one' and 'two' into master_merge
| |\ \
|/ / /
| | * 0a8f459 (two) ccc
| | * 2084353 c
| | * 006bcdd c
* | | f8ceeb0 d
| |/
|/|
| | *   9b42732 (refs/stash) On one: a
| | |\
| |/ /
| | * c12cffd index on one: c3bba58 bb
| |/
| * c3bba58 (one) bb
| * 18fd299 aa
|/
* fe67af3 b
* 5e42cd1 a
```
`-.` is printed in color, which it should not.

I will try to report it to git. I'm not sure how easily it can be done since I think git issues is tracked in mailing lists.

Issue #679, opened 2017-06-23
Amending the very first commit of a repo fails


3v1n0:
It's classic, you fail your first commit of a new repo... :-D

But amending crashes:
```
`/usr/bin/git diff --no-color --cached HEAD^` failed with following output:

fatal: ambiguous argument 'HEAD^': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

asfaltboy:
This (and other oddities) were reported by myself a while back, in #557 . Unfortunately, no one has yet to tackle these yet.

stoivo:
Since it is a duplicate we will close this issue.

Issue #557, opened 2016-11-05
Odd behavior on bare repo / first commit


asfaltboy:
Here are some things I noticed that are broken when acting on a "bare repo" (or a repo with 1 commit):

- [x] BARE: Stage dir + unstage (reset) a specified file

    ### Exception

      `/usr/bin/git reset HEAD testproj/__init__.py` failed with following output:
      fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.

    ### Solution

      $ git rm --cached testproj/__init__.py

- [x] BARE: Amend 1st commit (oops forgot .gitignore):

    ### Exception

      /usr/bin/git diff --no-color --cached HEAD^` failed with following output:
      fatal: ambiguous argument 'HEAD^': unknown revision or path not in the working tree.

    ### Solution

      $ git commit --amend  # works well on command line

- [ ] BARE: Log/RefLog/Graph/Reset commands all fail since head does not exist.

    ### Exception

      # any git command would fail with this message
      fatal: your current branch 'master' does not have any commits yet

    ### Solution

    Nothing much can be done here, but perhaps we can show a nicer message if we expect this to happen.

- [ ] 1 COMMIT: Reset (log/reflog) from 1st to "zero-commit" only 1 record displayed, can't go back to nothing. This is probably ok, and iI couldn't find a workaround other than deleting ".git" dir.

_NOTE:_ This has most likely been like this for a while but we never bothered to clean it up.

_NOTE 2:_ I'm probably going to want to convert this into a PR with failing tests, once I figure how to write and run them 😊

kaste:
The first two issues herein have been fixed.  The others are not fixable.

kaste:
Note: Unstaging a file has been fixed with bcacd55e (Make staging/unstaging multiple files one operation) (in v2.20.0, Dec 2019)

Amending the first commit has been fixed in #1419 (v2.27.5)

PR #680, opened 2017-06-24
[WIP] Fix: amend commit failure for first commit

asfaltboy:
As described in #557 and  #679 and reproduced in failing test (last commit 1ca6dd3)

### DIFF
--- a/tests/test_core_commands/test_gs_commit.py
+++ b/tests/test_core_commands/test_gs_commit.py
@@ -1,0 +1,98 @@
+import os
+import subprocess
+
+from GitSavvy.core import git_command
+from GitSavvy.tests.common import GitRepoTestCase
+from GitSavvy.common.util import super_key
+
+
+class CommonCommitMixin(object):
+    """ Help methods for commit command testing """
+
+    def get_view(self):
+        """
+        Get the currently active view, set it to self.view and
+        return it's line_count
+        """
+        self.view = self.window.active_view()
+        line_count = self.view.rowcol(self.view.size())[0]
+        # print('Got view: %s, line count: %s' % (self.view, line_count))
+        return line_count
+
+    def assert_commit_view_rendered(self, first_line=None):
+        """ Assert the commit view is correctly rendered """
+        yield self.get_view  # yield back control when view has any lines
+        first_rows = self.get_rows(*range(0, 2))
+        self.assertEqual(first_rows, [first_line or '', ''])
+
+        first_comment_row = self.get_row(2)
+        self.assertEqual(first_comment_row, '## To make a commit, type your '
+                         'commit message and press {key}-ENTER. To cancel'
+                         .format(key=super_key))
+
+    def finish_commit(self, message=None):
+        """
+        Execute the commit while inserting the
+        given message in the start of the buffer
+        """
+        if message:
+            self.view.run_command("insert", {"characters": message}),
+        self.window.run_command('gs_commit_view_do_commit')
+        yield lambda: not self.window.views()  # yield control when view closed
+
+    def assert_commit_log_updated(self, message, count=2):
+        """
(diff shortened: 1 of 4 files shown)


tests/test_core_commands/test_gs_commit.py
```diff
@@ -0,0 +1,98 @@
...
+    def test_add_and_commit_file(self):
+        """ Test adding a new file and committing """
+        foo = os.path.join(self.repo_path, "foo")
+        with open(foo, "w") as f:
+            f.write("foo")
+
+        self.stage_file(foo)
+        self.window.run_command("gs_commit")
+
+        yield from self.assert_commit_view_rendered(first_line=None)
```
stoivo: Can you explain what `yield` yields back too?

asfaltboy: This specific `yield from` calls into every yield command in the `self.assert_commit_view_rendered` method. Without this `yield from`, simply calling the method would not call the yield expression inside the method, thus failing the test.

For more information see here [the changelog description in Python 3.3][1].

The reason we use yield at all here, is due to the commands' use of `sublime.set_timeout_async` which basically defers and returns control back to the text thread. This a feature of @randy3k 's [excellent UnitTesting plugin][2].

[1]: https://docs.python.org/3/whatsnew/3.3.html#pep-380-syntax-for-delegating-to-a-subgenerator
[2]: https://github.com/randy3k/UnitTesting-example/tree/090f271592788d584d41b92e810b0c7577e066e9

stoivo: Ok

asfaltboy:
Hey all, I've rebased over latest `dev` branch and added a relatively simple fix that uses log instead of diff to get the content of last commit ... However, it's not realistic diff, as it excludes the staged changes. Does anyone know of a way to get the diff of the "zero-eth" commit?

TODOs:

- [ ] need to improve diff display (either hide it or learn to include staged
      changes) - google it!
- [ ] we also need to test for all the flags we allow to supply to commit w/ amend
- [x] the failing test passes now :)

kaste:
Hi @asfaltboy! 

You should be optimistic here. Like nobody amends its first commit. So try with what we usually do and fallback and repair only if the git command fails.

The parent of the inital commit is `4b825dc642cb6eb9a060e54bf8d69288fbee4904`. So you can use 

```
git diff 4b825dc642cb6eb9a060e54bf8d69288fbee4904 
```

for include unstaged and 

```
git diff 4b825dc642cb6eb9a060e54bf8d69288fbee4904  --cached
```

otherwise if iirc the options we have here.

asfaltboy:
@kaste how do you get the parent of an initial commit? How is that even possible, is there a way to get that 0-state commit?

kaste:
I have it from GitGUI source: https://github.com/git/git/blob/master/git-gui/git-gui.sh#L1423-L1434

This sha is currently a constant. It is though safer to do what gitGUI does. In the console 

```
>>> git = GitCommand()
>>> git.view = view
>>> git.git('mktree', stdin='')
'4b825dc642cb6eb9a060e54bf8d69288fbee4904\n'
```

Look at the empty check -> https://github.com/git/git/blob/master/git-gui/git-gui.sh#L1431

kaste:
Basically, a factoring like this would be dope

```
try:
	diff = self.do_diff(..., parent='HEAD^')
except ...:
	parent = self.git('mktree', stdin='').strip()
	diff = self.do_diff(..., parent=parent)
```

asfaltboy:
That is awesome! I was missing that an empty tree is a hash constant, cool stuff!

Refs:
- https://f-o.org.uk/2017/empty-trees-in-git.html
- https://stackoverflow.com/a/9766506/484127

I will try to play with this in the weekend

PR #682, opened 2017-06-25
Docs: Update to new command names and add use case

asfaltboy:
As the commands were modified in commit 2cd23be, we need to update the documentation to show the correct command names.

I also took the opportunity to add the step-by-step instructions which we can refer to from CONTRIBUTING.md


stoivo:
Perfect

PR #688, opened 2017-07-06
Update install.txt

### LINKED ISSUE #686: So how does one actually use it?
This looks great but nothing in any of the documentation actually tells one how to use this package. It tells you what it can do but not how to do it.

There are no key-bindings, no menus... so I can't figure out how to even get a simple "git status" to work. Looking in the documentation all it says is "git status: status.md" whatever that means.

A little help?

stoivo:
Fix #686

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,8 @@
 # Contribute
 
+
+
+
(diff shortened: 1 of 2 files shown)


divmain:
Unfortunately, this is not true.  GitSavvy _does_ add key-bindings to Sublime.  I'll comment in the issue itself.

stoivo:
What about this @divmain ?

randy3k:
I don't think we need to address that issue at all. This paragraph may make a user more confused.

PR #694, opened 2017-07-17
compare commits related enhancements

randy3k:
This allow users to compare commits against HEAD while using `git: log` or `git: graph`.

-----

 _(required)_

- [ ]  major
- [ ]  minor
- [x]  patch
- [ ]  documentation only

### DIFF
--- a/core/commands/commit_compare.py
+++ b/core/commands/commit_compare.py
@@ -88,20 +88,30 @@ class GsCompareCommitShowDiffCommand(TextCommand, GitCommand):
     Refresh view of all commits diff between branches.
     """
 
-    def run(self, edit):
+    def run(self, edit, reverse=False):
+        self._reverse = reverse
         sublime.set_timeout_async(self.run_async)
 
     def run_async(self):
         base_commit = self.view.settings().get("git_savvy.compare_commit_view.base_commit")
         target_commit = self.view.settings().get("git_savvy.compare_commit_view.target_commit")
         file_path = self.file_path
-        self.view.window().run_command("gs_diff", {
-            "base_commit": base_commit,
-            "target_commit": target_commit,
-            "file_path": file_path,
-            "disable_stage": True,
-            "title": "DIFF: {}..{}".format(base_commit, target_commit)
-        })
+        if self._reverse:
+            self.view.window().run_command("gs_diff", {
+                "base_commit": target_commit,
+                "target_commit": base_commit,
+                "file_path": file_path,
+                "disable_stage": True,
+                "title": "DIFF: {}..{}".format(target_commit, base_commit)
+            })
+        else:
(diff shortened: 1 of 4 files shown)


core/commands/commit_compare.py
```diff
@@ -172,6 +172,11 @@ class GsCompareAgainstCommand(PanelActionMixin, WindowCommand, GitCommand):
     def run(self, target_commit=None, file_path=None, current_file=False):
         self._file_path = self.file_path if current_file else file_path
         self._target_commit = target_commit
+
+        # if there is a target, show option to compare against HEAD
+        if target_commit:
+            self.default_actions.insert(0, ["compare_against_head", "HEAD"])
```
divmain: This will mutate the static class element.

```python
class T:
    array = []
    def print(self):
        print(self.array)
    def add(self, num):
        self.array.insert(0, num)

a = T()
b = T()

a.print()
b.print()
b.add(1)
b.print()
a.print()

# []
# []
# [1]
# [1]
```

Instead, you could do something like:
```python
self.default_actions = [["compare_against_head", "HEAD]] + self.default_actions
```

randy3k: Good catch!

randy3k: Ya, you are right.

PS: It seems that there are some random failures on CIs. I am tracking the issue.

randy3k: I just note that `PanelActionMixin` has a method `update_actions`.

randy3k:
It seems a problem of Package Control

```
Package Control: Installing 1 missing dependency
Package Control: The dependency 'pyyaml' is not available
Package Control: Skipping automatic upgrade, last run at 2017-07-17 16:51:22, next run at 2017-07-17 17:51:22 or after
```

randy3k:
When I use the compare commit tools, I always get the opposite direction result. For example, when I want `git diff foo..bar`, I always end up with `git diff bar..foo`. My second commit interchanges `base_commit` and `target_commit`. Some users perhaps really what `git diff foo..bar` when they run the command. For this purpose, I have added a F key which will show a reversed diff.

randy3k:
@divmain 

I believe you have `Require branches to be up to date before merging` turned on. However, this setting is somehow conflicting with `maintainerd` since we cannot merge a PR if it is not up to date. And if we hit the [Update Branch] button, it will create a merge commit which fails `maintainerd` checks.

screen shot 2017-07-18 at 3 45 34 am

divmain:
Fixed @randy3k.  Thanks for pointing that out.  Given that GitHub now supports rebase "merging", it's a pity they don't allow the same for branch updating...

randy3k:
I also thought that it would be a rebasing merging, but it wasn't.

PR #709, opened 2017-07-19
Refactor show_log_panel

randy3k:
Now it supports on_highlight(entry) and is used in GsBlameCommand.
It also fixes a bug when canceling the panel of `Pick a new commit to blame`

-----

 _(required)_

- [ ]  major
- [ ]  minor
- [x]  patch
- [ ]  documentation only

### DIFF
--- a/common/util/view.py
+++ b/common/util/view.py
@@ -71,7 +71,7 @@ def refresh_gitsavvy(view, refresh_sidebar=False, refresh_status_bar=True):
     if view.settings().get("git_savvy.branch_commit_history_view") is not None:
         view.run_command("gs_branches_diff_commit_history_refresh")
     if view.settings().get("git_savvy.blame_view") is not None:
-        view.run_command("gs_blame_initialize_view")
+        view.run_command("gs_blame_refresh")
 
     if refresh_status_bar:
         view.run_command("gs_update_status_bar")
--- a/common/util/view.py
+++ b/common/util/view.py
@@ -68,10 +68,6 @@ def refresh_gitsavvy(view, refresh_sidebar=False, refresh_status_bar=True):
 
     if view.settings().get("git_savvy.interface") is not None:
         view.run_command("gs_interface_refresh")
-    if view.settings().get("git_savvy.branch_commit_history_view") is not None:
-        view.run_command("gs_branches_diff_commit_history_refresh")
-    if view.settings().get("git_savvy.blame_view") is not None:
-        view.run_command("gs_blame_refresh")
 
     if refresh_status_bar:
         view.run_command("gs_update_status_bar")
--- a/core/commands/blame.py
+++ b/core/commands/blame.py
@@ -8,7 +8,7 @@ from sublime_plugin import WindowCommand, TextCommand
 from ..commands import GsNavigate
 from ..git_command import GitCommand
 from ...common import util
-from ..ui_mixins.quick_panel import PanelActionMixin, LogPanel, show_log_panel
+from ..ui_mixins.quick_panel import PanelActionMixin, show_log_panel
 
 
 BlamedLine = namedtuple("BlamedLine", ("contents", "commit_hash", "orig_lineno", "final_lineno"))
@@ -60,7 +60,7 @@ class GsBlameCommand(PanelActionMixin, WindowCommand, GitCommand):
         view.set_scratch(True)
         view.set_read_only(True)
 
-        view.run_command("gs_blame_initialize_view")
+        view.run_command("gs_blame_refresh")
         view.run_command("gs_handle_vintageous")
 
     def pick_commit(self):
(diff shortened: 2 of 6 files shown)


randy3k:
@stoivo free to review? I remember you are on vacation though.

core/ui_mixins/quick_panel.py
```diff
@@ -139,24 +142,15 @@ def show_paginated_panel(items, on_done, flags=None, selected_index=None, on_hig
...
-
-def show_log_panel(entries, on_done, limit=6000, selected_index=None):
-    """
-    Display log entries in quick panel with pagination, and execute on_done(commit)
-    when item is selected. `entries` can be either a list or a generator of LogEnty.
-
-    """
-    lp = LogPanel(entries, on_done, limit=limit, selected_index=selected_index)
-    lp.show()
+    pp
```
asfaltboy: typo ?

core/ui_mixins/quick_panel.py
```diff
@@ -232,6 +239,22 @@ def on_selection(self, index):
...
+
+    """
+    lp = LogPanel(
+        entries,
+        on_done,
+        limit=limit,
+        selected_index=selected_index,
+        on_highlight=on_highlight)
+    lp.show()
+    lp
```
asfaltboy: typo again ?

core/ui_mixins/quick_panel.py
```diff
@@ -168,7 +162,7 @@ def __init__(self, items, on_done, **kwargs):
         self.on_done = on_done
         for option in ['flags', 'selected_index', 'on_highlight',
                        'limit', 'next_message', ]:
-            if option in kwargs:
+            if option in kwargs and kwargs[option]:
```
asfaltboy: why add `and kwargs[option]` ? won't this effectively skip any "falsy" value?

E.g doing: `PaginatedPanel(selected_index=0, limit=0)` would ignore both values and keep them at the default `None`.

core/ui_mixins/quick_panel.py
randy3k: I meant `return pp`. Doing julia too much earlier. In julia, the last expression always returns

core/ui_mixins/quick_panel.py
randy3k: I meant `return lp`.

core/ui_mixins/quick_panel.py
randy3k: right, it should be `kwargs[option] is not None`. If the nullness is not checked, it may overwrite the default methods, e.g `on_highlight`.

core/commands/blame.py
```diff
@@ -376,52 +378,33 @@ def show_file_at_commit(self, from_line=False):
...
 
+
 class GsBlamePickCommitCommand(TextCommand, GitCommand):
 
     def run(self, *args, commit_hash=None):
         sublime.set_timeout_async(lambda: self.run_async(self.file_path), 0)
 
     def run_async(self, file_path):
-        settings = self.view.settings()
-        settings.set("git_savvy.commit_hash_old", settings.get("git_savvy.commit_hash"))
```
stoivo: Where did this go?

randy3k: if the panel is cancelled, the commit will set to `self.commit_hash`, which is the original commit (line 407).

core/ui_mixins/quick_panel.py
randy3k: It seems that your comment has been discarded. The reason that we need to check the nullness is the usage of LogPanel. LogPanel has a default on_highlight method. If the passing argument is not checked for beingNone, the None value would override the default.

asfaltboy: Oh I think I see now.. we do this so that show_log_panel can accept `on_highlight`?

I truly found it a bit hard to follow, maybe we need to add some comments or docstrings, at least so we don't forget why it's done in the future?

common/util/view.py
```diff
@@ -68,10 +68,6 @@ def refresh_gitsavvy(view, refresh_sidebar=False, refresh_status_bar=True):
 
     if view.settings().get("git_savvy.interface") is not None:
         view.run_command("gs_interface_refresh")
-    if view.settings().get("git_savvy.branch_commit_history_view") is not None:
-        view.run_command("gs_branches_diff_commit_history_refresh")
```
randy3k: and `branch_commit_history_view` no longer exists.

randy3k:
Just want to let you guys know that I am working of a refactoring of the Blame walking and hoping to bring some new features, for example

1. syntax highlight
2. line forwarding

core/ui_mixins/quick_panel.py
randy3k: comment added.

randy3k:
@asfaltboy does it look good now?

PR #713, opened 2017-07-20
Other: show_branch_panel and show_remote_panel

### LINKED ISSUE #546: Branch prompts are not consistent
There are many commands which require a branch prompt. For example 
- `git: cherry-pick`
- `git: pull`
- `git: log` -> `Filtered by branch`
- `git: merge` etc

However, the ways that branches are displayed are not consistent. Some commands ask for a remote first then branch, some commands ask to choose `remote/branch`.  I guess it would be more consistent to replace the prompts with a function. Maybe something like,

```python
branch_prompt(on_done, selected_branch=None, ask_remote_first=True)
```

- How do you guys think? @divmain @stoivo @asfaltboy 

- A related question is: where should the function `branch_prompt` be placed? It is a UI related function but also very git related. There is no home for such functions currently. There are quite a few functions/classes in `common` which fall into this category, e.g. `common.quick_panel.show_log_panel`. 

- Similarly, there is a need for remote prompt.

- Also related to #371

randy3k:
close #546 

It is a major refactor of the branch and remote panels. The two related functions are [show_remote_panel](https://github.com/divmain/GitSavvy/blob/194ef3cbac540501e94b7c904aab35ff4c63b5f3/core/ui_mixins/quick_panel.py#L116-L116) and [show_branch_panel](https://github.com/divmain/GitSavvy/blob/194ef3cbac540501e94b7c904aab35ff4c63b5f3/core/ui_mixins/quick_panel.py#L175-L175).

The git flow related code is kept untouched as they are written in a different style with the other code. 
I tried very hard to make sure it doesn't break any functionality, though not 100% sure. 

-----

 _(required)_

- [ ]  major
- [ ]  minor
- [x]  patch
- [ ]  documentation only

### DIFF
--- a/core/commands/checkout.py
+++ b/core/commands/checkout.py
@@ -4,9 +4,12 @@ from sublime_plugin import WindowCommand
 
 from ..git_command import GitCommand
 from ...common import util
+from ..ui_mixins.quick_panel import show_branch_panel
 
 
 NEW_BRANCH_PROMPT = "Branch name:"
+NEW_BRANCH_INVALID = "`{}` is a invalid branch name.\nRead more on $(man git-check-ref-format)"
+_is_valid_branch_name = re.compile(r"^(?!\.|.*\.\..*|.*@.*|\/)[a-zA-Z0-9\-\_\/\.\u263a-\U0001f645]+(?<!\.lock)(?<!\/)(?<!\.)$")
 
 
 class GsCheckoutBranchCommand(WindowCommand, GitCommand):
@@ -16,32 +19,24 @@ class GsCheckoutBranchCommand(WindowCommand, GitCommand):
     user selected.
     """
 
-    def run(self):
-        sublime.set_timeout_async(self.run_async, 0)
+    def run(self, branch=None):
+        sublime.set_timeout_async(lambda: self.run_async(branch), 0)
 
-    def run_async(self):
-        stdout = self.git("branch", "--no-color")
-        branch_entries = (branch.strip() for branch in stdout.split("\n") if branch)
-
-        # The line with the active branch will begin with an asterisk.
-        self.local_inactive_branches = [branch for branch in branch_entries if not branch[0] == "*"]
-
-        if not self.local_inactive_branches:
-            self.window.show_quick_panel(["There are no branches available."], None)
+    def run_async(self, branch):
+        if branch:
+            self.on_branch_selection(branch)
         else:
-            self.window.show_quick_panel(
-                self.local_inactive_branches,
-                self.on_selection,
-                flags=sublime.MONOSPACE_FONT
-                )
+            show_branch_panel(
+                self.on_branch_selection,
+                local_branches_only=True,
+                ignore_current_branch=True)
 
-    def on_selection(self, branch_name_index):
-        if branch_name_index == -1:
+    def on_branch_selection(self, branch):
+        if not branch:
             return
 
-        branch_name = self.local_inactive_branches[branch_name_index]
-        self.git("checkout", branch_name)
-        sublime.status_message("Checked out `{}` branch.".format(branch_name))
+        self.git("checkout", branch)
+        sublime.status_message("Checked out `{}` branch.".format(branch))
         util.view.refresh_gitsavvy(self.window.active_view(), refresh_sidebar=True)
 
 
@@ -51,21 +46,24 @@ class GsCheckoutNewBranchCommand(WindowCommand, GitCommand):
     Prompt the user for a new branch name, create it, and check it out.
     """
 
-    def run(self):
-        sublime.set_timeout_async(self.run_async)
+    def run(self, base_branch=None):
+        sublime.set_timeout_async(lambda: self.run_async(base_branch))
 
-    def run_async(self, name=""):
-        self.window.show_input_panel(NEW_BRANCH_PROMPT, name, self.on_done, None, None)
+    def run_async(self, base_branch=None):
+        self.base_branch = base_branch
+        self.window.show_input_panel(NEW_BRANCH_PROMPT, base_branch or "", self.on_done, None, None)
 
     def on_done(self, branch_name):
-        pattern = r"^(?!\.|.*\.\..*|.*@.*|\/)[a-zA-Z0-9\-\_\/\.\u263a-\U0001f645]+(?<!\.lock)(?<!\/)(?<!\.)$"
-        match = re.match(pattern, branch_name)
+        match = _is_valid_branch_name.match(branch_name)
         if not match:
-            sublime.error_message("`{}` is a invalid branch name.\nRead more on $(man git-check-ref-format)".format(branch_name))
+            sublime.error_message(NEW_BRANCH_INVALID.format(branch_name))
             sublime.set_timeout_async(self.run_async(branch_name))
             return None
 
-        self.git("checkout", "-b", branch_name)
+        self.git(
+            "checkout", "-b",
+            branch_name,
+            self.base_branch if self.base_branch else None)
         sublime.status_message("Created and checked out `{}` branch.".format(branch_name))
         util.view.refresh_gitsavvy(self.window.active_view())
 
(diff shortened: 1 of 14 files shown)


randy3k:
So `maintainerd` will also be triggered if the PR is merging to branches other than `master`?

divmain:
`maintainerd` is triggered, but there won't be an auto-release unless the PR is being merged into `master`.  So no issues there.  If you want, we can aggregate changes into a `develop` branch, and then do a single PR to merge all of those changes into `master` at a later time.

What are your primary concerns regarding auto tagging @randy3k?  In my mind, it seems like if we're confident enough to merge something into master, in most cases we should be confident enough to release.  And even refactors are code changes and can introduce bugs, so a version change is "correct".  It also makes sense from a practical standpoint, since a release that includes a refactor+intentional-behavior-change can obscure any regressions or new bugs introduced with the refactor itself.

So that's my thinking, but I'm curious if there are holes in that thinking.  I also want it to be _useful_ - so if it is actively interfering, we should definitely address that somehow.

randy3k:
> `maintainerd` is triggered, but there won't be an auto-release unless the PR is being merged into `master`.  So no issues there.  If you want, we can aggregate changes into a `develop` branch, and then do a single PR to merge all of those changes into `master` at a later time.

Then it should not ask for a semver type. Not choosing one would result in a failure check.


> What are your primary concerns regarding auto tagging @randy3k? 

Auto tagging is not a issue per se. I am worried about the auto-update of package control. Sublime Text users will obtain a very new piece of GitSavvy once a tag is released. It may not be idea in our case. It is not guaranteed that a hot fix in one location won't break some code in other locations. In the end, we haven't fully implemented the tests. This concern will be reduced only by enough tests.
 
> It also makes sense from a practical standpoint, since a release that includes 1) a refactor, and 2) an intentional behavior-change, can obscure any regressions or new bugs introduced with refactor itself.

In terms of the git flow terminology, it means the `master` branch is now dedicated to hot fixes and releases. The idea of using a "develop" branch and pushing fixes/releases to `master` for auto-tagging is not much different to using a "master" branch for development and do manually tagging. The "merge to master" part still has to be done manually.

With that said, I have no objection of either models. For now, I would just create PR to `develop` unless it is a simple and clear fix.

stoivo:
This is a lot of changes, is it possible to split it up into smaller commits? That would make it somewhat easier to review.

randy3k:
Sure.

randy3k:
@stoivo 
Please review.

core/ui_mixins/quick_panel.py
```diff
@@ -112,6 +113,182 @@ def get_arguments(self, selected_action):
...
+        )
+
+    def get_pre_selected_branch_index(self, remote):
+        pre_selected_index = None
+        if self.selected_branch is None:
+            selected_branch = self.get_current_branch_name()
+        else:
+            selected_branch = self.selected_branch
+
+        if pre_selected_index is None:
```
stoivo: This will always be None since it is set to None on line 260.

randy3k: You are right, it is redundant.

core/ui_mixins/quick_panel.py
```diff
@@ -112,6 +113,182 @@ def get_arguments(self, selected_action):
...
+            self.all_branches,
+            self.on_branch_selection,
+            flags=sublime.MONOSPACE_FONT,
+            selected_index=self.get_pre_selected_branch_index(remote)
+        )
+
+    def get_pre_selected_branch_index(self, remote):
+        pre_selected_index = None
+        if self.selected_branch is None:
+            selected_branch = self.get_current_branch_name()
```
stoivo: Can't we set it to `self.selected_branch` to avoid `self.selected_branch` and `selected_branch`?

core/commands/log_graph.py
```diff
@@ -129,26 +131,13 @@ def get_graph_args(self):
...
-        else:
-            pre_selected_index = self.all_branches.index(self.get_current_branch_name())
-
-        self.window.show_quick_panel(
-            self.all_branches,
-            self.on_branch_selection,
-            flags=sublime.MONOSPACE_FONT,
-            selected_index=pre_selected_index
-        )
+    def run_async(self, **kwargs):
```
stoivo: Why do we need to add `**kwargs`?

core/commands/checkout.py
```diff
@@ -51,21 +45,25 @@ class GsCheckoutNewBranchCommand(WindowCommand, GitCommand):
...
-    def run(self):
-        sublime.set_timeout_async(self.run_async)
+    def run(self, base_branch=None):
+        sublime.set_timeout_async(lambda: self.run_async(base_branch))
 
-    def run_async(self, name=""):
-        self.window.show_input_panel(NEW_BRANCH_PROMPT, name, self.on_done, None, None)
+    def run_async(self, base_branch=None):
+        self.base_branch = base_branch
+        self.window.show_input_panel(NEW_BRANCH_PROMPT, "", self.on_done, None, None)
```
stoivo: The second argument should not an empty string. When I created this it worked like if you enter an invalid branch name, for example with a space in the middle, it would open the prompt again with the name you enter prefilled.

core/commands/checkout.py
```diff
@@ -51,21 +45,25 @@ class GsCheckoutNewBranchCommand(WindowCommand, GitCommand):
...
 
     def on_done(self, branch_name):
         pattern = r"^(?!\.|.*\.\..*|.*@.*|\/)[a-zA-Z0-9\-\_\/\.\u263a-\U0001f645]+(?<!\.lock)(?<!\/)(?<!\.)$"
         match = re.match(pattern, branch_name)
         if not match:
-            sublime.error_message("`{}` is a invalid branch name.\nRead more on $(man git-check-ref-format)".format(branch_name))
+            sublime.error_message(NEW_BRANCH_INVALID.format(branch_name))
             sublime.set_timeout_async(self.run_async(branch_name))
```
stoivo: Here is the recursion if you enter an invalid branch name.

core/commands/checkout.py
```diff
@@ -78,25 +76,42 @@ class GsCheckoutRemoteBranchCommand(WindowCommand, GitCommand):
...
+        v = self.window.show_input_panel(
+                NEW_BRANCH_PROMPT,
+                local_name,
+                self.on_enter_local_name,
+                None,
+                None)
+        v.run_command("select_all")
+
+    def on_enter_local_name(self, branch_name):
+        pattern = r"^(?!\.|.*\.\..*|.*@.*|\/)[a-zA-Z0-9\-\_\/\.\u263a-\U0001f645]+(?<!\.lock)(?<!\/)(?<!\.)$"
```
stoivo: Should be moved into a constant to remove duplication.

core/commands/checkout.py
```diff
@@ -78,25 +76,42 @@ class GsCheckoutRemoteBranchCommand(WindowCommand, GitCommand):
...
         local_name = remote_branch.split("/", 1)[1]
-        self.git("checkout", "-b", local_name, "--track", remote_branch)
-        sublime.status_message("Checked out `{}` as local branch `{}`.".format(remote_branch, local_name))
+        v = self.window.show_input_panel(
+                NEW_BRANCH_PROMPT,
+                local_name,
+                self.on_enter_local_name,
+                None,
+                None)
+        v.run_command("select_all")
```
stoivo: This should be done for the the above one too

core/commands/log_graph.py
randy3k: it is not needed, will be removed.

core/ui_mixins/quick_panel.py
randy3k: yes~

core/commands/commit_compare.py
```diff
@@ -146,35 +146,22 @@ def run(self, base_commit=None, target_commit=None, file_path=None):
         self._target_commit = target_commit
         sublime.set_timeout_async(self.run_async)
 
-    def run_async(self):
-        self.all_branches = [b.name_with_remote for b in self.get_branches()]
+    def run_async(self, **kwargs):
```
randy3k: must be copied from other place.

github/commands/open_on_remote.py
```diff
@@ -38,27 +39,12 @@ def run_async(self, remote, preselect, fpath):
         if remote:
             self.open_file_on_remote(remote)
         else:
-            try:
-                pre_selected_index = self.remote_keys.index(self.last_remote_used)
-            except ValueError:
-                pre_selected_index = 0
-
-            self.view.window().show_quick_panel(
-                self.remote_keys,
```
stoivo: Remove line https://github.com/divmain/GitSavvy/pull/713/commits/63229608a3ce90b342c9c407e557c10351b603ea#diff-7307dc85c05c5908333cc4d57b32b2d8R34 unused variable
https://github.com/divmain/GitSavvy/pull/713/commits/63229608a3ce90b342c9c407e557c10351b603ea#diff-7307dc85c05c5908333cc4d57b32b2d8R91

core/commands/checkout.py
```diff
@@ -51,21 +46,24 @@ class GsCheckoutNewBranchCommand(WindowCommand, GitCommand):
...
-            sublime.error_message("`{}` is a invalid branch name.\nRead more on $(man git-check-ref-format)".format(branch_name))
+            sublime.error_message(NEW_BRANCH_INVALID.format(branch_name))
             sublime.set_timeout_async(self.run_async(branch_name))
             return None
 
-        self.git("checkout", "-b", branch_name)
+        self.git(
+            "checkout", "-b",
+            branch_name,
+            self.base_branch if self.base_branch else None)
```
stoivo: Why do we need to pass a starting point? 
I don't think we need to pass a starting point, if so self.base_branch is not necessary.

randy3k: You mean `base_branch`? It will be used in dashboard to create new branch for other branches.

github/commands/open_on_remote.py
randy3k: done

randy3k:
all rebased.

core/commands/checkout.py
stoivo: Now the command would look like 
```
git checkout -b <branch_nane> <starting point>
```
Should we pass the starting point?

randy3k: It is useful to specific the starting point in some cases. The default value of `base_branch` is None. It would not produce any side effects in the normal use case.

core/commands/checkout.py
```diff
@@ -51,21 +46,24 @@ class GsCheckoutNewBranchCommand(WindowCommand, GitCommand):
...
 
-    def run(self):
-        sublime.set_timeout_async(self.run_async)
+    def run(self, base_branch=None):
+        sublime.set_timeout_async(lambda: self.run_async(base_branch))
 
-    def run_async(self, name=""):
-        self.window.show_input_panel(NEW_BRANCH_PROMPT, name, self.on_done, None, None)
+    def run_async(self, base_branch=None):
+        self.base_branch = base_branch
```
stoivo: You are setting it here so it does not work if you change the name.

randy3k: I am lost. If `base_branch` is None, it is just like previously. Else if `base_branch` is not None, it will checkout a new branch from `base_branch`.

stoivo: hm, I thought I broke it. But when I test it now I works perfectly

PR #717, opened 2017-07-22
[WIP] refactoring blame walker

randy3k:
~~depends on #709 , if that is merge, this will be rebased.~~ # 709 has merged.

It is a refactoring of the GsBlameCommand, specifically it brings these new features.

1. syntax highlight

2. line forwarding (the cursor will follow via history instead of going to a random place)

3. use [m] to switch between `detect within file`, `detect within commit` and `detect within all commits` instead of showing a panel to ask every time. The origin popup is also modified to show less text.

4.  the options of the BlameActionCommand are renamed. Actually, still not sure the best to name them.

untitled

-----

 _(required)_
- [ ] My commit messages are cleaned up and ready to merge. _(required)_

- [ ]  major
- [ ]  minor
- [x]  patch
- [ ]  documentation only

### DIFF
--- a/core/commands/blame.py
+++ b/core/commands/blame.py
@@ -1,6 +1,7 @@
 import re
 from collections import namedtuple, defaultdict
 import unicodedata
+import os
 
 import sublime
 from sublime_plugin import WindowCommand, TextCommand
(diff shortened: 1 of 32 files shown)


randy3k:
I tried it on a large repo of 30k commits. These are significant delay (~5s - 7s) from the quick panel to the highlighting. I did some profiling, 50% of time was spent in [`scope_name`](https://github.com/divmain/GitSavvy/blob/25617ff0252153a1d3d432b9ba86d1e6678eecef/core/commands/blame.py#L409-L409) and 25% of time was spent in looking through history to [detect file name change](https://github.com/divmain/GitSavvy/blob/25617ff0252153a1d3d432b9ba86d1e6678eecef/core/git_mixins/history.py#L148-L148).

mtomov:
That looks great! Thank you!

randy3k:
The current implementation is quite slow for large files. I guess we might have to disable syntax highlight for large files.

asfaltboy:
What are these `detect within X`  options, can you elaborate? Personally, I would also love to have an option of traversal jumping to last commit mentioned in the blame (where we changed the given hunk), rather than HEAD -/+ 1 ... But we can leave it for another change

core/commands/blame.py
```diff
@@ -238,147 +320,175 @@ def couple_partitions_and_commits(partitions, commit_infos, left_pad):
...
+            "syntax", self.view.settings().get('git_savvy.original_syntax', None))
+
+        blame_view_pt = 0  # a start pointer to search for line number
+        flags = sublime.DRAW_NO_OUTLINE | sublime.PERSISTENT
+
+        count = 0
+        for lineno, line in enumerate(output_view.lines(sublime.Region(0, output_view.size()))):
+            if not line.empty():
+                pattern = r".{{30}} \| {lineno: >4}\s".format(lineno=lineno+1)
+                corresponding_region = self.view.find(pattern, blame_view_pt)
```
asfaltboy: I think blame is  designed to always output the same column widths (commit info | code) for a single blame. So there's no need to  refetch this every time.

core/commands/blame.py
```diff
@@ -238,147 +320,175 @@ def couple_partitions_and_commits(partitions, commit_infos, left_pad):
...
+                blame_view_pt = corresponding_region.end()
+
+                offset = blame_view_pt - line.begin()
+                pt = line.begin()
+                last_scope = output_view.scope_name(pt)
+                last_pt = pt
+                end_pt = line.end()
+                while pt <= end_pt:
+                    pt += 1
+                    scope = output_view.scope_name(pt)
```
asfaltboy: Do we really expect the scope to change for each char and each line?

I may be misunderstanding how this works; but, can we instead only test a couple of lines across the file to be sure? Or, even use the scope detected for the original file (outside of blame)?

core/commands/blame.py
randy3k: This PR is still working in progress, I am still testing and profiling the code.

asfaltboy: Oh sorry, feel free to ignore review then 😄

core/commands/blame.py
randy3k: it is scanning the scopes of the syntax and then applying the scopes to the blame view. I don't really know if there are any fast ways to captures all the scopes of a file.

core/commands/blame.py
randy3k: No problem. Btw, come back to your questions, we although know which column the code starts, but the line numbers of the original code gives on information of the line numbers of the corresponding lines for the blame view. For example, line 3 of the original file is line 5 of the blame view.

screen shot 2017-08-10 at 2 53 09 am

randy3k:
the `detect_with_xxx` options are the `-M` ,`-C` and `-CCC` options of the blame view: http://jfire.io/blog/2012/03/07/code-archaeology-with-git/

randy3k:
@stoivo 
I found that the `--follow` option of `git log` is very slow for large repos (> 30000 commits). We may need to address this.

randy3k:
It is hard to merge this. I am going to split this PR.

Issue #716, opened 2017-07-21
Unable to commit changes to via the plugin anymore


harshal-gajare:
I am unable to commit changes via the plugin anymore. I am able to Stage, unstage changes , change branches but when I try to commit the changes I get an error " Unable to determine Git repo path."

harshal-gajare:
```
[
  {
    "type": "git", 
    "secounds": 0.043879032135009766, 
    "stderr": "", 
    "command": [
      "status", 
      "--porcelain", 
      "-z"
    ], 
    "stdin": null, 
    "stdout": "M  src/approvalProcesses/Order_Index__c.Sample_Order_Single_Approver_Level_US.approvalProcess\u0000"
  }, 
  {
    "type": "git", 
    "secounds": 0.026094913482666016, 
    "stderr": "", 
    "command": [
      "status", 
      "-b", 
      "--porcelain"
    ], 
    "stdin": null, 
    "stdout": "## fix-inactive-user-approval-process...origin/fix-inactive-user-approval-process\nM  src/approvalProcesses/Order_Index__c.Sample_Order_Single_Approver_Level_US.approvalProcess\n"
  }, 
  {
    "type": "git", 
    "secounds": 0.010597944259643555, 
    "stderr": "", 
    "command": [
      "log", 
      "-n 1", 
      "--pretty=format:%h %s", 
      "--abbrev-commit"
    ], 
    "stdin": null, 
    "stdout": "0028a14c Merged in SFDC-2089-build-dior-email-page4 (pull request #899)"
  }, 
  {
    "type": "git", 
    "secounds": 0.06950616836547852, 
    "stderr": "", 
    "command": [
      "stash", 
      "list"
    ], 
    "stdin": null, 
    "stdout": ""
  }, 
  {
    "type": "git", 
    "secounds": 0.048963069915771484, 
    "stderr": "", 
    "command": [
      "status", 
      "--porcelain", 
      "-z"
    ], 
    "stdin": null, 
    "stdout": "M  src/approvalProcesses/Order_Index__c.Sample_Order_Single_Approver_Level_US.approvalProcess\u0000"
  }, 
  {
    "type": "git", 
    "secounds": 0.030784130096435547, 
    "stderr": "", 
    "command": [
      "status", 
      "-b", 
      "--porcelain"
    ], 
    "stdin": null, 
[...]

Issue #105, opened 2015-03-19
[WinError 267] The directory name is invalid


p3lim:
Whenever I open up or select a tab with content that should exist on the harddrive but doesn't (file opened from a Zip file so it's in /Temp, or simply opening the Default settings/keybindings for sublime), I will get the error `[WinError 267] The directory name is invalid`, marking the file as dirty.

Also, if I have any of those files opened in Sublime, close Sublime and reopen it, the tab is gone.

I will try and see if I can fix it myself.

p3lim:
Full error:
```
Unable to open /C/Users/Administrator/AppData/Roaming/Sublime Text 3/Packages/Default/Preferences.sublime-settings
Traceback (most recent call last):
  File "./subprocess.py", line 1112, in _execute_child
NotADirectoryError: [WinError 267] The directory name is invalid

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\Administrator\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\git_command.py", line 98, in git
    startupinfo=startupinfo)
  File "./subprocess.py", line 824, in __init__
  File "./subprocess.py", line 1118, in _execute_child
NotADirectoryError: [WinError 267] The directory name is invalid

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\Administrator\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\commands\status_bar.py", line 34, in run_async
    if not self.file_path or not self._repo_path(throw_on_stderr=False):
  File "C:\Users\Administrator\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\git_command.py", line 171, in _repo_path
    throw_on_stderr=throw_on_stderr
  File "C:\Users\Administrator\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\git_command.py", line 103, in git
    raise_error(e)
  File "C:\Users\Administrator\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\git_command.py", line 84, in raise_error
    raise GitSavvyError(msg)
GitSavvy.core.git_command.GitSavvyError: [WinError 267] The directory name is invalid
Traceback (most recent call last):
  File "./subprocess.py", line 1112, in _execute_child
NotADirectoryError: [WinError 267] The directory name is invalid

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #719, opened 2017-07-24
Fix: find_git_root handles multiple folders

### LINKED ISSUE #718: Odd behaviour when trying to find current working directory using git status
Hey,

I have recently done a fresh install of GitSavvy and noticed a strange behaviour change from my previous installation (work laptop).

Assuming the following project setup `test-project.sublime-project`:

```json
{
	"folders":
	[

		{
			"path": "frontend",
			"name": "A generic frontend"
		},
		{
			"path": "api",
			"name": "An API"
		},
		{
			"path": "gateway",
			"name": "A gateway"
		},
		{
			"path": "service",
			"name": "A service"
		}
	]
}
```

_Note_: All of these folders are initialized as git repos separately.

Using `git status` from an arbitrary file in one of these directories will always show the git status for the project `frontend`.

I did some digging in the code and came across https://github.com/divmain/GitSavvy/blob/c2d42fb8b7799a109cc00ad35215f87ae9668326/core/git_command.py#L215, which recently got refactored in #712. 

`find_working_dir` will _always_ prefer returning the first dir in `open_folders`, which explains the behaviour to me, but I gotta ask: 

Was this intended? It's a huge bummer to me when it comes to UX, as i expect `git status` to open the status page for the git repo I am currently in, i.e. the file I have currently open. Additionally, I cannot switch quickly between the individual repos anymore, which I greatly enjoyed beforehand.

Anyway, if this was intended, please close the issue.

randy3k:
Fix #718 and the hope to also address #716 

The fix in #712 doesn't handle multiple folders. Git root is now detected in the following logic

1. if the current file is in a repo, use that repo
2. if sidebar exists and the first folder is in a repo, use that repo
3. if sidebar exists, show the initialization popup
4. else, show "Unable to determine Git repo path."

-----

 _(required)_

- [ ]  major
- [ ]  minor
- [x]  patch
- [ ]  documentation only

### DIFF
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -230,6 +230,32 @@ class GitCommand(StatusMixin,
 
         return None
 
+    def find_repo_path(self, throw_on_stderr=False):
+        """
+        Similar to find_working_dir, except that it does not stop on the first
+        directory found, rather on the first git repository found.
+        """
+        view = self.window.active_view() if hasattr(self, "window") else self.view
+        file_name = view.file_name()
+        repo_path = None
+
+        # try the current file first
+        if file_name:
+            file_dir = os.path.dirname(file_name)
+            if os.path.isdir(file_dir):
+                repo_path = self.find_git_toplevel(file_dir, throw_on_stderr=False)
+
+        # fallback: use the first folder if the current file is not inside a git repo
+        if not repo_path:
+            window = view.window()
+            if window:
+                folders = window.folders()
+                if folders and os.path.isdir(folders[0]):
+                    # we don't set "git_savvy.repo_path" for a out-of-git-repo file
+                    repo_path = self.find_git_toplevel(
+                        folders[0], throw_on_stderr=throw_on_stderr)
+        return repo_path
+
     def find_git_toplevel(self, folder, throw_on_stderr):
         stdout = self.git(
             "rev-parse",
@@ -263,35 +289,17 @@ class GitCommand(StatusMixin,
         # The below condition will be true if run from a WindowCommand and false
         # from a TextCommand.
         view = self.window.active_view() if hasattr(self, "window") else self.view
-        repo_path = view.settings().get("git_savvy.repo_path")
+        repo_path = view.settings().get("git_savvy.repo_path") if view else None
 
         if not repo_path or not os.path.exists(repo_path):
-            file_name = view.file_name()
-
-            # try the current file first
-            if file_name:
-                file_dir = os.path.dirname(file_name)
-                if os.path.isdir(file_dir):
-                    repo_path = self.find_git_toplevel(file_dir, throw_on_stderr=False)
-                    # only set `git_savvy.repo_path` for current file
-                    if repo_path:
-                        view.settings().set("git_savvy.repo_path", os.path.realpath(repo_path))
-
-            # fallback: use the first folder if the current file is not inside a git repo
-            if not repo_path:
-                window = view.window()
-                if window:
-                    folders = window.folders()
-                    if folders and os.path.isdir(folders[0]):
-                        # we don't set "git_savvy.repo_path" for a out-of-git-repo file
-                        repo_path = self.find_git_toplevel(
-                            folders[0], throw_on_stderr=throw_on_stderr)
-
+            repo_path = self.find_repo_path(throw_on_stderr=throw_on_stderr)
             if not repo_path:
                 if throw_on_stderr:
                     raise ValueError("Unable to determine Git repo path.")
                 else:
                     return None
+            else:
+                view.settings().set("git_savvy.repo_path", os.path.realpath(repo_path))
 
         return os.path.realpath(repo_path) if repo_path else repo_path
 
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -247,7 +247,7 @@ class GitCommand(StatusMixin,
 
         # fallback: use the first folder if the current file is not inside a git repo
         if not repo_path:
-            window = view.window()
+            window = sublime.active_window()
             if window:
                 folders = window.folders()
                 if folders and os.path.isdir(folders[0]):
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -220,10 +220,12 @@ class GitCommand(StatusMixin,
 
         file_path = view.file_name()
         if file_path:
-            return os.path.dirname(file_path)
+            file_dir = os.path.dirname(file_path)
+            if os.path.isdir(file_dir):
+                return os.path.dirname(file_path)
 
         open_folders = view.window().folders()
-        if open_folders:
+        if open_folders and os.path.isdir(open_folders[0]):
             return open_folders[0]
 
         return None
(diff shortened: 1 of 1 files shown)


randy3k:
Anyone free to review this? @stoivo @asfaltboy @divmain 

I think it is affect quite a number of users.

core/git_command.py
```diff
@@ -245,38 +254,40 @@ def _repo_path(self, throw_on_stderr=True):
...
-                raise ValueError("Unable to determine Git repo path.")
-            return None
 
+        Do not raise error when throw_on_stderr is `False`. It is needed
+        in GsUpdateStatusBarCommand, otherwise, spurious popup will be shown.
+        """
         # The below condition will be true if run from a WindowCommand and false
         # from a TextCommand.
         view = self.window.active_view() if hasattr(self, "window") else self.view
+        window = view.window()
```
asfaltboy: Can be moved closer to 280 (`if window` since it doesn't seem to be used elsewhere).

core/git_command.py
```diff
@@ -245,38 +254,40 @@ def _repo_path(self, throw_on_stderr=True):
...
+                    view.settings().set("git_savvy.repo_path", os.path.realpath(repo_path))
 
+            # fallback: use the first folder if the current file is not inside a git repo
             if not repo_path:
-                return invalid_repo()
+                if window:
+                    folders = window.folders()
+                    if folders:
+                        repo_path = self.find_git_toplevel(
+                            folders[0], throw_on_stderr=throw_on_stderr)
```
asfaltboy: The above 2 outer if clauses seem to be doing the same as done in `find_working_dir`.

It can be cleaner to add a view/window parameter (if required) to that method, instead of repeating the functionality here, what do you think ?

asfaltboy: Also, if we get `repo_path` here, why should we not call `view.settings().set("git_savvy.repo_path")` as above? And if we should, why don't we call it below in an else clause to `if not repo_path` (after line 290)

core/git_command.py
```diff
@@ -245,38 +254,40 @@ def _repo_path(self, throw_on_stderr=True):
...
 
-            if view.file_name() and working_dir in view.file_name():
-                # only set repo_path for file in the working directory
-                # it allows a file living outside the repo to be edited in the current window
-                view.settings().set("git_savvy.repo_path", os.path.realpath(repo_path))
+            if not repo_path:
+                if throw_on_stderr:
+                    raise ValueError("Unable to determine Git repo path.")
```
asfaltboy: I like the throw here, but I'm wondering what are the side-effects when running in a sublime window with non-git directories/files? I'd like to test this if you haven't yet.

randy3k: if the folder is not a git repo, only GsStatusBarEventListener will trigger and the `throw_on_stderr` argument will make sure it won't do anything.

If a user run a git command, e.g. `git: log`, it will report errors. Then,
- if there is an opened folder, pop up the initialization confirmation
- if there is no folder in the sidebar, it will just show a panel of `Unable to determine Git repo path`.

core/git_command.py
randy3k: `find_working_dir` has a different usage in GsClone command, so we will need a different method for finding git root.

randy3k: Image there are two windows, W1 and W2. Each of them has a folder of git repo.
Now you run `subl .bashrc` and it opens at W1

While editing the `.bashrc` file on W1, you may still use the repo of W1. But we don't want to set the `git_savvy.repo_path` to the first folder of W1 incase you want to move `.bashrc` to W2 and continue editing.

core/git_command.py
randy3k: moved

core/git_command.py
randy3k: added a comment for `repo_path`.

randy3k:
@asfaltboy 👍 ?

core/git_command.py
asfaltboy: Thank you, I now understand the logic, what I'm worried about such issues as #589.

I just checked out the PR and played around with the different scenarios, the only issue I saw is that after opening and closing files a bunch of times, occasionally the output panel popped out with the "Unable to determine Git repo path" error.

core/git_command.py
asfaltboy: I'm sorry but I don't get this. Both `find_working_dir` and the code here *first* try to find a directory of the open file and only after this fails will they look for an open folder in the window.

The call (`view.settings().set("git_savvy.repo_path", os.path.realpath(repo_path))`) sets the repo path for the **view**, **not for the window**; so, if the `.bashrc` file is currently open the setting applies to the open file buffer only, and if the `.bashrc` is in a git repo (e.g a dotfiles repo), then the git commands should run on this repo. If we close the file or move to another file in the current window, then we should try to find it again. Finally, If the `.bashrc` is moved to another window, I'm not sure what usually happens to the view, whether it's reused or reconstructed, but we should be able to handle both cases with the same logic.

Solely from a DRY code perspective, I cannot see a reason not to refactor these to reuse `find_working_dir ` since they do exactly the same.

asfaltboy:
@randy3k added some replies above. Also, the error described in [this comment][1] indicates a regression, so we may want to handle it here as well. This can be done same as in [original fixing commit][2], by asserting `os.path.isdir(os.path.dirname(file_path))` before [returning it in `find_working_dir `][3].

[1]: https://github.com/divmain/GitSavvy/issues/105#issuecomment-318066892
[2]: https://github.com/divmain/GitSavvy/commit/3098732961bc669a380011633935f9ed4600679c
[3]: https://github.com/divmain/GitSavvy/pull/719/files#diff-bed7ab550fd42a65fddcac8e9a73197cR223

core/git_command.py
randy3k: Another use of `find_working_dir` is to find a [directory][1] to initialize. In there, we do not assume that that folder is a git repo. We might end up putting a bunch of different logic into `find_working_dir`, something like `find_working_dir(find_git_root=True)` which I find is of the same complexity to separate them into two functions.


>  if the .bashrc is in a git repo (e.g a dotfiles repo), then the git commands should run on this repo

if it is in a git repo, the [first if clause][2] had already return the repo. The logic is triggered only for files out of git repo. Setting `git_savvy.repo_path` is a shortcut for files staying in a git repo.




[1]: https://github.com/divmain/GitSavvy/blob/41f1d7b2d6ceaecff3c896ed82891aff94afa4a6/core/commands/init.py#L64-L64

[2]: https://github.com/divmain/GitSavvy/blob/41f1d7b2d6ceaecff3c896ed82891aff94afa4a6/core/git_command.py#L271-L271

core/git_command.py
randy3k: #589 should be okay since the popup shows [at most one time](https://github.com/divmain/GitSavvy/blob/41f1d7b2d6ceaecff3c896ed82891aff94afa4a6/core/commands/init.py#L42-L42) per view.

randy3k:
I made some refactoring which should address [the comment](https://github.com/divmain/GitSavvy/issues/105#issuecomment-318066892) and some of your concerns.

core/git_command.py
randy3k: I cannot exactly reproduce your issue of "Unable to determine Git repo path".

asfaltboy:
@randy3k sorry I'm really blind, until now I didn't understand that the whole reason we're not calling `find_working_dir` there is because we *want* to find a git repo of dir in the case a file without git repo is open. The refactor I had in mind would simply fail to find the repo for this kind of file, which may be fine by the way.

Ahh, the refactor you made addresses my issue, but much harder to read... In that case, I do prefer we use another method instead of refactoring the existing one. Something like the following asfaltboy/GitSavvy@e3d173c078611f9c4064a0e60412d6ccc1ccff32 maybe ?

Sorry for dragging this out for nothing.

randy3k:
Though I still think that we should only set "git_savvy.repo_path" for the current file, but I would rather postpone the discussion later and merge this first.

randy3k:
Alright, using your latest commit. Maybe we should merge it first and discuss the issue of "git_savvy.repo_path" later.

asfaltboy:
Agreed, sorry for the delay, I was looking to add a quick fix to the mentioned issue. See the Gitter chat for details.

randy3k:
Thanks.

PR #721, opened 2017-07-27
Fix: avoid setting repo_path when a file is not in the repo

### LINKED ISSUE #726: Unable to determine Git repo path
I've been getting this error when I try to commit recently. 
I don't know how to reproduce, or what causes the error.
Tried to open/close sublime, with and without project. Nothing seems to help.

randy3k:
continue the discussion in #719 for not setting "git_savvy.repo_path" if the current file is not in the found repo.

close #726

-----

 _(required)_

- [ ]  major
- [ ]  minor
- [x]  patch
- [ ]  documentation only

### DIFF
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -236,12 +236,11 @@ class GitCommand(StatusMixin,
         directory found, rather on the first git repository found.
         """
         view = self.window.active_view() if hasattr(self, "window") else self.view
-        file_name = view.file_name()
         repo_path = None
 
         # try the current file first
-        if file_name:
-            file_dir = os.path.dirname(file_name)
+        if view and view.file_name():
+            file_dir = os.path.dirname(view.file_name())
             if os.path.isdir(file_dir):
                 repo_path = self.find_git_toplevel(file_dir, throw_on_stderr=False)
 
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -251,10 +251,10 @@ class GitCommand(StatusMixin,
             if window:
                 folders = window.folders()
                 if folders and os.path.isdir(folders[0]):
-                    # we don't set "git_savvy.repo_path" for a out-of-git-repo file
                     repo_path = self.find_git_toplevel(
                         folders[0], throw_on_stderr=throw_on_stderr)
-        return repo_path
+
+        return os.path.realpath(repo_path) if repo_path else None
 
     def find_git_toplevel(self, folder, throw_on_stderr):
         stdout = self.git(
@@ -298,8 +298,11 @@ class GitCommand(StatusMixin,
                     raise ValueError("Unable to determine Git repo path.")
                 else:
                     return None
-            else:
-                view.settings().set("git_savvy.repo_path", os.path.realpath(repo_path))
+            elif view:
+                file_name = view.file_name()
+                # only set "git_savvy.repo_path" when the current file is in repo_path
+                if file_name and os.path.realpath(file_name).startswith(repo_path + os.path.sep):
+                    view.settings().set("git_savvy.repo_path", repo_path)
 
         return os.path.realpath(repo_path) if repo_path else repo_path
 
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -236,6 +236,7 @@ class GitCommand(StatusMixin,
         directory found, rather on the first git repository found.
         """
         view = self.window.active_view() if hasattr(self, "window") else self.view
+        window = view.window() if view else None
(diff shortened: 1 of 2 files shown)


randy3k:
@asfaltboy I have to switch back to `view.window()` at line [249](https://github.com/divmain/GitSavvy/blob/4404a05085a3f02ef08ca24784adbe4287334b84/core/git_command.py#L249-L249) because it causes error when switching windows quickly.

Instead, I have put a check [here](https://github.com/divmain/GitSavvy/blob/4404a05085a3f02ef08ca24784adbe4287334b84/core/git_command.py#L297-L297) to prevent throwing error when the view is detected.

Hope it works.

randy3k:
@divmain Given `maintainerd` [bug](https://github.com/divmain/maintainerd/issues/1), may be`Require status checks to pass before merging` should be disabled  for now so that we can still merge the PR manually?

asfaltboy:
Agreed!

randy3k:
@asfaltboy could you test if my fix reproduces the "not found" error?

asfaltboy:
I couldn't reproduce the original "panel pops with Unable to determine repo path", but what happens now is (rarely) the exception breaks through and I see:

```
`/usr/bin/git status -b --porcelain` failed with following output:

fatal: Not a git repository (or any of the parent directories): .git
```

Alongside the dialog suggesting to "Initialize a new git repo".

Just as a quick attempt at bisect, I went back to e51c47a and could not reproduce the issue. I assume this is related to a later change in `git_command`. @randy3k you probably know better than me which is the commit that introduced this; question: is it necessary? Can we just revert it?

randy3k:
It is an attempt to allow editing a file out of a Git repo. Before these refactorings, if you do `subl ~/.bashrc`, it would prompt the initialization dialog. I think I would what is happening, most likely that the two window.active_view() gets different values when we switch views very quickly.

randy3k:
We now only have pure Julia methods for A' * B, A * B' and A' * B'. Theoretically, it should be faster to use lapack routines if bandwidths is large (simulations suggested that lapack routines are better for bandwidth >100).

Other than that, it is working.

randy3k:
The bug is more subtle than I originally thought. The commit e51c47a was not completely free of bug, it seems to work just because it takes no time to finish the call. If I add a little bit [delay](https://github.com/divmain/GitSavvy/blob/bd22c270041a79d109b7b882c89b718cf016e715/core/git_command.py#L248-L248), it will break [appveyor](https://ci.appveyor.com/project/divmain/gitsavvy).

We see a similar bug earlier because `git rev-parse --show-toplevel` takes some time to execute.

randy3k:
I guess in order to fix this issue, GsUpdateStatusBarCommand needs something like

```python
with disable_throw_on_stderr():
      self.any_git_command(.....)
```

asfaltboy:
@randy3k maybe due to it's (currently) rare occurrence, we should merge this PR and handle the error raised from `GsUpdateStatusBarCommand` (if we're certain that is the cause) in another PR ?

randy3k:
I guess I have fixed the issue at `GsUpdateStatusBarCommand` by putting a `try` block there.

randy3k:
and also by limiting `self.repo_path` to throw `ValueError` but not `GitSavvyError`.

randy3k:
@asfaltboy please feel free to merge this if you are happy with it

core/git_command.py
```diff
@@ -268,42 +270,41 @@ def find_git_toplevel(self, folder, throw_on_stderr):
...
+                # only set "git_savvy.repo_path" when the current file is in repo_path
+                if file_name and os.path.realpath(file_name).startswith(repo_path + os.path.sep):
+                    view.settings().set("git_savvy.repo_path", repo_path)
 
         return os.path.realpath(repo_path) if repo_path else repo_path
 
     @property
+    def short_repo_path(self):
+        if "HOME" in os.environ:
+            return self.repo_path.replace(os.environ["HOME"], "~")
```
asfaltboy: Is this necessary for some cases? I don't recall an  issue discussing it. Maybe, note the reason in a comment / docstring?

randy3k: I didn't add this. It was in line 274.

randy3k:
Let's merge it first and fix the exception later.

randy3k:
I have a quick look at the exception, it seems that there is no easy way to catch this if a dashboard is closed quickly. We need a way to gracefully terminate the rendering code of the dashboard.

PR #724, opened 2017-07-27
Enhancement: disable "reset line" and "reset hunk" in inline diff cached mode

randy3k:
The concept of reseting lines an hunks in cached mode is confusing. What it does now is exactly equal to unstaging lines or hunks. So, "l" and "L" (also "h" and "H") are doing the same task in the cached mode. This PR disables the "reset" keys "H" and "L" in the cached mode.

We actually also disabled "reset hunk" for the diff view, so it does also make sense to disable it for the inline diff view.

-----

 _(required)_

- [ ]  major
- [ ]  minor
- [x]  patch
- [ ]  documentation only

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -26,7 +26,8 @@
         "args": { "reset": true },
         "context": [
             { "key": "setting.command_mode", "operator": "equal", "operand": false },
-            { "key": "setting.git_savvy.inline_diff_view", "operator": "equal", "operand": true }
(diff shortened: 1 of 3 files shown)


asfaltboy:
Just to clarify, the current behavior of the keys on a cached file for H == h and l == L yes? And we want to clear the dupe, rather than adding "reset" functionality to a staged file?

randy3k:
Yes, you are right.

PR #737, opened 2017-08-15
GitHub integration API limit

### LINKED ISSUE #528: Add fork as remote does not include all forks
For large projects, like GitSavvy, only the first few forks (one page?) are shown. If you're looking for a specific fork, which is often the case, you may have trouble finding it.

I suggest keeping the list short, and adding 2 options: _"Next page"_ and _"Custom"_, to allow manually entering a `<user>/<repo>` formatted fork label.

randy3k:
Github API has a limit of 30 items per request (can be changed to 100), this PR uses `show_paginated_panel` to show all the items in various commands.

1. `github: add fork as remote`, fix #528

aug-14-2017 22-28-52

2. `github: review pull request`

aug-14-2017 22-30-27

3. contributors and issues completion

aug-14-2017 22-35-03

-----

 _(required)_

- [ ]  major
- [ ]  minor
- [x]  patch
- [ ]  documentation only

### DIFF
--- a/core/ui_mixins/quick_panel.py
+++ b/core/ui_mixins/quick_panel.py
@@ -113,7 +113,7 @@ class PanelCommandMixin(PanelActionMixin):
 
 
 def show_paginated_panel(items, on_done, flags=None, selected_index=None, on_highlight=None,
-                         limit=6000, next_message=None):
+                         limit=6000, format_item=None, next_message=None):
     """
     Display items in quick panel with pagination, and execute on_done
     when item is selected.
@@ -124,6 +124,8 @@ def show_paginated_panel(items, on_done, flags=None, selected_index=None, on_hig
     selected_index: an integer or a callable returning boolean.
                     If callable, takes either an integer or an entry.
     on_highlight: a callable, takes either an integer or an entry.
+
+    format_item: a function to format each item
 
     next_message: a message of next page, default is ">>> NEXT PAGE >>>"
 
@@ -140,6 +142,7 @@ def show_paginated_panel(items, on_done, flags=None, selected_index=None, on_hig
             selected_index=selected_index,
             on_highlight=on_highlight,
             limit=limit,
+            format_item=format_item,
             next_message=next_message)
     pp.show()
     return pp
@@ -161,9 +164,9 @@ class PaginatedPanel:
         self.item_generator = (item for item in items)
         self.on_done = on_done
         for option in ['flags', 'selected_index', 'on_highlight',
-                       'limit', 'next_message', ]:
-            # need to check the nullness of the options to avoid the default
-            # method `on_hightight` of LogPanel to be overrided.
+                       'limit', 'format_item', 'next_message', ]:
+            # need to check the nullness of the options to avoid overriding the default
+            # methods, e.g. `format_item` and `on_hightight` of LogPanel
             if option in kwargs and kwargs[option] is not None:
                 setattr(self, option, kwargs[option])
 
--- a/core/ui_mixins/quick_panel.py
+++ b/core/ui_mixins/quick_panel.py
@@ -113,21 +113,27 @@ class PanelCommandMixin(PanelActionMixin):
 
 
 def show_paginated_panel(items, on_done, flags=None, selected_index=None, on_highlight=None,
-                         limit=6000, format_item=None, next_message=None):
+                         limit=6000, format_item=None, next_message=None, status_message=None):
     """
     Display items in quick panel with pagination, and execute on_done
     when item is selected.
 
     items: can be either a list or a generator.
+
     on_done: a callback will take one argument
+
     limit: the number of items per page
+
     selected_index: an integer or a callable returning boolean.
                     If callable, takes either an integer or an entry.
+
     on_highlight: a callable, takes either an integer or an entry.
 
     format_item: a function to format each item
 
     next_message: a message of next page, default is ">>> NEXT PAGE >>>"
+
+    status_message: a message to display at statusbar while loading the entries.
 
     If the elements are tuples of the form `(value1, value2)`,
     `value1` would be displayed via quick panel and `value2` will be passed to
@@ -143,7 +149,8 @@ def show_paginated_panel(items, on_done, flags=None, selected_index=None, on_hig
             on_highlight=on_highlight,
             limit=limit,
             format_item=format_item,
-            next_message=next_message)
+            next_message=next_message,
+            status_message=status_message)
     pp.show()
     return pp
 
@@ -155,6 +162,7 @@ class PaginatedPanel:
     """
     flags = sublime.MONOSPACE_FONT | sublime.KEEP_OPEN_ON_FOCUS_LOST
     next_message = ">>> NEXT PAGE >>>"
+    status_message = None
     limit = 6000
     selected_index = None
     on_highlight = None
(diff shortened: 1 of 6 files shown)


github/github.py
```diff
@@ -140,8 +140,55 @@ def query_github(api_url_template, github_repo):
...
+    """
+
+    is_enterprise, fqdn = get_api_fqdn(github_repo)
+    base_path = "/api/v3" if is_enterprise else ""
+    path = base_path + api_url_template.format(
+        owner=github_repo.owner,
+        repo=github_repo.repo
+    )
+
+    path = path + "?per_page=100"
```
asfaltboy: To avoid hard-coding this value, 100 can be made settable

github/commands/add_fork_as_remote.py
```diff
@@ -18,37 +20,36 @@ def run(self, edit):
...
+            source = base_repo_data["source"]
+            if not parent["clone_url"] == source["clone_url"]:
+                forks.append(source)
+
+        forks = chain(forks, github.get_forks(base_remote))
+
+        show_paginated_panel(
+            forks,
+            self.on_select,
+            limit=100,
```
asfaltboy: Another hard-coded 100

github/commands/pull_request.py
```diff
@@ -31,27 +22,40 @@ class GsPullRequestCommand(TextCommand, GitCommand, git_mixins.GithubRemotesMixi
...
 
     def run_async(self):
         base_remote = github.parse_remote(self.get_integrated_remote_url())
         self.pull_requests = github.get_pull_requests(base_remote)
-        if not self.pull_requests:
+
+        pp = show_paginated_panel(
+            self.pull_requests,
+            self.on_select_pr,
+            limit=100,
```
asfaltboy: Another hardcoded 100

github/commands/commit.py
```diff
@@ -49,18 +52,35 @@ def run_async(self, nondefault_repo):
             )
 
         issues = github.get_issues(remote)
-
-        if not issues:
+        pp = show_paginated_panel(
+            issues,
+            self.on_done,
+            format_item=self.format_item,
+            limit=100,
```
asfaltboy: Another hardcoded 100

github/commands/commit.py
```diff
@@ -81,13 +81,21 @@ def run_async(self):
 
         contributors = github.get_contributors(remote)
 
-        if not contributors:
-            return
+        pp = show_paginated_panel(
+            contributors,
+            self.on_done,
+            format_item=self.format_item,
+            limit=100,
```
asfaltboy: Another hardcoded 100

core/ui_mixins/quick_panel.py
```diff
@@ -263,6 +271,12 @@ def on_selection(self, value):
         self.value = value
         self.on_done(value)
 
+    def is_empty(self):
+        return self._is_empty
```
randy3k: I was just too lazy to implement the `@property`. Additionally, I prefer `foo.is_empty()` over `foo.is_empty`, no particular reason, just personal taste.

core/ui_mixins/quick_panel.py
```diff
@@ -255,7 +255,7 @@ def _on_highlight(self, index):
     def _on_selection(self, index):
         if index == self.limit:
             self.skip = self.skip + self.limit
-            sublime.set_timeout(self.show, 10)
```
randy3k: The blocking next page would freeze the UI. Why do you prefer blocking? Implementing <kbd>esc</kbd> would be challenging.

github/github.py
randy3k: The value 100 is the maximum `per_page` value of the gihub api. I guess we should keep it as 100.

github/commands/commit.py
randy3k: of course, these should be customizable

core/ui_mixins/quick_panel.py
asfaltboy: You're right, frozen UI sucks, but I'm just imagining if a user may want to wait for the next result list, rather than be surprised by the callback when it comes back...

Though in most cases it will happen faster than an attention span loss, so let's use async.

github/github.py
asfaltboy: I just dislike hard-coding. What if the 100 value changes and we want to update it? Do we want the grep the code looking for those 100s to change? It will also make more sense if we had `GITHUB_PER_PAGE_MAX = 100` which is self explanatory.

core/ui_mixins/quick_panel.py
```diff
@@ -113,20 +113,28 @@ def get_arguments(self, selected_action):
...
     selected_index: an integer or a callable returning boolean.
                     If callable, takes either an integer or an entry.
+
     on_highlight: a callable, takes either an integer or an entry.
 
+    format_item: a function to format each item
+
     next_message: a message of next page, default is ">>> NEXT PAGE >>>"
 
+    status_message: a message to display at statusbar while loading the entries.
```
asfaltboy: How do you feel about moving this docstring to the `PaginatedPanel` class below instead, separating the class' static properties from overwritable methods to make it clear?

core/ui_mixins/quick_panel.py
```diff
@@ -140,7 +148,9 @@ def show_paginated_panel(items, on_done, flags=None, selected_index=None, on_hig
             selected_index=selected_index,
             on_highlight=on_highlight,
             limit=limit,
-            next_message=next_message)
```
asfaltboy: I think I see now what is bothering me with this part, it's the tight coupling with the caller show_paginated_panel.

I think we can make this more readable if we move this "protection from None clause" to line 137 in show_paginated_panel, since it accepts kwargs explicitly defaulting to None.

PaginatedPanel itself can instead simply set any **kwargs given to it when initializing it, which should be mentioned in a docstring.

Please let me know how it sounds to you?

randy3k: 👍 It sounds a better refactoring. I always have troubles keeping the values from being `None`.

randy3k:
I agree.  `show_paginated_panel` makes everything much easier by allowing generators.

randy3k:
@asfaltboy Please approve again. I made the hardcode valued customizable and editing the commit messages.

we should refactor `show_paginated_panel`  in another PR.

asfaltboy:
@randy3k thanks looks good, though I'd be happy with using one single const in this case.

Couple of last questions about the `github_per_page_max` setting:

1. Would the panel support `limit` smaller than 100; how will that work?
2. What happens if a larger than 100 value is provided; will github iterator still split to smaller than 100 and then combine them for user?

I also think having both the settings and the constant may confuse users, but not sure there's a side effect, and if so is there a better way to allow override without allowing users to "step over their feet".

randy3k:
1. any positive value will work. It will just cache the 100 items in memory and `limit` amount of items will be [fetched][1].

2. it will also work. `iteratively_query_github` will continually feeding [the reader][1] until `limit` is hit.

[1]: https://github.com/randy3k/GitSavvy/blob/9f013763eda9fd5506f75b77c802827de0c76f78/core/ui_mixins/quick_panel.py#L186-L186

I think we should keep [`GITHUB_PER_PAGE_MAX=100`][2] since it will minimize the number of requests.

[2]: https://github.com/randy3k/GitSavvy/blob/9f013763eda9fd5506f75b77c802827de0c76f78/github/github.py#L15-L15

PR #744, opened 2017-08-20
Feature: git stash in command panel

### LINKED ISSUE #628: Commands to manipulate the stash
There are no commands in the palette to pop and add to the stash.
Any objection to adding those?

stoivo:
fix #628

-----

 _(required)_

- [ ]  major
- [ ]  minor
- [x]  patch
- [ ]  documentation only

### DIFF
--- a/core/ui_mixins/quick_panel.py
+++ b/core/ui_mixins/quick_panel.py
@@ -502,15 +502,22 @@ class LogPanel(PaginatedPanel):
         self.on_done(commit)
 
 
-def show_stash_panel(git, on_done):
+def show_stash_panel(on_done, **kwargs):
     """
     Display stash entries in quick panel with pagination, and execute on_done(stash)
     when item is selected. `entries` can be either a list or a generator of LogEnty.
     """
 
-    lp = PaginatedPanel(
-        git.get_stashes(),
-        on_done,
-        format_item=(lambda entry: entry.id + " " + entry.description))
-    lp.show()
-    return lp
+    sp = StashPanel(on_done, **kwargs)
+    sp.show()
+    return sp
+
+
+class StashPanel(PaginatedPanel, GitCommand):
+
+    def __init__(self, on_done, **kwargs):
+        self.window = sublime.active_window()
+        super().__init__(self.get_stashes(), on_done, **kwargs)
+
+    def format_item(self, entry):
+        return (entry.id + " " + entry.description, entry.id)
--- a/core/ui_mixins/quick_panel.py
+++ b/core/ui_mixins/quick_panel.py
@@ -334,19 +334,21 @@ class PaginatedPanel:
     """
     flags = sublime.MONOSPACE_FONT | sublime.KEEP_OPEN_ON_FOCUS_LOST
     next_message = ">>> NEXT PAGE >>>"
+    empty_list_message = None
     status_message = None
     limit = 6000
     selected_index = None
     on_highlight = None
 
     def __init__(self, items, on_done, **kwargs):
-        self._is_empty = True
+        self._is_empty = False
         self._is_done = False
         self.skip = 0
         self.item_generator = (item for item in items)
         self.on_done = on_done
         for option in ['flags', 'selected_index', 'on_highlight',
-                       'limit', 'format_item', 'next_message', 'status_message']:
+                       'limit', 'format_item', 'next_message', 'status_message',
+                       'empty_list_message']:
             # need to check the nullness of the options to avoid overriding the default
             # methods, e.g. `format_item` and `on_hightight` of LogPanel
             if option in kwargs and kwargs[option] is not None:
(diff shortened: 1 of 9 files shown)


randy3k:
should we call the commands `git: stash xxxx` instead of `git: xxx stash`? Just to match the regular command line input order.

randy3k:
We should also add some docs for `git: stash`. And we forgot the docs for `git: revert` too.

core/ui_mixins/quick_panel.py
```diff
@@ -500,3 +500,30 @@ def on_selection(self, commit):
     def on_selection_async(self, commit):
         sublime.active_window().run_command("hide_panel", {"panel": "output.show_commit_info"})
         self.on_done(commit)
+
+
+def show_stash_panel(entries, on_done):
```
randy3k: Do we really need a new function to show stash if `entries` are to be provided?  `show_paginated_panel` should just work. If you really want to implement a panel for stashes. Then it should be something similar to `show_branch_panel` which only takes the `on_done` argument.

stoivo:
> should we call the commands git: stash xxxx instead of git: xxx stash? Just to match the regular command line input order.

> Edit: but it is a bit strange to have git: stash show and git: stash create...

git: xxx stash reads more natural but it is the wrong order compared to the command line. 

> We should also add some docs for git: stash. And we forgot the docs for git: revert too.

I can add docs for both tomorrow.

core/ui_mixins/quick_panel.py
stoivo: I think is it better to have this function than specifying everything each time.  Maybe StashPanel 
 https://github.com/divmain/GitSavvy/pull/744/files#diff-6809e629d1384bcdee2aa618a5425cbfR517  is an overkill. I just copied the above one and removed all we don't need.

randy3k: Then the entries should be generated in the `show` method of `StashPanel` as in the [`show_remote_panel`](https://github.com/divmain/GitSavvy/blob/cf2b6e3992a93b5bbae1d3d8b963d327949a1ea6/core/ui_mixins/quick_panel.py#L140)

stoivo:
Please review again. @randy3k

randy3k:
added a few more commits, mostly for the StashPanel.

randy3k:
I am still struggle between `git: create stash` and `git: stash create`. 

@aristidesfl since you requested it, do you have any thoughts?

randy3k:
@stoivo please free to merge it if 👌

core/ui_mixins/quick_panel.py
```diff
@@ -502,15 +502,22 @@ def on_selection_async(self, commit):
...
-        on_done,
-        format_item=(lambda entry: entry.id + " " + entry.description))
-    lp.show()
-    return lp
+    sp = StashPanel(on_done, **kwargs)
+    sp.show()
+    return sp
+
+
+class StashPanel(PaginatedPanel, GitCommand):
```
stoivo: I thought you wanted to avoid creating this class

randy3k: Oh, I meant either creating  `StashPanel` which computes stashes directly, or just use `show_paginated_panel`. But you were in between, your `StashPanel` still needs the list of stashes as an argument.

aristidesfl:
@randy3k I think it is simpler to follow the command line order. There are more benefits to it, than otherwise. Also the command line uses `git stash save`, not `git stash save`.

Have you thought about also providing `git: stash`  as an alias to `git stash save`?

randy3k:
Ya..you are right about the save option. I also agree it should be consistent with the command line. 

I don't think we should alias `git: stash`, but rather it will pop up a quickpanel of all the available suboptions.

randy3k:
Ya..you are right about the save option. I also agree they should be consistent with the command line. 

But I don't think we should alias `git: stash`, but rather it will pop up a quickpanel of all the available suboptions.

stoivo:
Also the behavur could be better when you have no stashed.

aristidesfl:
So all stash options will be available via a sub-menu? 
Stash save also supports an optional description parameter: `git stash save 'description'`.
Does the PR support that?

randy3k:
They are now available as `git: stash save`, `git stash pop` etc. You will be asked a description when you create a stash.

randy3k:
👍 ? @stoivo

stoivo:
I would like to improve the behaviour if there no stashes.

randy3k:
We could extend PaginatedPanel by adding an argument like `empty_message` along with `next_message`

stoivo:
I did not want to change the defaut bevavour. But i think that would make sense. Can you commit the changes you thought of?

core/ui_mixins/quick_panel.py
```diff
@@ -460,33 +468,31 @@ def is_done(self):
         return self._is_done
 
 
-def show_log_panel(entries, on_done, limit=6000, selected_index=None, on_highlight=None):
+def show_log_panel(entries, on_done, **kwargs):
```
stoivo: You really like `**kwargs` :D
This looks good.

stoivo:
Merge?

randy3k:
Let 's merge and test it in the dev branch.

PR #768, opened 2017-09-02
Enhancement: Add some default syntax specific settings

deathaxe:
This commit adds some syntax specific settings which intent to make the GitSavvy widgets look a bit cleaner and disable some unused features to save some resources.

-----

 _(required)_

- [ ]  major
- [ ]  minor
- [x]  patch
- [ ]  documentation only


syntax/status.sublime-settings
```diff
@@ -0,0 +1,17 @@
+{
+	"always_show_minimap_viewport": false,
+	"auto_indent": false,
+	"detect_indentation": false,
+	"draw_centered": true,
```
asfaltboy: How come you made only the status dashboard is centered?

deathaxe: I saw the commit status panel rendering everything centered, which looked a bit strange. As I did not yet use each of the panels provided, by GitSavvy, I decided to remove it from each of them. 

We may add it to others, too.

Same with word_wrap. I know word_wrap might be a good idea for commit messages, but should kept disabled in other dashboards to avoid ugly line breaks.

deathaxe:
> Can you point to some documentation on what each of these options do?

The settings are described in the default "Preferences.sublime-settings"

randy3k:
by the way, would you change the merge base to `dev` branch? thanks

syntax/blame.sublime-settings
```diff
@@ -0,0 +1,16 @@
+{
+	"always_show_minimap_viewport": false,
```
stoivo: This does seem like a good idea. We should discuss each setting. I would not like to set more and we can confirm is good, the reason for this is that the user can set this in his/her user settings.

syntax/blame.sublime-settings
```diff
@@ -0,0 +1,16 @@
...
+	"detect_indentation": false,
+	"draw_indent_guides": false,
+	"draw_white_space": "None",
+	"font_options": [],
+	"gutter": false,
+	"line_numbers": false,
+	"margin": 20,
+	"match_brackets": false,
+	"match_tags": false,
+	"scroll_past_end": false,
```
stoivo: I would not merge this. I LOVE to scroll past end. That was to do we reading the information straight in front of my eyes rather than at the bottom of the screen.

syntax/blame.sublime-settings
```diff
@@ -0,0 +1,16 @@
+{
```
stoivo: Several places in the code we have lines like this which we should remove if we go for adding these settings.

syntax/blame.sublime-settings
```diff
@@ -0,0 +1,16 @@
...
+	"always_show_minimap_viewport": false,
+	"auto_indent": false,
+	"detect_indentation": false,
+	"draw_indent_guides": false,
+	"draw_white_space": "None",
+	"font_options": [],
```
stoivo: Docs ses 
```
// Valid options are "no_bold", "no_italic", "no_antialias", "gray_antialias",
// "subpixel_antialias", "no_round" (OS X only), "gdi" (Windows only) and
// "directwrite" (Windows only)
 ```
But why do we want to abonded the user generall setting ? 
To me honest I not sure I understand the setting. Maybe I just dump.  Just explain why we want to set this?

File: syntax/blame.sublime-settings [hunk 6]
```diff
@@ -0,0 +1,16 @@
...
+	"gutter": false,
```
stoivo: Great to remove the gutter from the blame.

syntax/blame.sublime-settings
```diff
@@ -0,0 +1,16 @@
...
+	"line_numbers": false,
+	"margin": 20,
+	"match_brackets": false,
```
stoivo: why?

syntax/dashboard.sublime-settings
```diff
@@ -0,0 +1,16 @@
+{
```
stoivo: Noboy should use this syntax file. It is just a base for branch.sublime-syntax, rebase.sublime-syntax, status.sublime-syntax and tags.sublime-syntax. https://github.com/deathaxe/GitSavvy/blob/fd7ad09d1782cf8c43bc82ef87baa22e55830a7d/syntax/rebase.sublime-syntax#L9-L9

syntax/make_commit.sublime-settings
```diff
@@ -0,0 +1,17 @@
+{
+	"always_show_minimap_viewport": false,
+	"auto_indent": false,
+	"detect_indentation": false,
+	"draw_indent_guides": false,
+	"draw_white_space": "None",
+	"font_options": [],
+	"gutter": false,
+	"line_numbers": false,
```
stoivo: Line numbers are great when  I am writing a commit message since I want to see that 2. line is empty and who many lines I wrote in the body.

syntax/status.sublime-settings
stoivo: I don't like the centered position.

> I saw the commit status panel rendering everything centered

Where? Because we are not setting that property anywhere in GitSavvy.

syntax/blame.sublime-settings
deathaxe: ???? 

A settings file needs to start with `{` as it represents a JSON object. Removing it would break the settings file.

syntax/blame.sublime-settings
deathaxe: This setting makes most sense for output panel syntaxes, propably.

syntax/blame.sublime-settings
deathaxe: This setting targets "no_bold", "no_italic" to be disabled. But you are right, disabling everything might disturb.

syntax/blame.sublime-settings
deathaxe: Doesn't make sense anyway, does it?

syntax/blame.sublime-settings
deathaxe: I always hated empty looking output panels caused by accidentally scrolling past end. They are located at the bottom of the screen anyway. 

Scolling past end of normal views might be a question of taste. As I didn't check which of the syntaxes is used in display output panels or normal views, I just copied the settings.

syntax/dashboard.sublime-settings
deathaxe: Ok, so we can remove this settings file.

syntax/make_commit.sublime-settings
```diff
@@ -0,0 +1,17 @@
...
+	"margin": 20,
+	"match_brackets": false,
+	"match_tags": false,
+	"scroll_past_end": false,
+	"translate_tabs_to_spaces": false,
+	"word_wrap": true,
+	"wrap_width": 100,
```
deathaxe: How about word_wrap? Do you think it is a good idea to set a fixed value? Was not sure about it.

Maybe only enabling?

syntax/status.sublime-settings
deathaxe: The question was: Why **only** in status? I created the settings file for status panel with "draw_centered" true first and copied the settings to all other files, when I realized the "show_commit" output panel rendering everything centered looks ugly and misplaced.

deathaxe:
Added a fixup commit to address the requested changes to the settings files.

syntax/blame.sublime-settings
stoivo: That can be fixed by all ways focusing on line 1

syntax/blame.sublime-settings
stoivo: https://github.com/divmain/GitSavvy/blob/master/core/commands/blame.py#L57-L59
I lost this link. Sorry

syntax/make_commit.sublime-settings
stoivo: I am not sure. I need to think about forcing this on all users without a way to disable it.

syntax/blame.sublime-settings
deathaxe: Aaahh. Know what you mean.

deathaxe: Removed the duplicated settings from the plugins.

stoivo:
Hi, can I get push access to your repo?
I would like to push a commit

I would like to add something like this to open the setting on a each syntax spesfic format. 
```
    {
        "keys": ["alt+super+,"],
        "command": "edit_settings",
        "args": {
            "base_file": "${packages}/GitSavvy/syntax/status.sublime-settings",
            "default": "// Override settings for status syntax-specific.\n{\n\t$0\n}\n"
        },
        "context": [
            { "key": "setting.syntax", "operator": "eq", "operand": "Packages/GitSavvy/syntax/status.sublime-syntax" },
            // { "key": "setting.syntax", "operator": "regex_contains", "operand": "status\\.sublime-syntax$", "match_all": true },
        ]
    },
```
Any idea why this is not working, @randy3k, @asfaltboy, @divmain ?

stoivo:
Solved it by using a context like this.
```
       "context": [
           { "key": "selector", "operator": "equal", "operand": "git-savvy.blame" }
       ]
```

stoivo:
I don't think we should touch 
```
	"match_brackets": false,
	"match_tags": false,
```
I don't see what good then change. I think the defaults are good and if the user hates them for some reason hates them for some reason. She/he would change the setting in there user setting and no reason for us to overwrite it. 

Maybe I am missing something???

deathaxe:
> 	"match_brackets": false,
>	"match_tags": false,

I added those settings as I don't see any reason why this feature is required in one of the dashboards. It even might look a bit "stupid" or strange if one moves the cursor over a bracket, which might be part of a static "control" of a dashboard but highlighted by ST core. 

The only syntax it might be removed from, is the _make_commit.sublime-syntax_ as it is used by user actively to type some text. But I would keep it disabled for all "read-only" syntaxes/dashboards.

> ... She/he would change the setting in there user setting ...

The intend of this PR is to provide neat defaults. Sure a user can change anything in ST, but this is in fact an overwhelming circumstance. I took the time to find out how to make GitSavvy panels look a bit better and want to share the results. Those panels are the most obvious part of GitSavvy and might be responsible for the "first impression".  I think most people don't want to find out any detail of any plugin to tweak it perfectly to their needs but would like to see good defaults anyway.

Settings can still be changed by adding syntax specific settings files to user package.

stoivo:
Thanks again for your effort.

I push some code for opening the syntax specific settings with `super+,` or `ctrl+,`. 

You are right that `match_brackets` and `match_tags` should be tured of for most panels. I think we should not disable it for blame. Since you are reading code. Thoughts? 

Other than that is is all reasonable defaults.

deathaxe:
Considered disabling it for blame, too, but with a quick look on some blame dashboards I saw only pieces of code being displayed to the right. Therefore the correctly matching counter part of a bracket or tag might not always be available and cause false positives. Therefore I decided to keep it disabled.

But I wouldn't be opposed too much to enable it for blame.

stoivo:
Are there anything else we should consider before we can merge it?

randy3k:
I would believe in you two.

deathaxe:
I am fine with it.

asfaltboy:
I was going to suggest to refactor/reuse the code in each py module with some sort of mixin, but we can also do it later

deathaxe:
You can still do. Nothing is merged yet.

deathaxe:
As I saw all/most of your _*.sublime-..._ files use 4 spaces as indention, I decided to create a fixup! which changes indention of the settings files to match all your other files in the project. It also removes some commas from last items, to ensure valid JSON format.

stoivo:
I will use this branch not the next couple of days. Then merging 👍  🎉

asfaltboy:
Agreed, I have no time to review the code and anyway, we can always refactor later.

stoivo:
Since the where couple of fixup commits I rebases and merged locally. Not is is on dev branch. https://github.com/divmain/GitSavvy/commits/dev. 

Thanks @deathaxe 🎉

Issue #772, opened 2017-09-04
Can't commit: "Window does not exist"


franciscolourenco:
```
Traceback (most recent call last):
  File "core.git_command in /Users/user/Library/Application Support/Sublime Text 3/Installed Packages/GitSavvy.sublime-package", line 103, in git
  File "core.git_command in /Users/user/Library/Application Support/Sublime Text 3/Installed Packages/GitSavvy.sublime-package", line 297, in repo_path
RuntimeError: Window does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "core.commands.commit in /Users/user/Library/Application Support/Sublime Text 3/Installed Packages/GitSavvy.sublime-package", line 142, in <lambda>
  File "core.commands.commit in /Users/user/Library/Application Support/Sublime Text 3/Installed Packages/GitSavvy.sublime-package", line 162, in run_async
  File "core.git_command in /Users/user/Library/Application Support/Sublime Text 3/Installed Packages/GitSavvy.sublime-package", line 106, in git
  File "core.exceptions in /Users/user/Library/Application Support/Sublime Text 3/Installed Packages/GitSavvy.sublime-package", line 7, in __init__
TypeError: GitSavvyError does not take keyword arguments
```

stoivo:
Can you give the steps to reproduce?

aristidesfl:
@stoivo I don't know them. It seems to happen randomly.

aristidesfl:
```
>>> import shutil, sys, sublime; print("which git:", shutil.which("git")); print("os:", sys.platform); print("sublime version:", sublime.version()); print("GitSavvy git:", sublime.load_settings("GitSavvy.sublime-settings").get("git_path"))
which git: /usr/bin/git
os: darwin
sublime version: 3133
GitSavvy git: 
```

astephensen:
I'm also having this issue. It's very random, maybe 1 in 10 commits won't actually be committed. If I recommit it again it usually works fine.
```
which git: /usr/local/bin/git
os: darwin
sublime version: 3142
GitSavvy git: 
```

stoivo:
Are there some specific files that you have this issue with. Files with spaces? files with emoji in?

astephensen:
I don't think the files were that strange but I'll definitely take more notice now. I found it weird that committing the files the second time around works fine, even without making any other changes.

[...]

PR #773, opened 2017-09-07
Fix: Decode stdin to debug log

stoivo:
This bug was intoroduces in
  Other: refactor log_git and add log_ ...
  6fbefaf97720

### DIFF
--- a/common/util/debug.py
+++ b/common/util/debug.py
@@ -58,7 +58,7 @@ def log_git(command, stdin, stdout, stderr, seconds):
         seconds=seconds
         )
     for field in ['stdin', 'stdout', 'stderr']:
-        if isinstance(field, bytes):  # decode standard I/O bytes
+        if isinstance(locals()[field], bytes):  # decode standard I/O bytes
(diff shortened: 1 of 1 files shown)


stoivo:
This should be pushed to packagecontroll.io since this will fix debug log for all users using inline diff

stoivo:
Write a comment in #772 that this PR is merged.

stoivo:
Since maintenerd in gone, make a release manually?

stoivo:
We can add 
```
    "2.16.3": "messages/2.16.2.txt"
```
`2.16.3` to point to the same ad `2.16.4` if we think there are someone how was not updated yet. If we do that we will show `messages/2.16.2.txt` twice for those updating one by one.

asfaltboy:
I believe we can skip updating messages all together. Just make a small hotfix branch off master, and cherry pick this PRs merge and done

asfaltboy:
https://github.com/divmain/GitSavvy/releases/tag/2.16.3

All done, now package control should detect the tag and push the update.

PR #775, opened 2017-09-11
Enhancement: refactor blame command

randy3k:
It replaces my earlier PR #717 by removing the syntax highlight code because of performance issue.

1. line forwarding (the cursor will follow via history instead of going to a random place)

2. use [f], [c] and [a] to switch between `detect within file`, `detect within commit` and `detect within all commits` instead of showing a panel to ask every time. The popup view is also modified to show less text.

3.  the options of the BlameActionCommand are renamed.

-----

 _(required)_

- [ ]  major
- [ ]  minor
- [x]  patch
- [ ]  documentation only

### DIFF
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -215,7 +215,7 @@
     },
     {
         "caption": "git: blame current file",
-        "command": "gs_blame"
+        "command": "gs_blame_current_file"
     },
     {
         "caption": "GitSavvy: reload modules (debug)",
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -766,9 +766,27 @@
         ]
     },
     {
-        "keys": ["m"],
+        "keys": ["f"],
         "command": "gs_blame_toggle_setting",
-        "args": { "setting": "detect_move_or_copy_within" },
+        "args": { "setting": "detect_move_or_copy_within", "value": "file" },
+        "context": [
+            { "key": "setting.command_mode", "operator": "equal", "operand": false },
+            { "key": "setting.git_savvy.blame_view", "operator": "equal", "operand": true }
+        ]
+    },
+    {
+        "keys": ["c"],
+        "command": "gs_blame_toggle_setting",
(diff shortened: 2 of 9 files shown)


stoivo:
I would very much like to go through it but I can't find the time now while I am traveling.

randy3k:
No problem. Take your time

stoivo:
I rebased to resolve merge conflicts and added a commit with some additions fix. My commit should be squashed into yours. Please pull it down and take the ownership if the commit again. I am not completely sure how I took the ownership but I think it was because I did `git commit` instead of `git rebase --contintue` while rebasing.

When I went through it now you did not really change that much, mostly for the better. It is good that we moved the selection of `detect_move_or_copy_within` into an after preference since mostly you would not change it. I do think it was better when you could see all the options and select one rather than tabbing through.

Is it possible for you to push a branch with these changes https://github.com/divmain/GitSavvy/pull/717/commits?

randy3k:
Check again.

randy3k:
the `--follow` option of the `git log` could be quite slow for a large project. Do we want to use the version without `--follow` as default and create a toggle/settings for following file name changes?

(cf, we also do not follow file name change in `git: log current file`)

randy3k:
@stoivo what do you think about the --follow option?

stoivo:
Maybe `--follow` is more dangerous than I thought first. I started out only working on GitSavvy which has less than 1,5 k commits. So turn it off by default seems alright.

I think most users would like to have --follow enabled so maybe have it on by default and option to turn it off?

stoivo:
> Is it possible for you to push a branch with these changes  https://github.com/divmain/GitSavvy/pull/717/commits?

Possible?

randy3k:
I don't understand. Why is it necessary?

randy3k:
let's merge this first and discuss the --follow option later.

randy3k:
let's merge this first and discuss the --follow option later.

stoivo:
With `--follow` you have a better experience if the file name changes? maybe we can do --follow on a rescue.

PR #802, opened 2017-11-02
Add mv command.

### LINKED ISSUE #801: rm/mv commands
I'd like to be able to remove or move files with GitSavvy.  Is there an existing way to do this?  If not, would you be interested in a PR to add them?

ehuss:
Fixes #801.


core/commands/mv.py
```diff
@@ -0,0 +1,35 @@
...
+class GsMvCurrentFileCommand(WindowCommand, GitCommand):
+
+    """
+    Prompt the user for a new name for the current file.
+    """
+
+    def run(self):
+        if self.file_path:
+            parent, base_name = os.path.split(self.file_path)
+            on_done = functools.partial(self.on_done,
```
stoivo: What is functools.partial?

core/commands/mv.py
```diff
@@ -0,0 +1,35 @@
...
+            v.sel().clear()
+            v.sel().add(sublime.Region(0, len(name)))
+
+    def on_done(self, file_path, parent, base_name, new_name):
+        if new_name == base_name:
+            return
+        new_path = os.path.join(parent, new_name)
+
+        self.git("mv", file_path, new_path)
+        v = self.window.find_open_file(file_path)
```
stoivo: I did not see any documentation for `find_open_file` at https://www.sublimetext.com/docs/api-reference, where did you find this method?
How does it handle if you have multiple instances of the same file open?

core/commands/mv.py
```diff
@@ -0,0 +1,35 @@
...
+        if v:
+            v.retarget(new_path)
```
stoivo: Just use 
```python
v.run_command("select_all")
```

core/commands/mv.py
stoivo: Checked the doc.
seems like a good alternative to setting instance variables

stoivo:
Thanks for helping and improving GitSavvy.
We have a thing here, that we make all PR merge into dev, I will change it for this PR now.

core/commands/mv.py
ehuss: The Sublime 3 docs are now at https://www.sublimetext.com/docs/3/api_reference.html

core/commands/mv.py
ehuss: I'm not sure I understand.  Is this referring to the selection code above?  If so, that's not the same.  It specifically is selecting the name without the extension.

core/commands/mv.py
ehuss: If there are multiple views to the same file, it will just return one of the views.  That's fine, since the retargeting will affect the underlying buffer and all views will be updated.

stoivo: Durp about the docs, I have seen it before.

core/commands/mv.py
```diff
@@ -0,0 +1,35 @@
...
+    def run(self):
+        if self.file_path:
+            parent, base_name = os.path.split(self.file_path)
+            on_done = functools.partial(self.on_done,
+                self.file_path, parent, base_name)
+            v = self.window.show_input_panel("New Name:", base_name,
+                on_done, None, None)
+            name, ext = os.path.splitext(base_name)
+            v.sel().clear()
+            v.sel().add(sublime.Region(0, len(name)))
```
stoivo: I meant it here, but as you stated. You only select the name without the extension.

core/commands/mv.py
randy3k: I think `retarget` is perfect for this usage.

randy3k:
@stoivo there is a function `Rebase and merge`.

randy3k:
Anyway, I did it manually already  17f6fb2

stoivo:
I know that `Rebase and merge` exists. The problem was that if I did that I would get so the changes for  master into dev in this commit.

PR #803, opened 2017-11-03
set the default for max_items_in_tags_dashboard to -1

randy3k:
so that PackageDev will be able to pick up the setting
-1 is in effect of selecting all tags.


stoivo:
PackageDev, what is it that PackageDev is doing?

randy3k:
https://packagecontrol.io/packages/PackageDev
It has a lot of handy features to help package development.

PR #805, opened 2017-11-09
Fix: switch back to `open`

### LINKED ISSUE #799: "OS error" message not going away.
- Sublime Text version: 3143
- Git version number: 2.7.4
- OS type and version: Ubuntu Linux 16.04

I got stuck with the message bellow after closing and reopening Sublime for the first time after adding the package. After closing the dialogue window, the message got right back not allowing me to do anything.
`GitSavvy encountered an OS error: `
`[Errno 2] No such file or directory: '/var/www/html/[path skipped]/.git/rebase-apply/head-name'`

I had to remove GitSavvy manually from the filesystem.

Best regards.

randy3k:
fix #799, as these files are not user-created files

### DIFF
--- a/core/git_mixins/rewrite.py
+++ b/core/git_mixins/rewrite.py
@@ -209,7 +209,7 @@ class RewriteMixin():
 
     def rebase_orig_head(self):
         path = os.path.join(self._rebase_dir, "orig-head")
-        with util.file.safe_open(path, "r") as f:
+        with open(path, "r") as f:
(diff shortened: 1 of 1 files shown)


stoivo:
I just got the time to look at this myself. #806

I think we should check if the file exists instead of raising exceptions. It does not look nice in the console.

randy3k:
move the discussion to #806.

PR #806, opened 2017-11-09
WIP: "OS error" message not going away

stoivo:
issue #799

Hmm, this is not a good, though it removes the annoying message. 
This will now raise `RuntimeError: generator didn't yield` in the console.

I think we just need to add some file.exits here and there.

### DIFF
--- a/common/util/file.py
+++ b/common/util/file.py
@@ -73,6 +73,14 @@ def get_file_contents(repo_path, file_path):
         return binary.decode('latin-1')
 
 
+def safe_read(path):
+    try:
+        with open(path, 'r') as f:
+            return f.read().strip()
+    except FileNotFoundError:
+        return False
+
+
 @contextmanager
 def safe_open(filename, mode, *args, **kwargs):
     try:
--- a/common/util/file.py
+++ b/common/util/file.py
@@ -73,14 +73,6 @@ def get_file_contents(repo_path, file_path):
         return binary.decode('latin-1')
 
 
-def safe_read(path):
-    try:
-        with open(path, 'r') as f:
-            return f.read().strip()
-    except FileNotFoundError:
-        return False
-
-
 @contextmanager
 def safe_open(filename, mode, *args, **kwargs):
     try:
--- a/core/git_mixins/rewrite.py
+++ b/core/git_mixins/rewrite.py
@@ -209,26 +209,22 @@ class RewriteMixin():
 
     def rebase_orig_head(self):
         path = os.path.join(self._rebase_dir, "orig-head")
-        with util.file.safe_open(path, "r") as f:
-            return f.read().strip()
+        return util.file.safe_read(path)
 
     def rebase_conflict_at(self):
         if self.in_rebase_merge():
             path = os.path.join(self._rebase_merge_dir, "current-commit")
         else:
             path = os.path.join(self._rebase_apply_dir, "original-commit")
-        with util.file.safe_open(path, "r") as f:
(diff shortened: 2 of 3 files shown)


randy3k:
We should probably throw an error if the file is not found to stop any downstream code.

stoivo:
I think this will do the trick

common/util/file.py
```diff
@@ -73,6 +73,14 @@ def get_file_contents(repo_path, file_path):
         return binary.decode('latin-1')
 
 
+def safe_read(path):
+    try:
+        with open(path, 'r') as f:
+            return f.read().strip()
+    except FileNotFoundError:
+        return False
```
randy3k: why should it return False? Returning a `False` will break the downstream code, for example, line 245.
I still think that we should just use `open` and stop any downstream code.

stoivo: What I don't like with that is that then the cosole output look like this 
```
  File "/Users/user/Library/Application Support/Sublime Text 3/Packages/GitSavvy/common/ui.py", line 289, in run_async
    existing_interface.render(nuke_cursors=self.nuke_cursors)
  File "/Users/user/Library/Application Support/Sublime Text 3/Packages/GitSavvy/common/ui.py", line 99, in render
    rendered = self._render_template()
  File "/Users/user/Library/Application Support/Sublime Text 3/Packages/GitSavvy/common/ui.py", line 117, in _render_template
    keyed_content = self.get_keyed_content()
  File "/Users/user/Library/Application Support/Sublime Text 3/Packages/GitSavvy/common/ui.py", line 155, in get_keyed_content
    for key, render_fn in self.partials.items()
  File "./python3.3/collections/__init__.py", line 56, in __init__
  File "./python3.3/collections/abc.py", line 578, in update
  File "/Users/user/Library/Application Support/Sublime Text 3/Packages/GitSavvy/common/ui.py", line 155, in <genexpr>
    for key, render_fn in self.partials.items()
  File "/Users/user/Library/Application Support/Sublime Text 3/Packages/GitSavvy/core/interfaces/rebase.py", line 191, in render_diverged_commits
    end=self.rebase_orig_head() if self._in_rebase else "HEAD"
  File "/Users/user/Library/Application Support/Sublime Text 3/Packages/GitSavvy/core/interfaces/rebase.py", line 229, in get_diverged_commits_info
    if self._in_rebase else
  File "/Users/user/Library/Application Support/Sublime Text 3/Packages/GitSavvy/core/interfaces/rebase.py", line 234, in _get_diverged_in_rebase
    conflict_commit = self.rebase_conflict_at()
  File "/Users/user/Library/Application Support/Sublime Text 3/Packages/GitSavvy/core/git_mixins/rewrite.py", line 220, in rebase_conflict_at
    with util.file.open(path, "r") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/Users/user/Library/Application Support/Sublime Text 3/Packages/GitSavvy/.git/rebase-merge/current-commit'
```

And it's not completely correct that is would fail. At line 242 we check if the file exist.
https://github.com/divmain/GitSavvy/blob/c5bb2af86a840392045b4c02315cc48e851824f8/core/git_mixins/rewrite.py#L242-L243

randy3k:
It was not a good example, line 223 and 227 are better examples. I think a better and correct way to suppress the trace back is to throw an error and catch the error downstream but not to return a False earlier.

stoivo:
@randy3k, like this?

randy3k:
it looks much better now, thank you @stoivo

PR #808, opened 2017-11-11
Fix: handle cancellation of the panel

randy3k:
`selection` would be `None` if the quick panel was canceled.

### DIFF
--- a/core/interfaces/rebase.py
+++ b/core/interfaces/rebase.py
@@ -965,10 +965,12 @@ class GsRebaseOnTopOfCommand(GsRebaseDefineBaseRefCommand):
         ["select_ref", "Rebase to ref"],
     ]
 
-    def rebase_to_base_ref(self, selection):
+    def rebase_to_base_ref(self):
         self.set_base_ref(self.view.settings().get("git_savvy.rebase.base_ref"))
 
     def set_base_ref(self, selection):
+        if not selection:
+            return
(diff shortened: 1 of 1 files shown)


core/interfaces/rebase.py
```diff
@@ -965,10 +965,12 @@ class GsRebaseOnTopOfCommand(GsRebaseDefineBaseRefCommand):
         ["select_ref", "Rebase to ref"],
     ]
 
-    def rebase_to_base_ref(self, selection):
+    def rebase_to_base_ref(self):
```
asfaltboy: Good catch 👍 !

core/interfaces/rebase.py
```diff
@@ -965,10 +965,12 @@ class GsRebaseOnTopOfCommand(GsRebaseDefineBaseRefCommand):
...
         self.set_base_ref(self.view.settings().get("git_savvy.rebase.base_ref"))
 
     def set_base_ref(self, selection):
+        if not selection:
+            return
```
asfaltboy: I don't think `set_base_ref` is ever called directly from a panel, I couldn't reproduce the issue with selection being missing here ... can you provide reproduction step by step instructions please?

asfaltboy:
Oops, I see you already merged it and @stoivo reviewed this. My bad, if it's good to go then please ignore my comment above

PR #809, opened 2017-11-11
changelog: [IMP] add BREAKING syntax

blaggacao:
Should result in something like:

```
changelog:
 -  [IMP] add BREAKING syntax
        BREAKING: Line1 ...
                  Line 2 ...
                  Line 3 ...
```

### DIFF
--- a/core/commands/changelog.py
+++ b/core/commands/changelog.py
@@ -59,6 +59,11 @@ class GsGenerateChangeLogCommand(WindowCommand, GitCommand):
             contributors.add(entry.author)
             if entry.long_hash in ancestor:
                 messages.append("{} (Merge {})".format(entry.summary, ancestor[entry.long_hash]))
+            elif entry.raw_body.find('BREAKING:'):
+                pos_start = entry.raw_body.find('BREAKING:')
+                key_length = len('BREAKING:')
+                indented_sub_msg = ('\n\t\t' + ' ' * key_length + ' ').join(entry.raw_body[pos_start:].split('\n'))
+                messages.append("{}\n\t\t{})".format(entry.summary, indented_sub_msg))
             else:
                 messages.append(entry.summary)
 


blaggacao:
Would be really nice to have this, as it empowers better (more structured) information flow between Developers and DevOps...

randy3k:
I am not sure if I like it. Though I agree that you should make it more flexible and customizable.

blaggacao:
@randy3k There are for sure an infinite amount of better solutions. I guess, BREAKING is a widely used keyword. Probably enough for getting this through before thinking about a long term solutions?

stoivo:
@blaggacao, can you add a setting to set the text, and set the default to  `BREAKING:`?

randy3k:
I just wonder is it really make sense to look into each commit message to just search for "BREAKING"? But, I don't really have any idea on generalizing it for other discoveries.

stoivo:
I agree, I personally think that I would not use it. But I think is better to put it in the commit message then anywhere else. 

For the keyword, no thoughts.

blaggacao:
@stoivo Do you mean something which, in the commit message screen, would listen to something lig CTRL+ENTER (eg CTRL+SHIFT) and would append a default `BREAKING:` template?

stoivo:
No. What I meant was add a new setting to change the word. Like this
```diff
 GitSavvy.sublime-settings  | 9 ++++++++-
 core/commands/changelog.py | 6 ++++--
 2 files changed, 12 insertions(+), 3 deletions(-)

diff --git a/GitSavvy.sublime-settings b/GitSavvy.sublime-settings
index 16c0a14..40051fc 100755
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -339,5 +339,12 @@
         When set to `true`, no views will receive the prompt asking to initialize
         Git in the current view's directory when not found.
      */
-    "disable_git_init_prompt": false
+    "disable_git_init_prompt": false,
+
+    /*
+        When this keyword is found in a commit message `generate changelog` would pick
+        up the rest of the commit message and append it to the changelog.
+     */
+    "generate_changelog_detail_keyword": "BREAKING:"
+
 }
diff --git a/core/commands/changelog.py b/core/commands/changelog.py
index a198144..7c0b45d 100644
--- a/core/commands/changelog.py
+++ b/core/commands/changelog.py
@@ -60,8 +60,10 @@ class GsGenerateChangeLogCommand(WindowCommand, GitCommand):
             if entry.long_hash in ancestor:
                 messages.append("{} (Merge {})".format(entry.summary, ancestor[entry.long_hash]))
             elif entry.raw_body.find('BREAKING:'):
-                pos_start = entry.raw_body.find('BREAKING:')
-                key_length = len('BREAKING:')
+                savvy_settings = sublime.load_settings("GitSavvy.sublime-settings")
+                keyword = savvy_settings.get('generate_changelog_detail_keyword')
+                pos_start = entry.raw_body.find(keyword)
+                key_length = len(keyword)
                 indented_sub_msg = ('\n\t\t' + ' ' * key_length + ' ').join(entry.raw_body[pos_start:].split('\n'))
                 messages.append("{}\n\t\t{})".format(entry.summary, indented_sub_msg))
             else:
```

stoivo:
@randy3k what do you think ?

randy3k:
I don't mind merging it now, but I think we should keep it internal.

PR #810, opened 2017-11-11
Fix: ignoring file was not working

### LINKED ISSUE #781: Ignore file fails with no debug output
I want to use gitsavvy to ignore a file, so trying with both the 'i' and 'I' command on the status screen, I attempted to ignore a main.pdf file. I get the following error:
\`D:\Programs and Stuff\Git\cmd\git.EXE config --global core.autocrlf\` failed with following output:

However, no error message is printed. The full call stack can be seen in the developer console:

> Traceback (most recent call last):
>   File "core.commands.ignore in C:\Users\Raphael A. Meyer\AppData\Roaming\Sublime Text 3\Installed Packages\GitSavvy.sublime-package", line 41, in on_done
>   File "core.git_mixins.ignore in C:\Users\Raphael A. Meyer\AppData\Roaming\Sublime Text 3\Installed Packages\GitSavvy.sublime-package", line 19, in add_ignore
>   File "core.git_command in C:\Users\Raphael A. Meyer\AppData\Roaming\Sublime Text 3\Installed Packages\GitSavvy.sublime-package", line 162, in git
> GitSavvy.core.exceptions.GitSavvyError: \`D:\Programs and Stuff\Git\cmd\git.EXE config --global core.autocrlf\` failed with following output:

randy3k:
close #781

- set default ignore file as current file
- catch the error when running `config --global core.autocrlf` when the setting was not set.

### DIFF
--- a/core/commands/ignore.py
+++ b/core/commands/ignore.py
@@ -18,9 +18,9 @@ class GsIgnoreCommand(WindowCommand, GitCommand):
     at the Git repo's root.
     """
 
-    def run(self, file_path_or_pattern):
+    def run(self, file_path_or_pattern=None):
         if not file_path_or_pattern:
-            file_path_or_pattern = self.file_path_or_pattern
+            file_path_or_pattern = self.file_path
 
         self.add_ignore(os.path.join("/", file_path_or_pattern))
         sublime.status_message("Ignored file `{}`.".format(file_path_or_pattern))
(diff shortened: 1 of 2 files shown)


asfaltboy:
@randy3k how do you feel about 61c74b8 I think it's a bit easier to read?

randy3k:
👍 have forgotten the `throw_on_stderr` argument.

PR #811, opened 2017-11-12
Fix: Use correct argument for blame all commits

ddevlin:
For `blame` to detect moved or copied lines in all commits, the command line argument is `-CCC`, this was changed to `-CC` in the blame refactor.

- `-C` detects in the modifying commit
- `-CC` detects in the modifying commit AND the commit that created the file
- `-CCC` detects moved or copied lines in all commits

See: https://git-scm.com/docs/git-blame#git-blame--Cltnumgt


randy3k:
Thanks.

Issue #812, opened 2017-11-16
Show hook output during commit, push, etc.


tomv564:
Thanks for this plugin, especially the interactive staging support is a huge win.

Unfortunately I lose a lot of time due to pre-commit hooks running a full lint and test suite (not my choice). When I press `ctrl+enter`, all GitSavvy shows is `GitSavvy Make Commit` in the status bar for a half minute until this hook completes - there is no UI to show what work is being done in the background. (I do get the output at the end when the commit completes or fails).

I did some work capturing `stdout` or `stderr` output into the output panel while the hook is running. Is this something you would be interested in? I'll add a demo gif in about a day.

asfaltboy:
Thank you for the kind words @tombv564 ! I would say that if there's a use case, we can always add a toggle to enable the stream, or even default to it if there's little to no performance hit.

Please submit a PR and we can discuss the details.

tomv564:
Added some screen recordings to PR #813

randy3k:
#813 is merged.

PR #813, opened 2017-11-16
Log githook progress

tomv564:
This is an exploration for new feature to show stdout/stderr output in a Sublime Text panel while a hook is running, as discussed in #812

Currently pre-push (logs to stdout) and pre-commit (logs to stderr) hooks have been tested on OS-X.

Some demos (bad gif conversion results in 2x slower playback, sorry for your time :) )

**Commit** (Commit view closes automatically)

gitsavvy-commit-hook

**Push** 

gitsavvy-push-hook

### DIFF
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -110,6 +110,14 @@
     "live_panel_output": false,
 
     /*
+        Command will exit if on the timeout if the timeout reached.
+        You may want to increase this if you have a slow disk or a very
+        large repository. If you think there is something GitSavvy can do
+        to prevent this long running git command, open an issue.
+    */
+    "live_panel_output_timeout": 10000,
+
+    /*
         Use the Sublime configured syntax for COMMIT_EDITMSG rather than
         the custom bundled syntax that comes with GitSavvy.
     */
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -139,6 +139,11 @@
     "show_input_in_output": true,
 
     /*
+        Change this to `true` to print the stdin in the panel output.
+     */
+    "show_stdin_in_output": false,
+
+    /*
         Change this to `false` to suppress Git status in ST3 status bar.
      */
     "git_status_in_status_bar": true,
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -104,6 +104,12 @@
     "close_panel_for": [],
 
     /*
+        Whether git and githook output is logged live
+        If false, results are shown after execution.
+    */
+    "live_panel_output": false,
+
+    /*
         Use the Sublime configured syntax for COMMIT_EDITMSG rather than
         the custom bundled syntax that comes with GitSavvy.
     */
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -10,6 +10,8 @@ import os
 import subprocess
 import shutil
 import re
+from contextlib import contextmanager
+import threading
 
 import sublime
 
@@ -50,6 +52,46 @@ GIT_TOO_OLD_MSG = "Your Git version is too old. GitSavvy requires {:d}.{:d}.{:d}
 GIT_REQUIRE_MAJOR = 1
 GIT_REQUIRE_MINOR = 9
 GIT_REQUIRE_PATCH = 0
+
+
+class LoggingProcessWrapper(object):
+    def __init__(self, process):
+        self.process = process
+        self.stdout = b''
+        self.stderr = b''
+
+    def read_stdout(self):
+        # TODO: check for except IOError as err:
+        for line in self.process.stdout:
+            self.stdout = self.stdout + line
+            util.log.panel_append(line.decode())
+
+    def read_stderr(self):
+        for line in self.process.stderr:
+            self.stderr = self.stderr + line
+            util.log.panel_append(line.decode())
+
+    def communicate(self, stdin):
+        """
+        Emulates Popen.communicate
+        Writes stdin (if provided)
+        Logs output from both stdout and stderr
+        Returns stdout, stderr
+        """
+        if stdin is not None:
+            self.process.stdin.write(stdin)
+            self.process.stdin.flush()
+            self.process.stdin.close()
+
+        stdout_thread = threading.Thread(target=self.read_stdout)
+        stdout_thread.start()
+        stderr_thread = threading.Thread(target=self.read_stderr)
+        stderr_thread.start()
+
+        self.process.wait()
+        # TODO: manage threads here?
+
+        return self.stdout, self.stderr
 
 
 class GitCommand(StatusMixin,
@@ -103,10 +145,7 @@ class GitCommand(StatusMixin,
         if args[0] in close_panel_for:
             sublime.active_window().run_command("hide_panel", {"cancel": True})
 
-        log_stderr_for = savvy_settings.get("log_stderr_for") or []
-        log_stderr = args[0] in log_stderr_for
-        log_stdout_for = savvy_settings.get("log_stdout_for") or []
-        log_stdout = args[0] in log_stdout_for
+        live_logging = savvy_settings.get("live_output_logging") or False
 
         stdout, stderr = None, None
 
@@ -140,22 +179,16 @@ class GitCommand(StatusMixin,
                                  env=environ,
                                  startupinfo=startupinfo)
 
-            if log_stderr or log_stdout:
-                util.log.panel(command_str + "\n\n")
-                if stdin:
-                    p.stdin.write(stdin.encode(encoding=stdin_encoding) if encode else stdin)
-                    p.stdin.flush()
-                    p.stdin.close()
-                    util.log.panel_append(stdin + "\n\n")
+            if stdin is not None and encode:
+                stdin = stdin.encode(encoding=stdin_encoding)
 
-                if log_stderr:
-                    stdout, stderr = self.log_and_wait(p, p.stderr)
-                else:  # log_stdout
-                    stdout, stderr = self.log_and_wait(p, p.stdout)
-
+            if show_panel and live_logging:
+                if savvy_settings.get("show_input_in_output"):
+                    util.log.panel(command_str + "\n\n")
+                wrapper = LoggingProcessWrapper(p)
+                stdout, stderr = wrapper.communicate(stdin)
             else:
-                stdout, stderr = p.communicate(
-                    (stdin.encode(encoding=stdin_encoding) if encode else stdin) if stdin else None)
+                stdout, stderr = p.communicate(stdin)
 
             if decode:
                 stdout, stderr = self.decode_stdout(stdout, savvy_settings), stderr.decode()
@@ -217,25 +250,6 @@ class GitCommand(StatusMixin,
                         sublime.error_message(FALLBACK_PARSE_ERROR_MSG)
                         raise fallback_err
                 raise unicode_err
-
-    def log_and_wait(self, process, pipe):
-        captured_output = b''
-
-        for line in pipe:
-            captured_output = captured_output + line
-            util.log.panel_append(line.decode())
-
-        process.wait()
-        stderr = process.stderr.read()
-        stdout = process.stdout.read()
-
-        if len(captured_output) > 0:
-            if pipe == process.stdout:
-                stdout = captured_output + stdout
-            else:
-                stderr = captured_output + stderr
-
-        return stdout, stderr
 
     @property
     def encoding(self):
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -192,11 +192,18 @@ class GitCommand(StatusMixin,
 
             if show_panel and live_panel_output:
                 if savvy_settings.get("show_input_in_output"):
-                    util.log.panel(command_str + "\n\n")
+                    util.log.panel("> {}\n\n".format(command_str))
                 wrapper = LoggingProcessWrapper(p)
                 stdout, stderr = wrapper.communicate(stdin)
             else:
                 stdout, stderr = p.communicate(stdin)
+                if show_panel:
+                    util.log.panel("")
+                    if savvy_settings.get("show_stdin_in_output") and stdin is not None:
+                        util.log.panel_append("STDIN\n{}\n".format(original_stdin))
+                    if savvy_settings.get("show_input_in_output"):
+                        util.log.panel_append("> {}\n".format(command_str))
+                    util.log.panel_append("{}\n{}".format(stdout, stderr))
 
             if decode:
                 stdout, stderr = self.decode_stdout(stdout, savvy_settings), stderr.decode()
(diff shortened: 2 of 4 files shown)


randy3k:
Awesome, could you also share your toy hooks for testing purpose?

core/git_command.py
```diff
@@ -96,6 +96,11 @@ def git(self, *args,
         if args[0] in close_panel_for:
             sublime.active_window().run_command("hide_panel", {"cancel": True})
 
+        log_stderr_for = savvy_settings.get("log_stderr_for") or []
+        log_stderr = args[0] in log_stderr_for
```
stoivo: @randy3k?
What is more readable?
```
log_stderr = args[0] in savvy_settings.get("log_stderr_for", [])
```

tomv564:
Yep, this was in `.git/hooks/pre-commit`:
```bash
#!/bin/sh

sleep 1
echo "yo"
sleep 1
echo "hey"
sleep 1
echo "who loves slow commits?"
sleep 2
echo "this guy"
sleep 2

exit 0
```

And this in `.git/hooks/pre-push`:

```bash
#!/bin/sh

sleep 1
echo "yo"
sleep 1
echo "hey"
sleep 1
echo "who loves slow pushes?"
sleep 2
echo "this guy"
sleep 2

exit 0
```

tomv564:
Some notes about the implementation:
* Appending to the panel assumes the panel was already created
* Avoids mucking with individual commands, adding support for _all_ commands in `git_command`
* Can only capture one stream at a time and the user needs to know which one per command 
  * With threads both could be read, with extra complexity.
* Essentially a poor `Popen.communicate` implementation, potential for nasty bugs here

I first tried to use the build system to run the 'hooked' git commands (see first commit), this solved the problem but did not report back to GitSavvy when it was done.

core/git_command.py
```diff
@@ -127,8 +132,24 @@ def git(self, *args,
                                  cwd=working_dir,
                                  env=environ,
                                  startupinfo=startupinfo)
-            stdout, stderr = p.communicate(
-                (stdin.encode(encoding=stdin_encoding) if encode else stdin) if stdin else None)
+
+            if log_stderr or log_stdout:
```
stoivo: It would be cool if we could write the code  a bit like this 
```

            if stdin:
                p.stdin.write(stdin.encode(encoding=stdin_encoding) if encode else stdin)
                p.stdin.flush()
                p.stdin.close()

            if show_panel:
                util.log.panel_append(stdin + "\n\n")
                if savvy_settings.get("show_input_in_output"):
                    util.log.panel(command_str + "\n\n")

            captured_output = b''
            for line in p.stdout:
                captured_output = captured_output + line
                if show_panel:
                    util.log.panel_append(self.decode_stdout(captured_output, savvy_settings))

            process.wait()
            stdout = captured_output + process.stdout.read()
            stderr = process.stderr.read()

            stdout = self.decode_stdout(stdout, savvy_settings)
            if show_panel:
                util.log.panel_append(self.decode_stdout(stdout, savvy_settings))

            stderr = self.decode_stdout(stderr, savvy_settings)
            if show_panel and len(stderr) > 0:
                util.log.panel_append("STDERR\n\n" + self.decode_stdout(stdout, savvy_settings))
```

I did not run this code so I don't think you can copy it completely.

And a  different thing, can a git command have stdout and stderr both as streams? if so should those be in two different threads?

tomv564: From this comment I see a few suggestions:
* new logic should respect `show_input_in_output` :+1:
* new logic should replace old reliable code using `Popen.communicate()` to reduce duplication and branching
  * We'll have to _really_ test this new logic :grin:

core/git_command.py
```diff
@@ -96,6 +96,11 @@ def git(self, *args,
         if args[0] in close_panel_for:
             sublime.active_window().run_command("hide_panel", {"cancel": True})
 
+        log_stderr_for = savvy_settings.get("log_stderr_for") or []
```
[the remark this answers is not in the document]
tomv564: Agree that few users would object to seeing the panel and its contents earlier. I'd love to make this the new `show_panel_for` behaviour.

The stdout / stderr split is an implementation detail: To avoid introducing threading I used blocking reads. We can only consume 1 stream while blocking at a time, so the user is forced to choose.

The real solution is to use threading to read and concatenate both streams.

core/git_command.py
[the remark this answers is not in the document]
randy3k: I think our convention is to do `args[0] in savvy_settings.get("log_stderr_for", [])`

asfaltboy:
I have just a couple of general comments:

- I would prefer not to make this the default behaviour at this time, at least until it's been tested thoroughly by us and other users.
- The condition forking we have going on, and the sheer length of the `GitCommand.git` method makes me think that we should extract the internals of processing the command output / communication in a separate function. This would also make somewhat easier to test.

tomv564:
@asfaltboy I'll update this PR this weekend with threaded stdout/stderr reading and cleaned up code. Thanks for the comments!

asfaltboy:
Cool, I'll try to dial in the weekend to help out

tomv564:
A quick summary of the changes:
* A single setting (`live_panel_output`) adds the new behaviour to `show_panel_for`
* Switched to threads: Both stdout and stderr are always live-logged to the panel
* Cleaned up the `git` command - the differences with Popen.communicate are now in a wrapper
* Now honours `show_input_in_output` setting.

stoivo:
Very cool, I will enable it for more command to see if it has some flows. 

I have found some other thing we should do while we are at it (if you add me as a contributor, I may have time to push some small things into this PR)

These lines should not be there anymore, 
https://github.com/tomv564/GitSavvy/blob/158b13a8a02ba46dc95ac74c0f2de885d6ff823f/core/git_command.py#L229-L233

I got an issue with commiting. 
http://recordit.co/DtTz12vDRd a screen cast what I did
Typo in the screencast : Where is STD ERR

randy3k:
@stoivo Recently, github added a new feature to allow maintainer of a repo to push to the branch of the submitted PR. You should be able to push commits to it unless @tomv564 has deliberately disabled it.

screen shot 2017-11-18 at 6 42 55 pm

tomv564:
@stoivo feel free to push to the branch, you should have the maintainer access @randy3k mentioned.

The lines you want to remove, are those not responsible for showing the output if `live_panel_output` is not enabled? 
They could be bypassed if `live_panel_output` is ON. 

I can help troubleshoot the issue you recorded later - did any errors show up sublime's panel?

stoivo:
No errors, no nothing. Just stop responsing.

tomv564:
Either `Popen.communicate` handles this automatically or I messed it up, but even an empty commit message must be sent into `git commit -F` or the process will hang indefinitely waiting for input. See demo of fix below:

empty commit message fixed

stoivo:
Very nice, seems to work.
I will continue to test it.

stoivo:
If you enable it for `log`, then `git: log` fails.

core/git_command.py
```diff
@@ -184,11 +184,16 @@ def git(self, *args,
 
             if show_panel and live_panel_output:
                 if savvy_settings.get("show_input_in_output"):
-                    util.log.panel(command_str + "\n\n")
+                    util.log.panel("> {}\n\n".format(command_str))
                 wrapper = LoggingProcessWrapper(p)
                 stdout, stderr = wrapper.communicate(stdin)
             else:
                 stdout, stderr = p.communicate(stdin)
+                if show_panel:
```
tomv564: stderr and stdout are no longer decoded before being shown in the panel. Intentional?

tomv564:
What's your opinion on the `git log` issue? Already sorted? Must fix?

core/git_command.py
stoivo: Not,  Intentional.  
```pyhton
            if show_panel and live_panel_output:
                util.log.panel("")
                if savvy_settings.get("show_stdin_in_output") and stdin is not None:
                    util.log.panel_append("STDIN\n{}\n".format(original_stdin))
                if savvy_settings.get("show_input_in_output"):
                    util.log.panel_append("> {}\n\n".format(command_str))
                wrapper = LoggingProcessWrapper(p)
                stdout, stderr = wrapper.communicate(stdin)
            else:
                stdout, stderr = p.communicate(stdin)

            if decode:
                stdout, stderr = self.decode_stdout(stdout, savvy_settings), stderr.decode()

            if show_panel and !live_panel_output:
                if savvy_settings.get("show_input_in_output"):
                    util.log.panel("> {}\n{}\n{}".format(command_str, stdout, stderr))
                else:
                    util.log.panel("{}\n{}".format(stdout, stderr))
```

stoivo:
I think the `git log` is a must fix issue. I would not recommend anybody to enable log for live log. But we need to know why it doesn't work

tomv564:
@stoivo, I hope the issue was specific to the graph view? 
All other `git log` commands succeeded with live output but `git log --pretty` hung forever.

Disabling the pager by adding `--no-pager` to git fixes the graph view (see last commit)
`Popen.communicate` must do something to avoid the pager (which we probably should duplicate, but I don't have time to research). 

At the very least it's good to document this issue in the settings?

randy3k:
The default for `live_panel_output` should be set `false` because we are not sure what other commands will break when it's one. Otherwise, I guess we are good to go.

stoivo:
@tomv564  we should add `--no-pager` and the first argument if `live_panel_output` is on. 
man git 
```
       -p, --paginate
           Pipe all output into less (or if set, $PAGER) if standard output is a terminal. This overrides the pager.<cmd>
           configuration options (see the "Configuration Mechanism" section below).

       --no-pager
           Do not pipe Git output into a pager.
```

I agree, @randy3k. But it is a really cool feature.

tomv564:
I'm sorry, no time to add this logic, but @stoivo please add it if you feel it is necessary.

stoivo:
Sorry for taking so long. I will see what I can do the  following days

stoivo:
https://github.com/divmain/GitSavvy/pull/813/commits/8ded5cd3bf6d8dfdadc1c5d1166452d753d67b7f solves the problem with not everything being returned in stdout but only the first line. 

Some references from cpython source 
https://github.com/python/cpython/blob/master/Lib/subprocess.py#L1151
https://github.com/python/cpython/blob/master/Lib/subprocess.py#L1175-L1182

So.. I think we are getting closer. 

Still one issue is that amend stop working for me. Dunno why

tomv564:
Nice fix waiting for the threads to exit. 
I don't recall any problems with amend (I used it just today, but I'm not on the latest commit in this branch).

core/git_command.py
```diff
@@ -90,6 +90,8 @@ def communicate(self, stdin):
         stderr_thread.start()
 
         self.process.wait()
+        stdout_thread.join()
+        stderr_thread.join()
```
asfaltboy: I think we can also provide [a timeout argument][1] here (asme as in the cpython source linked), to avoid ongoing communication from hanging indefinitely. What do you think of using the process method for this calculation? E.g:

```python
max_run_time = savvy_settings.get("live_panel_output_timeout", 10000)  # 10s timeout default

stdout_thread.join(self.process._remaining_time(max_run_time))
stderr_thread.join(self.process._remaining_time(max_run_time))
```

What do you think?

[1]: https://docs.python.org/3/library/threading.html#threading.Thread.join

GitSavvy.sublime-settings
```diff
@@ -102,6 +102,22 @@
      */
     "close_panel_for": [],
 
+    /*
+        Whether git and githook output is logged live
+        If false, results are shown after execution.
+     */
+    "live_panel_output": true,
```
asfaltboy: Agree with @randy3k this should be false by default, at least for the time being.

core/git_command.py
stoivo: Agree

stoivo:
I think this should be all. 

I am aware of one issue. `git log ...`. Git log is using a pager and there the method return before the log command is done. 

I have 2 different approaches to a solution.
1. Add the --no-pager if the is a log command. 
2. Investigate git and popen to understand why it works when we user popen.comuticate. 

Either way, we want to go. I would like to merge this PR like it is. And not recommend anyone to enable live for log.  And who really wants to read the anyway?

stoivo:
I rebased everything. I think this is solid now. Please read over and check if it looks okey.

I am aware of one issue. If you enable live log for `git log`  we will have problems. Somehow Popen maks git use a no pager. We could sneak in a --no-pager every time it is a git log command.

stoivo:
This should be all, Ci passes.

GitSavvy.sublime-settings
```diff
@@ -103,6 +103,30 @@
...
+        Whether git and githook output is logged live
+        If false, results are shown after execution.
+    */
+    "live_panel_output": false,
+
+    /*
+        Blacklist commands you don't want in the output panel at all.
+        Always start each entry with git ...
+    */
+    "show_panel_blacklist": [
```
randy3k: it seems a bit fragile to blacklist the full commands. Btw, why do we need to blacklist these commands at the first place?

core/git_command.py
```diff
@@ -127,8 +188,27 @@ def git(self, *args,
...
+                util.log.panel("")
+                if savvy_settings.get("show_stdin_in_output") and stdin is not None:
+                    util.log.panel_append("STDIN\n{}\n".format(original_stdin))
+                if savvy_settings.get("show_input_in_output"):
+                    util.log.panel_append("> {}\n\n".format(command_str))
+                wrapper = LoggingProcessWrapper(p)
+                stdout, stderr = wrapper.communicate(stdin)
+            else:
+                stdout, stderr = p.communicate(stdin)
+                if show_panel:
```
randy3k: also show stdin when `show_stdin_in_output` is True.

stoivo: What are you pointing at here?

I check `show_input_in_output` and the line below. And I don't want to show the stdin if you don' want to see stdout.

GitSavvy.sublime-settings
stoivo: Let's say I enable live panel for `status`. Without this blacklist, I would then show a new output every time I changed the file when trying to update the status bar. 

We could take another approach to this and rend one more argument to the `git`method. Then add `ignore_live_log` to those very uninteresting commands.

randy3k: is it even useful to allow live panel for `status`? I thought it is mainly useful for `git pull` and `git push`. It doesn't look like there are any hooks which are related to `git status`. 

Reading one more argument is also not a good solution, but I have no suggestions neither.

core/git_command.py
randy3k: I was thinking that we should insert
```python
if savvy_settings.get("show_stdin_in_output") and stdin is not None:
                   util.log.panel_append("STDIN\n{}\n".format(original_stdin))
```
to here

GitSavvy.sublime-settings
stoivo: I have enabled it for `apply` not because I there are any hooks to apply but because I want to see what is being applied. Right after the apply was done it would overwrite it with one of the commands with I blacklisted. At that time I might I enable live log for everything.

core/git_command.py
stoivo: Now I get it, and yes we should.

randy3k:
I will test this PR with `live_panel_output` turned off for a few days before merging it.

stoivo:
We should remove the blacklist option. It doesn't make much sense.

stoivo:
Updated

randy3k:
I believe we could merge it and give it a try.

PR #816, opened 2017-11-24
Commit view will respect include_unstaged

### LINKED ISSUE #815: Show diff view also when when committing unstaged files
When files are staged before they are committed, the `COMMIT` tab shows a complete diff commit view.

GitSavvy should allow to show the same view when doing ` [C] commit, including unstaged`; at the moment this action makes it look like nothing is being committed, and makes it harder to see what the commit diffs are.

stoivo:
Fix #815

### DIFF
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -93,6 +93,7 @@ class GsCommitInitializeViewCommand(TextCommand, GitCommand):
                      else COMMIT_HELP_TEXT)
         self.view.settings().set("git_savvy.commit_view.help_text", help_text)
 
+        include_unstaged = self.view.settings().get("git_savvy.commit_view.include_unstaged", False)
         option_amend = self.view.settings().get("git_savvy.commit_view.amend")
         if option_amend:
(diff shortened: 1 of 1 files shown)


core/commands/commit.py
```diff
@@ -110,22 +111,23 @@ def run(self, edit):
...
+
+        if not include_unstaged:
+            initial_git_command.append("--cached")
+
+        if option_amend:
+            initial_git_command.append("HEAD^")
+        elif include_unstaged:
+            initial_git_command.append("HEAD")
 
+        initial_text += self.git(*initial_git_command) if show_commit_diff != 'false' else ''
```
randy3k: calling it `initial_git_command` seems counterintuitive as it is the final git_commands...perhaps we could just call it `args`. For the same reason, `initial_text` should be renamed differently, maybe `diff_text`?

stoivo: I agree better call it git_args

stoivo: naming is hard :P

PR #824, opened 2017-12-07
git remote rename and show remote url

randy3k:
- new command `git: remote rename`

- Enhancement: show url when using `git: remote`

screen shot 2017-12-07 at 2 39 56 pm

### DIFF
--- a/core/commands/remote.py
+++ b/core/commands/remote.py
@@ -1,38 +1,38 @@
 import sublime
-from sublime_plugin import TextCommand
+from sublime_plugin import WindowCommand
 
 from ..git_command import GitCommand
 from ...common import util
 from ..ui_mixins.quick_panel import show_remote_panel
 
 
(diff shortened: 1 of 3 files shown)


core/commands/remote.py
```diff
@@ -1,39 +1,39 @@
...
 
 
-class GsRemoteAddCommand(TextCommand, GitCommand):
+class GsRemoteAddCommand(WindowCommand, GitCommand):
     """
     Add remotes
     """
 
-    def run(self, edit):
+    def run(self):
```
stoivo: Should WindowCommand not take an edit, 
I tested and removed it

stoivo:
Should merge when CI is done.

core/commands/remote.py
randy3k: You are right. 👍

PR #825, opened 2017-12-07
Enhancement: reset to original commit after fixup

randy3k:
- reset to original commit after fixup 
- fixup should be safe even if there are untracked files

### DIFF
--- a/core/commands/fixup.py
+++ b/core/commands/fixup.py
@@ -13,7 +13,7 @@ class GsFixupFromStageCommand(GsLogCurrentBranchCommand):
          untracked_entries,
          conflict_entries) = self.sort_status_entries(self.get_status())
 
-        if len(unstaged_entries) + len(untracked_entries) + len(conflict_entries) > 0:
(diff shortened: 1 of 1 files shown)


core/commands/fixup.py
```diff
@@ -59,6 +59,7 @@ def do_action(self, commit, **kwargs):
 
             self.rewrite_active_branch(base_commit, commit_chain)
         except Exception as e:
+            self.git("reset", "--hard", commit)
```
stoivo: Why do we want to do this?
If we do a reset --hard commit (This is the commit you want to rebase the changes into??), would we remove the commits from commit unto HEAD?

randy3k: Ai, we should reset only....reset —hard would be too much. The idea is to get rid of the fixup commit.

stoivo: `HEAD^1` ?
So one commit before HEAD?

randy3k: 👍

randy3k:
pushed, please review again.

Issue #644, opened 2017-05-02
Stash changes while using rebase dashboard.


randy3k:
Currently, manipulating commits are not allowed if the repo status is unclean. It could be solved by modifying [this](https://github.com/divmain/GitSavvy/blob/2938321db00b5d15ab9281b4a99e528d78e90eb7/core/interfaces/rebase.py#L477) as

```python
        if len(changed_files):
            def stage_changes_and_run_async():
                self.git("stash")
                try:
                    self.run_async()
                finally:
                    self.git("stash", "pop")

            sublime.set_timeout_async(stage_changes_and_run_async, 0)
        else:
            sublime.set_timeout_async(self.run_async, 0)
```

But it seems that it is quite dangerous to do so...any comments?

stoivo:
Maybe ask or have it as a setting. It would be really stage if you have some changes and suddenly those changes is in a stash after you renamed a old commit(and it failed)

```python
        if len(changed_files):
            def stage_changes_and_run_async():
                if not sublime.ok_cancel_dialog(
                        "Unable to manipulate commits while repo is in unclean state.",
                        "Would you like to stash while doing this operation?")
                    return

                self.git("stash")
                try:
                    self.run_async()
                finally:
                    self.git("stash", "pop")


            sublime.set_timeout_async(stage_changes_and_run_async, 0)
        else:
            sublime.set_timeout_async(self.run_async, 0)
```

It is frustrating when it complains that I have some changed files. I like the idea :D

randy3k:
I beleive that the try finally block will reapply the changes even the modification fails.

stoivo:
But the pop might fail.

randy3k:
I cannot think of any situations in which `run_async` could be executed without errors but `git stash pop` fails...Though it is exactly why I said this is a kind of dangerous.

stoivo:
Deleting a commit would make git stash pop fail. I think ?

randy3k:
Ai, you are very correct!

stoivo:
So, would you like to add it? 
or add a shortcut to create a stash from the rebase dashboard. 
or every time it fails because of an unclean state, ask it the user would like to stash and redo the action?

randy3k:
> every time it fails because of an unclean state, ask it the user would like to stash and redo the action?

👍 we should also undo absolutely everything if `stash pop` fails.

arturolinares:
[...]

PR #826, opened 2017-12-07
Enhancement: manipulate commits with untracked files

### LINKED ISSUE #789: Untracked files causes rebase not to work
The error "Unable to manipulate commits while repo is in unclean state" is shown even when in the repo there are files that are untracked, while this should not happen.

- Sublime Text version number: 3148
- Git version number: 1.9.1
- OS type and version: Ubuntu 14.04

1. `touch foo.bar` in the repo dir
2. enter in GitSavvy rebase mode
3. Try to do any action

The command will fail since with the said error.

randy3k:
It should be safe to manipulate commits when there are untracked files. Related to #825, #644 

close #789

### DIFF
--- a/core/interfaces/rebase.py
+++ b/core/interfaces/rebase.py
@@ -491,9 +491,12 @@ class RewriteBase(TextCommand, GitCommand):
     def run(self, edit):
         self.interface = ui.get_interface(self.view.id())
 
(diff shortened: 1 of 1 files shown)


randy3k:
Agree.

PR #828, opened 2017-12-07
Enhancement: use ssh url instead of clone url when adding fork

randy3k:
`https://github.com/randy3k/GitSavvy.git` -> `git@github.com:randy3k/GitSavvy.git`

### DIFF
--- a/core/commands/remote.py
+++ b/core/commands/remote.py
@@ -11,9 +11,12 @@ class GsRemoteAddCommand(WindowCommand, GitCommand):
     Add remotes
     """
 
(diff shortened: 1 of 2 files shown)


stoivo:
Do you know why Github recommends https?
https://help.github.com/articles/which-remote-url-should-i-use/

randy3k:
Not sure, perhaps we should prompt and ask user to choose from them.

randy3k:
Now it prompts a panel of urls and use `gs_remote_add`.

PR #830, opened 2017-12-07
Enhancement: support .sublime-color-scheme in inline diff view

### LINKED ISSUE #829: inline diff is not working with the new `.sublime-color-scheme``
As ThemeGenerator  https://github.com/divmain/GitSavvy/blob/9bc0743b2fd0ec91751baf13cc9dc6e874c3f8d2/common/theme_generator.py#L37-L50 only works with XML based `.tmTheme`.

randy3k:
close #829 

ThemeGenerator now understands the new json based color scheme (required dev channel ST).

### DIFF
--- a/common/theme_generator.py
+++ b/common/theme_generator.py
@@ -6,6 +6,8 @@ to a view.
 
 import os
 from xml.etree import ElementTree
+import json
+from collections import OrderedDict
 
 import sublime
 from . import util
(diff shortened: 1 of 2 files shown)


common/theme_generator.py
```diff
@@ -56,32 +86,45 @@ def add_scoped_style(self, name, scope, **kwargs):
...
+    """
+    A theme generator for the new syntax `.sublime-color-scheme`
+    """
 
-        # Sublime expects `/`-delimited paths, even in Windows.
-        theme_path = os.path.join("Packages", path_in_packages).replace("\\", "/")
-        target_view.settings().set("color_scheme", theme_path)
+    hidden_theme_extension = "hidden-color-scheme"
+
+    def __init__(self, original_color_scheme):
```
stoivo: Should we share most of this init method with `XMLThemeGenerator` since it is covering cases where the file was not found?

common/theme_generator.py
```diff
@@ -56,32 +86,45 @@ def add_scoped_style(self, name, scope, **kwargs):
...
+            paths = sublime.find_resources(original_color_scheme)
+            if not paths:
+                raise IOError("{} cannot be found".format(original_color_scheme))
+            for path in paths:
+                if path.statswith("Packages/User/"):
+                    color_scheme_json = sublime.load_resource(original_color_scheme)
+                    break
+            color_scheme_json = sublime.load_resource(paths[0])
+
+        self.dict = json.loads(color_scheme_json, object_pairs_hook=OrderedDict)
```
stoivo: The only thing which is different is this line?

common/theme_generator.py
randy3k: 👍

randy3k:
@stoivo please review again. Thanks for your fast respond. 👍

stoivo:
I did not test this but I that's why we have dev

thezimmee:
Can anybody tell me when this will be released? Or if there is a dev channel somewhere so I can get this feature?

randy3k:
Are you using the dev channel of ST? If it is the case, you could checkout the dev branch of GitSavvy.

thezimmee:
I am. I just didn't want to lose auto-update with installing from package control. Any idea when this might be released?

randy3k:
FYI, Package control also updates version controlled packages.

thezimmee:
That's great, I did not know that. Will give that a try. Thank you!

PR #833, opened 2017-12-09
Feature/pull remote tracking

randy3k:
The current implementation of `git: pull` requires users to choose a remote and then a branch to pull. However, 99% of times and 90% of users will just hit "enter" and "enter" to pull from the default remote tracking branch. 

This PR changes the behaviour of `git: pull` to pull from remote tracking branch directly if it is found. And the original command `git: pull` is now renamed as `git: pull from branch` to mirror `git: push from existing branch` (the word 'existing' seems redundant here).

### DIFF
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -141,8 +141,12 @@
         "command": "gs_pull"
     },
     {
-        "caption": "git: pull with rebase",
-        "command": "gs_pull",
+        "caption": "git: pull from branch",
+        "command": "gs_pull_from_branch"
+    },
+    {
+        "caption": "git: pull from branch with rebase",
+        "command": "gs_pull_from_branch",
         "args": { "rebase": true }
     },
     {
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -141,6 +141,11 @@
         "command": "gs_pull"
     },
     {
+        "caption": "git: pull with rebase",
+        "command": "gs_pull",
+        "args": { "rebase": true }
+    },
+    {
         "caption": "git: pull from branch",
         "command": "gs_pull_from_branch"
     },
--- a/core/commands/pull.py
+++ b/core/commands/pull.py
@@ -6,7 +6,19 @@ from ...common import util
 from ..ui_mixins.quick_panel import show_branch_panel
 
 
-class GsPull(WindowCommand, GitCommand):
+class GsPullBase(GitCommand):
+
+    def do_pull(self, remote, remote_branch, rebase):
+        """
+        Perform `git pull remote branch`.
+        """
+        sublime.status_message("Starting pull...")
+        self.pull(remote=remote, remote_branch=remote_branch, rebase=rebase)
+        sublime.status_message("Pull complete.")
+        util.view.refresh_gitsavvy(self.window.active_view())
+
+
+class GsPull(WindowCommand, GsPullBase):
     """
     Pull from remote tracking branch if it is found. Otherwise, use GsPullFromBranchCommand.
     """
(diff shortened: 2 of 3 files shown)


Default.sublime-commands
```diff
@@ -141,8 +141,12 @@
         "command": "gs_pull"
     },
     {
-        "caption": "git: pull with rebase",
-        "command": "gs_pull",
+        "caption": "git: pull from branch",
+        "command": "gs_pull_from_branch"
+    },
+    {
+        "caption": "git: pull from branch with rebase",
```
asfaltboy: Might make sense to also add a `git: pull with rebase` command, as this is the most common use case.

core/commands/pull.py
```diff
@@ -6,22 +6,41 @@
...
+
+    def run_async(self):
+        # honor the `pull.rebase` config implictly
+        rebase = self.git("config", "pull.rebase", throw_on_stderr=False) or False
+        if rebase and rebase.strip() == "true":
+            rebase = True
+        upstream = self.get_upstream_for_active_branch()
+        if upstream:
+            remote, remote_branch = upstream.split("/", 1)
+            self.pull(remote=remote, remote_branch=remote_branch, rebase=rebase)
```
asfaltboy: This is only a subset of what `GsPullFromBranchCommand.do_pull` does.

Instead of calling `pull()` here directly, can we perhaps extend the more specific class to accept remote/branch and:

```python
self.window.run_command("gs_pull_from_branch", {"rebase": rebase, "remote": remote, "remote_branch": remote_branch})
```

Alternatively, we can refactor these commands to both define `def do_pull(self, remote, remote_branch):`; either by inheritance or with a mixin...

Default.sublime-commands
randy3k: The `git: pull` in this PR is aware of `git config pull.rebase`. Maybe there is no need for `git: pull with rebase`?

Do you want to make it explicit that `git: pull` does not use `--rebase` and `git: pull with rebase` uses `--rebase`?

core/commands/pull.py
randy3k: added a `GsPullBase` class which is similar to `GsPushBase`

Default.sublime-commands
randy3k: Added `git: pull with rebase` which uses "rebase" explicitly while the behaviour of `git: pull` still depends on `git config pull.rebase`.

asfaltboy: I'm not sure which is best... If we mimic the `git pull` CLI function, defaulting to merge and optionally accepting e.g `--rebase` is the behaviour expected by users.

Now that we also allow override in the config, which is great, but makes one of the commands redundant. But, so does the CLI (user with `pull.rebase` set, won't ever need to use `git pull --rebase`). Those users who default to rebase will need to use I think `git pull --no-rebase` to get the merge behaviour.

In the end, it's about UX.. while I don't think adding commands for all arguments is really a good idea, some common commands shouldn't hurt. I think we should draw the line somewhere; in the end many of the use cases are down to personal (or team) preference. And since all of these commands can definitely be added by users in their local command files, I wonder if we need all of them enabled.

I have a suggestion, how about we add *but comment out* the more exotic cases (such as `git pull with merge`) ? `--pull` is really the only common option I know to `git pull`, even other branches is a rare use case.

randy3k: You are right that `git: pull` and `git: pull with rebase` are redundant when `pull.rebase` is true. But creating the command `git pull with merge` (`--no=rebase`)  just makes the situation more complicated. I call it done now.

PR #838, opened 2018-01-08
Internal: Use window.status_message instead of sublime.status_message

stoivo:
`sublime.status_message` is no longer officially there in the api doc so it is better to use in on a window. 

And  if we don't know which window we can use `sublime.active_window().status_message`


stoivo:
I have not tested every change but some of them

randy3k:
It is quite some changes. Will test and approve soon

Issue #364, opened 2016-03-03
GitLab support


divmain:
Work will occur in `gitlab-support` branch and can be tracked in #363 until it is merged.  The target feature-set for landing GitLab support in GitSavvy is as follows:

- [ ] support private GitLab servers
- [ ] issue number completion (with summary text in drop-down) in commit view
- [ ] collaborator completion (with names drop-down) in commit view
- [ ] `gitlab: open file on remote` with current line(s) highlighted
- [ ] `gitlab: open issues`
- [ ] `gitlab: open project`
- [ ] `gitlab: clone existing project`
  - will query the [project API](http://doc.gitlab.com/ce/api/projects.html#list-projects) for existing projects
  - once a selection is made, prompt user for location of clone
  - appropriate `git: clone` incantation
- [ ] `gitlab: add SSH key`
  - search `$HOME/.ssh/*.pub` for any existing keys
  - if found, will prompt user to accept one of the discovered keys (additional option will allow user to paste in key)
- [ ] `gitlab: create merge request`, which opens the link in browser.
- [ ] `gitlab: review merge request`, which displays a list of open merge requests.  Once a MR is selected, the user can select from:
  - [ ] `Checkout as detached HEAD.`
  - [ ] `Checkout as local branch.`
  - [ ] `Create local branch, but do not checkout.`
  - [ ] `View diff.`
  - [ ] `Set as in-progress.` so the MR can't be merged.
  - [ ] `Open in browser.`
- [ ] `gitlab: set remote for integration`
- [ ] `gitlab: add fork as remote`
- [ ] `gitlab: protect branch on remote` ([link](http://doc.gitlab.com/ce/api/branches.html#protect-repository-branch))
- [ ] `gitlab: unprotect branch on remote`
- [ ] `gitlab: builds`, which displays a list of project builds ([link](http://doc.gitlab.com/ce/api/builds.html)).  Once a build is selected, the user can select from:
  - [ ] `Cancel build.`
  - [ ] `Retry build.`
  - [ ] `Erase build.`
  - [ ] `Download artifacts.` (opens URL in browser)

Supersedes GitLab portion of #40.

asfaltboy:
Just wanted to mention that `github: open file on remote` already works _for gitlab_! Although jump to line works with hash `#L70-70` and NOT with `#L70-L70` as in github.
I, personally, would love to see these implemented: `gitlab: protect branch on remote (link)` and `gitlab: unprotect branch on remote`

[...]

PR #840, opened 2018-01-11
Gitlab merge-request review support

asfaltboy:
Hi guys, had some time to hack this up today; which, if accepted, would be the beginning of implementing PR #363

### DIFF
--- a/github/commands/commit.py
+++ b/github/commands/commit.py
@@ -14,7 +14,7 @@ from .. import git_mixins
 from ...common import util
 
 
-class GsGithubShowGithubIssuesCommand(TextCommand, GitCommand, git_mixins.GithubRemotesMixin):
+class GsGithubShowIssuesCommand(TextCommand, GitCommand, git_mixins.GithubRemotesMixin):
 
     """
     Display a panel of GitHub issues to either:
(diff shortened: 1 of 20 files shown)


github/commands/commit.py
```diff
@@ -14,7 +14,7 @@
 from ...common import util
 
 
-class GsShowGithubIssuesCommand(TextCommand, GitCommand, git_mixins.GithubRemotesMixin):
+class GsGithubShowGithubIssuesCommand(TextCommand, GitCommand, git_mixins.GithubRemotesMixin):
```
stoivo: Should there classes have Github twice in it's name

stoivo:
A good start. I think we can merge this into dev.

github/commands/commit.py
asfaltboy: Good point, I think I caught all of these in the command names, this one may have slipped through.

asfaltboy:
Thanks @stoivo for the review, I added a couple more commits, one that resolves the GsGithub naming issue mentions, and another that implements 3 of the "open on web" commands (file, repo home and issues). I also went ahead and marked the completed features in #364 todo list.

stoivo:
If you are not going to do anything more here for the next couple of weeks I would like to merge this.

asfaltboy:
Yes, I'm out of time for now, until next time... though I may need to add bitbucket support next :)

PR #849, opened 2018-02-04
Pedantic commit

### LINKED ISSUE #41: Pedantic commit messages

### LINKED ISSUE #423: Commit message: Visually represent first line limit of 50 characters
It is common practice to have the first line of a commit not exceed 50 characters. Currenty, GitSavvy does not "endorse" this because it only displays a ruler at column 80 for the entire message.
I have a a few ideas on how this could be implemented:
1. Show two rulers, one at 50 and one at 80. People that know about this convention will understand.
2. Adjust the ruler setting based on which line the caret is in (using `on_selection_modified`).
3. Only show ruler at 80 but make the syntax definition apply a different scope to any character past 50 in the first line.

All of these are feasable and comparatively easy to implement (especially 2.). However, 3. additionally has the downside of being rather obstrusive towards users that don't care about this convention.
2. has the advantage of wrap commands usually automatically adjusting to the first ruler in the view, which would be at 80 for the detailed message (and 50 when 1. is implemented). As such, 2. is highly preferred.

[Some recommend 72](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) as max-width for the message (for a console width of 80 with 4 characters padding on each side), so it makes sense to make this configurable imo. Setting to `80|80` would consequently disable this feature.
Related SO discussion: http://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting
[According to Linus](https://github.com/torvalds/linux/pull/17#issuecomment-5661185), this style is also used in the Linux kernel.

stoivo:
- The ruler appears when you are on those lines
- If the lines are too long they will be marked with the scope `invalid.deprecated.line-too-long.git-commit` where `invalid` is the same scope as SublimeLinter uses to is should be coverd by all Themes

fix #41 
fix #423

### DIFF
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -2,6 +2,7 @@ import os
 
 import sublime
 from sublime_plugin import WindowCommand, TextCommand
+from sublime_plugin import EventListener
 
 from ..git_command import GitCommand
 from ...common import util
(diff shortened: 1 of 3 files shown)


syntax/make_commit.sublime-settings
```diff
@@ -14,5 +14,4 @@
     "gutter": true,
     "line_numbers": true,
     "word_wrap": true,
```
randy3k: remove the ending comma

stoivo:
Thanks. I think this is a sensible default. And if anyone hates it. They can change it or turn it of.

core/commands/commit.py
```diff
@@ -139,6 +140,97 @@ def run(self, edit):
...
+            return
+
+        savvy_settings = sublime.load_settings("GitSavvy.sublime-settings")
+        if not savvy_settings.get('pedantic_commit'):
+            return
+
+        self.view = view
+        self.first_line_limit = savvy_settings.get('pedantic_commit_first_line_length')
+        self.body_line_limit = savvy_settings.get('pedantic_commit_message_line_length')
+        self.warrning_size = savvy_settings.get('pedantic_commit_warning_length')
```
asfaltboy: Typo in `warrning_size`; should this be `warning_size`, or even `warning_length` ?

core/commands/commit.py
```diff
@@ -139,6 +140,97 @@ def run(self, edit):
...
+
+        return new_rulers
+
+    def find_too_long_lines(self):
+        warning_lines = []
+        illegal_lines = []
+
+        first_line = self.view.lines(sublime.Region(0,0))[0]
+        length = first_line.b - first_line.a
+        if self.first_line_limit < length:
```
asfaltboy: I think the condition is clearer like this:

    if length > self.first_line_limit:

core/commands/commit.py
```diff
@@ -139,6 +140,97 @@ def run(self, edit):
...
+            warning_lines.append(sublime.Region(
+                first_line.a + self.first_line_limit,
+                min(first_line.a+self.first_line_limit+self.warrning_size, first_line.b)))
```
asfaltboy: I get vexed by the inconsistent spacing around the plus (+) operator... can we please always have space on either side of plusses?

core/commands/commit.py
stoivo: yes

stoivo:
Thanks. I will try to  finish #813

PR #850, opened 2018-02-04
Internal: Fix type in path for settings

stoivo:


asfaltboy:
Looks fine, but what's the actual problem we met with ? Can you give a link to discussion/issue please?

stoivo:
I did not create an issue.

If you where at the make_commit syntax we enable `super+,` to open set settings for this syntax, but the syntax and settings files are called with `_` not `-`, but the scope is with `-`

randy3k:
LGTM.

Issue #684, opened 2017-06-29
Command not found in custom `commit-msg` git hook


TheRealJoeLinux:
### Sublime Text version number:
- Build 3126

### Git version number:
- 2.7.4

### OS type and version:
- Ubuntu 16.04.2 LTS

### Console error output:
```
Traceback (most recent call last):
  File "core.commands.commit in /home/joelinux/.config/sublime-text-3/Installed Packages/GitSavvy.sublime-package", line 133, in <lambda>
  File "core.commands.commit in /home/joelinux/.config/sublime-text-3/Installed Packages/GitSavvy.sublime-package", line 153, in run_async
  File "core.git_command in /home/joelinux/.config/sublime-text-3/Installed Packages/GitSavvy.sublime-package", line 149, in git
  File "core.git_command in /home/joelinux/.config/sublime-text-3/Installed Packages/GitSavvy.sublime-package", line 112, in raise_error
GitSavvy.core.exceptions.GitSavvyError: `/usr/bin/git commit -q -F -` failed with following output:

Command-line tool 'sprintly' not found. Please ensure it is installed and on your path.
#0 - Proceed without Sprint.ly item number.


Error occurred. Commit aborted.
```

### A description of the problem:
I use [Sprint.ly](https://sprint.ly) for managing tickets. They provide a `commit-msg` git hook for repositories to automatically reference tickets within their database. However, even though `which sprintly` returns just fine from the command line, GitSavvy doesn't appear to find the command in my `$PATH`. The command is located on my system at `/usr/local/bin/sprintly`, and is properly located in `$PATH` for both my user and the root user. I'm not sure why it's not being found (or the error message is incorrect).

### Steps to reproduce, if possible:
If not already using Sprint.ly:
- Install [Sprintly-GitHub](https://github.com/sprintly/Sprintly-GitHub)
- Install the git hook by running `sprintly --install-hook` in the project directory (or just copy `commit-msg` into `.git/hooks/`)
- Attempt to commit _anything_ using GitSavvy

If the error message is correct, then I'm guessing this might be true of any custom git hook installed in the project, but *my* specific situation involves Sprint.ly, so I specified those steps.

stoivo:
Great issue report!

Seems like you are having the same issue as many Mac users, [this package](https://packagecontrol.io/packages/Fix%20Mac%20Path) fixes it for mac users.  Can you check if you have this issue by checking if sprintly is in your PATH inside sublime. 
[...]

PR #851, opened 2018-02-05
Docs: Add doc for setting PATH environment

stoivo:
As commented in #684

### DIFF
--- a/docs/debug.md
+++ b/docs/debug.md
@@ -38,3 +38,22 @@ Ocasionally when creating a new issue in GitSavvy, you will be requested to prov
       
       _Note: Take care to remove any sensitive information that may be recorded in GitSavvy._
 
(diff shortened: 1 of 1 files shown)


docs/debug.md
```diff
@@ -38,3 +38,22 @@ Ocasionally when creating a new issue in GitSavvy, you will be requested to prov
       
       _Note: Take care to remove any sensitive information that may be recorded in GitSavvy._
 
+# Other tools / PATH issues / Different behavure from terminal and inside Sublime
```
randy3k: behavure  -> behaviour / behavior

Also, AFAIK, the dev version of Sublime Text has addressed this issue of environmental variables.

stoivo: Thanks.

stoivo: Perhaps Sublime Text as improved the path issue. Still, I think will always be a possible issue.

PR #852, opened 2018-02-11
Configure pycodestyle

stoivo:
It would be good if we added this as a Unit test, or what do you think?

@randy3k, how would you do it?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,20 +26,18 @@ before_install:
 install:
     - sh travis.sh bootstrap
     - sh travis.sh install_package_control
+    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+            brew update;
+            brew install python3;
+      fi
+    - pip3 install python-coveralls pycodestyle;
 
 script:
     - sh travis.sh run_tests --coverage
     - sh travis.sh run_syntax_tests
+    - pycodestyle --statistics --count
 
 after_success:
-    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-            brew update;
-            brew install python3;
-            pip3 install python-coveralls;
-      fi
-    - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
-            pip install python-coveralls;
-      fi
     - coveralls
 
 # deploy:
--- a/GitSavvy.sublime-project
+++ b/GitSavvy.sublime-project
@@ -8,17 +8,8 @@
             ]
         }
     ],
-    "SublimeLinter":{
-        "linters":{
-            "pep8": {
-                "ignore": "E401",
-                // E401 multiple imports on one line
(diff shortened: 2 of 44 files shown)


randy3k:
We'll need a sublime dependency of pycodestyle in order write a test case. A simple solution is to run `pycodestyle` on travis.

.travis.yml
```diff
@@ -22,6 +22,8 @@ before_install:
       fi
     - git config --global user.email gitsavvy@gitsavvy.com
     - git config --global user.name GitSavvy
+    - pip install --upgrade pip
+    - pip install pycodestyle
```
randy3k: you may want to move the code block https://github.com/divmain/GitSavvy/blob/99c54673fb617b86d869b628d6df064fae78ec23/.travis.yml#L37-L46
to here

Note that, on macOS, `pip` corresponds to python 2, so `pip3` is needed.

stoivo:
```
pip install python-coveralls;
pip install pycodestyle;
```
vs
```
pip install python-coveralls pycodestyle;
```

randy3k: or
```
    - if [ "$TRAVIS_OS_NAME" == "osx" ]; then
            brew update;
            brew install python3;
      fi
    - pip3 install python-coveralls pycodestyle
```

stoivo: How did you check that? 
Am I not used to any CI?

.travis.yml
```diff
@@ -30,16 +30,14 @@ install:
...
 after_success:
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
             brew update;
             brew install python3;
-            pip3 install python-coveralls;
-      fi
-    - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
-            pip install python-coveralls;
       fi
+    - pip3 install python-coveralls pycodestyle;
```
randy3k: you need to install `pycodestyle` before running

stoivo:
Travis failed with 
```
$ pycodestyle --statistics --count
/Users/travis/.travis/job_stages: line 57: pycodestyle: command not found
The command "pycodestyle --statistics --count" exited with 
```

@randy3k ?

.travis.yml
randy3k: I just tried...

randy3k:
move 
```
    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
            brew update;
            brew install python3;
      fi
    - pip3 install python-coveralls pycodestyle
```
under `install`

.travis.yml
stoivo: move it to install section then ?

randy3k: ya

stoivo:
I will do a fixup rebase

randy3k:
Great. Thanks.

PR #853, opened 2018-02-11
Fix: log by author, also when you are not found

stoivo:
I don't know who to write it nicer. 

in ruby I would do 
```
"123".index("24") || 0
# or if it raises an exeption
Integer("ff") rescue 0
```


randy3k:
thanks.

PR #857, opened 2018-02-17
Tools to handle merge conflicts

randy3k:
Rebase dashboard has commands `use version from your commit` and `use version from new base` to manage rebase conflicts. The main contribution of this PR is to bring those commands to the Status dashboard when there are merge conflicts.

screen shot 2018-02-17 at 3 00 33 pm

This pr also improves active branch status when merging.
screen shot 2018-02-17 at 3 03 31 pm
screen shot 2018-02-17 at 3 03 36 pm

### DIFF
--- a/core/interfaces/status.py
+++ b/core/interfaces/status.py
@@ -56,7 +56,7 @@ class StatusInterface(ui.Interface, GitCommand):
       [u] unstage file                      [U] unstage all staged files
       [d] discard changes to file           [D] discard all unstaged changes
       [h] open file on remote
-      [M] launch external merge tool for conflict
+      [M] launch external merge tool
 
       [l] diff file inline                  [f] diff all files
       [e] diff file                         [F] diff all cached files
@@ -73,6 +73,8 @@ class StatusInterface(ui.Interface, GitCommand):
       [i] ignore file                       [t][g] create stash of staged changes only
       [I] ignore pattern                    [t][d] drop stash
 
+      [B] abort merge
+
       ###########
       ## OTHER ##
       ###########
@@ -83,8 +85,17 @@ class StatusInterface(ui.Interface, GitCommand):
       [SHIFT-tab] transition to previous dashboard
       [.]         move cursor to next file
       [,]         move cursor to previous file
-
+    {conflicts_bindings}
     -
+    """
+
+    conflicts_keybindings = """
+    ###############
+    ## CONFLICTS ##
+    ###############
+
+    [y] use version from your commit
+    [b] use version from the base
     """
 
     template_staged = """
@@ -111,6 +122,11 @@ class StatusInterface(ui.Interface, GitCommand):
       STASHES:
     {}
     """
+
+    def __init__(self, *args, **kwargs):
+        self.conflicts_keybindings = \
+            "\n".join(line[2:] for line in self.conflicts_keybindings.split("\n"))
+        super().__init__(*args, **kwargs)
 
     def title(self):
         return "STATUS: {}".format(os.path.basename(self.repo_path))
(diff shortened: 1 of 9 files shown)


core/interfaces/status.py
```diff
@@ -56,7 +56,7 @@ class StatusInterface(ui.Interface, GitCommand):
       [u] unstage file                      [U] unstage all staged files
       [d] discard changes to file           [D] discard all unstaged changes
       [h] open file on remote
-      [M] launch external merge tool for conflict
+      [M] launch external merge tool
```
asfaltboy: Should we perhaps move the `[M]` to the `CONFLICTS` section too?

core/interfaces/status.py
```diff
@@ -662,6 +683,64 @@ def run(self, edit):
...
+            else:
+                self.git("checkout", "--theirs", "--", path)
+                self.stage_file(path)
+        util.view.refresh_gitsavvy(self.view)
+
+    def is_commit_version_deleted(self, path, conflicts):
+        for conflict in conflicts:
+            if conflict.path == path:
+                return conflict.working_status == "D"
+        return False
```
asfaltboy: Correct me if I'm wrong, but this is section is mostly copy pasted from [`interfaces/rebase.py`][1] right? Can we perhaps DRY out the code by making the "line region col index" settable?

[1]: https://github.com/divmain/GitSavvy/blob/3c5b1ab49adc9b223d0a461d3e353a2e5cbfc64b/core/interfaces/rebase.py#L851-L877

core/interfaces/status.py
randy3k: I've never used an external merge tool. Is it useful to open unstaged file with merge tool? If the answer is yes, I guess we should put it here.

stoivo: Yes

core/interfaces/status.py
randy3k: Yes, you are right. I just copied and modified a few lines of the code. I agree we shall not repeat the code, but I am not sure the best way to refactor the path detection code.

randy3k: Let's merge this first and work on the refactoring later

asfaltboy: Could you please add a `# TODO: refactor this alongside interfaces.rebase. GsRebaseUseCommitVersionCommand ` comment to remind us?

core/interfaces/status.py
```diff
@@ -112,6 +123,11 @@ class StatusInterface(ui.Interface, GitCommand):
     {}
     """
 
+    def __init__(self, *args, **kwargs):
+        self.conflicts_keybindings = \
+            "\n".join(line[2:] for line in self.conflicts_keybindings.split("\n"))
```
stoivo: Is this line just not readable. 
I would use https://docs.python.org/3.5/library/textwrap.html#textwrap.dedent when I defining the variable instead.

randy3k: Good point. This line was also copied from the rebase dashboard
https://github.com/divmain/GitSavvy/blob/8011d33e935470df7f5d4a631fa4c7ddd14d06db/core/interfaces/rebase.py#L131-131

stoivo: Are you gonna make a PR for this?

randy3k: I created the new PR, but it doesn't yield the desired result
screen shot 2018-02-17 at 4 28 32 pm


It was because we actually do not want to dedent, we just want to remove two spaces from each line.

stoivo: Hm, is there a reason for why we have the two spaces at the beginning of every line?
I think we should remove those and id we want to add some more space to the left side we can increase margin in  https://github.com/divmain/GitSavvy/blob/3c5b1ab49adc9b223d0a461d3e353a2e5cbfc64b/syntax/status.sublime-settings#L9

randy3k: You are probably right but it will be a huge change since a lot of code was written in a form of `line[n:]` where `n` is the number of preceding spaces.

stoivo: hm, it should only be in this one file core/interfaces/status.py

I would change it and then keep it in dev for 2-3 weeks for us to test.

PR #862, opened 2018-02-17
Fix: branch relative regex

### LINKED ISSUE #860: Rebase dashboard is unacceptably slow for large repo
The rebase dashboard is near unusable for large repo (e.g., http://github.com/rstudio/rstudio). It is because [branch_relatives](https://github.com/divmain/GitSavvy/blob/8011d33e935470df7f5d4a631fa4c7ddd14d06db/core/git_mixins/rebase.py#L14-14) returns too many relatives when we are determining the rebase ref. We might need to restrict the number of relatives or change the way to determine the relatives.

A workaround, for now, is to set `rebase_default_base_ref` in project settings.

cc: @asfaltboy

randy3k:
an attempt to fix #860 

It seems it was caused by an incorrect regex. The original regex returns something like `foo~9^2` but not just the branch name `foo`.

### DIFF
--- a/core/git_mixins/rebase.py
+++ b/core/git_mixins/rebase.py
@@ -5,7 +5,7 @@ import time
 from ..exceptions import GitSavvyError
 from ...common import util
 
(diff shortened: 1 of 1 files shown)


randy3k:
Thanks for testing it.

Issue #835, opened 2017-12-19
Refresh status page on `git checkout`


gwenzek:
Feature request: Refresh status page on `git checkout`

Note: it works if you're running the checkout from the git status page but not from another tab. 

I usually have the status page open in a pane and I'm working in another pane.
Every time I'm checking out a branch (from the command palette), I have to select the status page to refresh it.
Could we refresh it automatically after a `git checkout` ?

Looking at the [CheckoutCommand](https://github.com/divmain/GitSavvy/blob/f87e7eefc0f81b5c04604c20906fa50cda87c740/core/commands/checkout.py#L40)
`util.view.refresh_gitsavvy(self.window.active_view(), refresh_sidebar=True)`
it seems we only refresh the active view. 
We should hunt for GitSavvy views (at least visible views in the current window) and refresh them.
Maybe we should modify `refresh_gitsavvy` to refresh all views or create a `refresh_all` method.
Other commands (pull, ...) could benefit from this.

WDYT ?

randy3k:
Alternatively, we would refresh the status dashboard when it is focused.

PR #869, opened 2018-02-22
Refresh all GitSavvy view on branch checkout

gwenzek:
Proof of concept implementation of #835.

Add a `refresh_all_visible_views` method.
It iterates through all the current window groups and refresh the active view of the group.

### DIFF
--- a/core/commands/checkout.py
+++ b/core/commands/checkout.py
@@ -38,7 +38,7 @@ class GsCheckoutBranchCommand(WindowCommand, GitCommand):
 
         self.git("checkout", branch)
         self.window.status_message("Checked out `{}` branch.".format(branch))
-        util.view.refresh_gitsavvy(self.window.active_view(), refresh_sidebar=True)
+        util.view.refresh_gitsavvy_interfaces(self.window, refresh_sidebar=True)
 
 
 class GsCheckoutNewBranchCommand(WindowCommand, GitCommand):
@@ -68,10 +68,9 @@ class GsCheckoutNewBranchCommand(WindowCommand, GitCommand):
             branch_name,
             self.base_branch if self.base_branch else None)
         self.window.status_message("Created and checked out `{}` branch.".format(branch_name))
-        util.view.refresh_gitsavvy(
-            self.window.active_view(),
-            refresh_sidebar=True,
-            interface_reset_cursor=True)
+        util.view.refresh_gitsavvy_interfaces(self.window,
+                                              refresh_sidebar=True,
+                                              interface_reset_cursor=True)
 
 
 class GsCheckoutRemoteBranchCommand(WindowCommand, GitCommand):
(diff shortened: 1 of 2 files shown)


core/commands/checkout.py
```diff
@@ -37,7 +37,12 @@ def on_branch_selection(self, branch):
 
         self.git("checkout", branch)
         sublime.status_message("Checked out `{}` branch.".format(branch))
-        util.view.refresh_gitsavvy(self.window.active_view(), refresh_sidebar=True)
+        refresh_all_visible_views(self.window)
+
+def refresh_all_visible_views(window):
```
stoivo: Can you move this into [common/util/view.py](https://github.com/divmain/GitSavvy/blob/master/common/util/view.py#L81)

core/commands/checkout.py
```diff
@@ -37,7 +37,12 @@ def on_branch_selection(self, branch):
 
         self.git("checkout", branch)
         sublime.status_message("Checked out `{}` branch.".format(branch))
-        util.view.refresh_gitsavvy(self.window.active_view(), refresh_sidebar=True)
+        refresh_all_visible_views(self.window)
```
stoivo: Move and rename to 
`util.view.refresh_all_gitsavvy`

stoivo:
Also rebase from `dev`

core/commands/checkout.py
gwenzek: Done.

core/commands/checkout.py
gwenzek: Done.

gwenzek:
As requested:
* rebased on dev
* moved the method to `common.util.view` and renamed it to `refresh_all_gitsavvy`
Improvements:
* only refresh the file list once (potentially costly)
* also use `refresh_all_gitsavvy` for `GsCheckoutNewBranchCommand` and `GsCheckoutRemoteBranchCommand`

asfaltboy:
While I ❤️  the improvements that this change introduces, I feel this defaults to a very generic refresh of all views. For example, sometimes I want to open some GitSavvy (a graph dashboard, a diff) multiple times, on different branches, to compare the code/commits in the versions. I fear that refreshing all views may break this use case (though I haven't tested it yet).

Should we perhaps limit this only to the status or branch dashboards? Or do you see any other views where this use case fits?

stoivo:
I think we can add the limit to only reload the interfaces. BranchInterface, RebaseInterface StatusInterface and TagsInterface. Any of those should always be as updated as possible.

gwenzek:
I modified the method to only refresh interfaces.
Should I rename it to something like `refresh_gitsavvy_interfaces` ?

The build tools are failing during the installation of Package Control, seems unrelated to my changes:
`Fail to install Package Control.`

gwenzek:
I rename the method to `refresh_gitsavvy_interfaces` since it only refresh interfaces now.
The build still fails at the Package Control step, so it's hard for me to tell if I have a messed up spacing or such.

PR #880, opened 2018-03-03
Fix: filter branch HEAD

randy3k:
`for-each-ref` may list `HEAD` as one of the branches.

Before
screen shot 2018-03-03 at 12 02 45 am

After
screen shot 2018-03-03 at 12 02 55 am


randy3k:
the failures of os builds are irrelevant, they are fixed in #881

PR #883, opened 2018-03-05
Fix: hash tag sign is a valid char for branch name

randy3k:
as title, branch names could contain `#`


randy3k:
Sth like `fix#1` and `fix#2` to fix issues.

PR #886, opened 2018-03-07
Fix: use check-ref-format to check branch name validity

randy3k:
Instead of using a regular expression to validate branch name, use the git command `check-ref-format` to validate it. It should give a better result than using a regular expression.

### DIFF
--- a/core/commands/checkout.py
+++ b/core/commands/checkout.py
@@ -1,4 +1,3 @@
-import re
 import sublime
 from sublime_plugin import WindowCommand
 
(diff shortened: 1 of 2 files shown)


stoivo:
`check#-ref-fo#rma👍ta` is valid but doesn't work

randy3k:
It works for me.
screen shot 2018-03-11 at 2 21 18 pm

stoivo:
Good for you, I will check again soon.

randy3k:
Thanks @stoivo

stoivo:
On another bug
First, try to create a branch with name `hello one`. It fails then set it to `hello-one` and you get the exception in the console.
http://recordit.co/yhKxnUZaSL

randy3k:
@stoivo 
I have refactored it and fixed some related bugs. Please check again.

PR #889, opened 2018-03-11
Enhancement: add option to suppress add edit message

### LINKED ISSUE #888: Feature Request: Add option to disable message dialogs for Git Add window etc
Currently the Git Add windows have a popup dialog to explain the keyboard shortcut for completing the GitAdd command.

After you've used the plugin for long enough these have served their purpose.

Having the ability to disable them would be useful.

randy3k:
close #888

### DIFF
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -367,5 +367,11 @@
         When set to `true`, no views will receive the prompt asking to initialize
         Git in the current view's directory when not found.
      */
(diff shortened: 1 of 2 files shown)


randy3k:
Off topic: I actually think that `git: add (edit)` is not clear to explain what it does. Maybe we should rename it to something like `git: edit diff and stage`?

stoivo:
👍  for rename

randy3k:
close in favor of #892

PR #899, opened 2018-03-19
Appveyor pip install fix

asfaltboy:
Just a quick PR to see if pip install in appveyor/windows can be easily fixed.


randy3k:
It does

PR #902, opened 2018-03-20
Fix: make log panel more asynchronous

### LINKED ISSUE #901: live_panel_output sometimes is out of sync
the log panel is still buggy

When I do `review pull request`, the output is out of sync

```
> /usr/local/bin/git fetch https://github.com/ydsdsa/AlignTab.git patch-1

[Done in 0.36s]From https://github.com/ydsdsa/AlignTab
 * branch            patch-1    -> FETCH_HEAD

```

Relevant settings
```
    "show_panel_for": [
        "commit",
        "apply",
        "fetch",
        "pull",
        "push"
    ],
    "live_panel_output": true
```

I have concluded that it is due to a synchronization issue between the io threads and the process thread. https://github.com/divmain/GitSavvy/blob/797f79ba9aa64c94efe2722eae49344b6655a66c/core/git_command.py#L106-L107

I don't have more time to further investigate now.

randy3k:
fix #901

### DIFF
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -65,7 +65,7 @@ class LoggingProcessWrapper(object):
 
     def read_stdout(self):
         try:
-            for line in self.process.stdout:
+            for line in iter(self.process.stdout.readline, b""):
                 self.stdout = self.stdout + line
                 util.log.panel_append(line.decode())
         except IOError as err:
@@ -73,7 +73,7 @@ class LoggingProcessWrapper(object):
 
     def read_stderr(self):
         try:
-            for line in self.process.stderr:
+            for line in iter(self.process.stderr.readline, b""):
                 self.stderr = self.stderr + line
                 util.log.panel_append(line.decode())
         except IOError as err:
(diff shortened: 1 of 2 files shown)


core/git_command.py
```diff
@@ -65,15 +65,15 @@ def __init__(self, process):
 
     def read_stdout(self):
         try:
-            for line in self.process.stdout:
+            for line in iter(self.process.stdout.readline, b""):
```
stoivo: Can you explain what you do here?

randy3k: The original code prints the buffer by bytes and I changed it to print the buffer by lines as what it is supposed to do

stoivo: ok, What are the arguments to iter?

randy3k: https://docs.python.org/3/library/functions.html#iter

stoivo: So this will read until it gets an empty line?

randy3k: No. It will read until the end of the buffer. `readline` will return `b"\n"` for an empty line.

stoivo: ok,

PR #908, opened 2018-03-30
refactoring GitSavvy settings and project-wise settings

randy3k:
this adds support of project-wise settings by refactoring settings related code.

A [new command](https://github.com/divmain/GitSavvy/blob/10b0ef94844eb823930be697a6fda659479e3e67/docs/README.md#editing-settings) `Preference: GitSavvy Project Settings` is also added to allow easier editing.

### DIFF
--- a/common/global_events.py
+++ b/common/global_events.py
@@ -77,10 +77,8 @@ class GsEditProjectSettingsCommand(WindowCommand):
 
         if project_data is None:
             project_data = {}
-        if "settings" not in project_data:
-            project_data["settings"] = {"GitSavvy": {}}
-        if "GitSavvy" not in project_data["settings"]:
-            project_data["settings"]["GitSavvy"] = {}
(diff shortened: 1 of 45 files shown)


common/global_events.py
```diff
@@ -61,3 +61,28 @@ class GsEditSettingsCommand(WindowCommand):
     """
     def run(self, **kwargs):
         self.window.run_command("edit_settings", kwargs)
+
```
stoivo: This should have been in the later commit but I ¯\_(ツ)_/¯

stoivo:
Looking good

randy3k:
Though I have been using this branch for a while, I am still not 100% nothing breaks. However, let's merge it and test.

common/global_events.py
randy3k: Ya, maybe.

PR #910, opened 2018-04-04
Pass startupinfo to `subprocess.check_output`

### LINKED ISSUE #909: Windows cmd.exe console pops up
Hi 👋 

the most recent version (checked via a fresh install) pops up the famous and annoying Windows console window. This only happens on startup, right after that GitSavvy shows the current branch etc in the status bar. 

I could not get a log, bc it is a startup problem I could not figure out how to enable actual logging fast enough. 

I saw that you don't set `startupinfo.wShowWindow = subprocess.SW_HIDE`. 

https://github.com/divmain/GitSavvy/blob/d7fd1b33e798e3873f384c33f4fed2c8b40fac92/core/git_command.py#L180-L182

I always thought that this is necessary when you set STARTF_USESHOWWINDOW. (I always read that as *I* use SHOWWINDOW so *I* must also provide it.)

On the other hand, that's probably super old code.

kaste:
Fixes #909

### DIFF
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -303,7 +303,13 @@ class GitCommand(StatusMixin,
                 git_path = shutil.which("git")
 
             try:
-                stdout = subprocess.check_output([git_path, "--version"]).decode("utf-8")
+                startupinfo = None
+                if os.name == "nt":
+                    startupinfo = subprocess.STARTUPINFO()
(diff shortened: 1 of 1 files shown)


randy3k:
Thanks.

randy3k:
I’ve merged it into our dev channel.

PR #911, opened 2018-04-04
Release/v2.17.1

randy3k:
It contains bug fixes for Windows, including the annoying bug of #909.

Let's make a quick release of this patch.

### DIFF
--- a/Default (Linux).sublime-keymap
+++ b/Default (Linux).sublime-keymap
@@ -48,15 +48,15 @@
     },
 
     //////////////////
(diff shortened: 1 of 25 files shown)


randy3k:
@stoivo 

You have tagged `v2.17.1` rather than our convention `2.17.1` and I am not sure if Package Control is smart enough to notice the difference. So I have re-tag the release as `2.17.1`.

stoivo:
Sorry, thanks

PR #915, opened 2018-04-07
checkout current file at commit

### LINKED ISSUE #914: Checkout current file commit
Like git blame current file, git show current file with choice commit

randy3k:
close #914 

1. added `git: checkout current file at commit` in the command palette
2. added a new action `checkout current file at commit` in log panel for the current file.

### DIFF
--- a/core/commands/checkout.py
+++ b/core/commands/checkout.py
@@ -36,7 +36,7 @@ class GsCheckoutBranchCommand(WindowCommand, GitCommand):
 
         self.git("checkout", branch)
         self.window.status_message("Checked out `{}` branch.".format(branch))
-        util.view.refresh_gitsavvy(self.window.active_view(), refresh_sidebar=True)
+        util.view.refresh_gitsavvy_interfaces(self.window, refresh_sidebar=True)
 
 
 class GsCheckoutNewBranchCommand(WindowCommand, GitCommand):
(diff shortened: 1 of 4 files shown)


randy3k:
Actually, shall we deprecate `git: checkout current file` and replace it with `git: checkout current file at commit`?

asfaltboy:
I never used it myself...  from the name it sounds like it would "checkout the file at current HEAD (i.e reset staged changes)".
Does `git: checkout current file` offer any other functionality other than `git: checkout current file at commit` ?

asfaltboy:
Agreed then, if anything leaving both is just confusing.

core/commands/checkout.py
```diff
@@ -129,3 +130,21 @@ def run(self):
...
+    """
+
+    def run(self):
+        if self.file_path:
+            super().run(file_path=self.file_path)
+
+    def do_action(self, commit_hash, **kwargs):
+        if commit_hash:
+            self.checkout_ref(commit_hash, self.file_path)
+            self.window.status_message("Successfully checked out {} from head.".format(self.file_path))
```
asfaltboy: Should we maybe mention that we checked out from {commit_hash} rather than from "head" ?

core/git_mixins/checkout_discard.py
```diff
@@ -23,8 +23,11 @@ def checkout_file(self, fpath):
         """
         self.git("checkout", "--", fpath)
 
-    def checkout_ref(self, ref):
+    def checkout_ref(self, ref, fpath=None):
```
asfaltboy: Maybe we should decorate this with the `destructive` decorator? E.g:

    @util.actions.destructive(description="discard uncommitted changes to file")
    def checkout_ref(self, ref, fpath=None):
        ...

And, we could also do it with the above couple of functions...

core/commands/checkout.py
randy3k: You are right.

core/git_mixins/checkout_discard.py
randy3k: It makes sense.

randy3k:
I've removed the original `git: checkout current file` and added a confirmation dialog.

PR #922, opened 2018-04-09
Fix: Pedantic commit checks allows comment on 2. line

### LINKED ISSUE #920: Commit syntax has weird highlighting
I see weird violet highlight of first comment line when committing. It shows up when there is something written on first (empty) line and there is no line break between it and the comment.

See the image:
git-commit-highlight

S3 b3161
git version 2.15.1 (Apple Git-101)
macOs 10.13

BTW. New S3 builds include syntax for commit messages and other git files. I wonder if it makes sense to use those instead.

stoivo:
fix #920

Should we patch this on master and release a patch release?

### DIFF
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -217,7 +217,8 @@ class GsPedanticEnforceEventListener(EventListener):
                 sublime.Region(first_line.a + self.first_line_limit + self.warning_length, first_line.b))
 
         # Add second line to illegal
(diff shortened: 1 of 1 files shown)


randy3k:
If you want to release a patch release, we should merge it to master.

stoivo:
Do we need to add a message? I dont think so

randy3k:
Not necessary strictly speaking, but it is not hard to add one either.

stoivo:
I will not add one since if someone dumps two versions it won't show up the first one.

PR #923, opened 2018-04-09
Enhancement: Opening commit from commadline respect setting

### LINKED ISSUE #921: use_syntax_for_commit_editmsg setting to use native syntax doesn't work in some cases
`use_syntax_for_commit_editmsg` setting, to use sublime's built-in syntax for commit messages, doesn't work in case when commit file is opened from external editor (from command line for example using  `git commit`).

I see in the code that there is GitCommit command that switches to native syntax but that command is obviously not trigged in that case (because commit is triggered externally). There is also a load listener which still applies gitsavvy commit syntax, without considering `use_syntax_for_commit_editmsg` setting. That seems to be the bug.

I expect that when `use_syntax_for_commit_editmsg` is enabled, native syntax will be used for git commit files, regardless of how commit file is opened.

Sublime 3 b3161
git version 2.15.1 (Apple Git-101)
macOs 10.13

stoivo:
Respect use_syntax_for_commit_editmsg
fix #921

### DIFF
--- a/common/global_events.py
+++ b/common/global_events.py
@@ -2,6 +2,7 @@ import sublime
 from sublime_plugin import EventListener, WindowCommand
 
 from . import util
+from ..core.settings import SettingsMixin
 
 
 class GsInterfaceFocusEventListener(EventListener):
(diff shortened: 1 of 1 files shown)


randy3k:
I think we should put it under `GitCommandFromTerminal.on_load`

common/global_events.py
```diff
@@ -26,8 +27,12 @@ def on_close(self, view):
...
-class GitCommandFromTerminal(EventListener):
+class GitCommandFromTerminal(EventListener, SettingsMixin):
     def on_load(self, view):
+        if self.savvy_settings.get("use_syntax_for_commit_editmsg"):
+            syntax_file = util.file.get_syntax_for_file("COMMIT_EDITMSG")
+            git_view_syntax['COMMIT_EDITMSG'] = syntax_file
+
         if view.file_name():
             name = view.file_name().split("/")[-1]
             if name in git_view_syntax.keys():
```
randy3k: I was thinking something like

```python
if name in git_view_syntax.keys():
    if name == "COMMIT_EDITMSG" and self.savvy_settings.get("use_syntax_for_commit_editmsg"):
        syntax_file = util.file.get_syntax_for_file("COMMIT_EDITMSG")
    else:
        syntax_file = git_view_syntax[name]

    view.set_syntax_file(syntax_file)
    view.settings().set("git_savvy.{}_view".format(name), True)
    view.set_scratch(True)
```

stoivo: ok

stoivo:
\o/

Issue #934, opened 2018-04-27
On startup in-memory `instances` cache can be empty


kaste:
If you startup Sublime, obviously all in-memory state is empty. But you can
still have an actual interface view open from the last session.

This will throw as follows 

```
Traceback (most recent call last):
  File "C:\Dev\Sublime Text 3\sublime_plugin.py", line 797, in run_
    return self.run()
  File "C:\Users\c-flo\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\interfaces\status.py", line 19, in run
    StatusInterface(repo_path=self.repo_path)
  File "C:\Users\c-flo\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\common\ui.py", line 40, in __new__
    return interfaces[view.id()]
KeyError: 62
```

https://github.com/divmain/GitSavvy/blob/1196f7819ebbc2b8de71222481c23ea58b4553ec/common/ui.py#L28-L41

Basically the code (subtle factory) will find the right interface view, but this interface is not registered in `interfaces`.

PR #935, opened 2018-04-27
Fix: On startup in-memory `instances` cache can be empty

kaste:
If you startup Sublime, obviously all in-memory state is empty. But you can
still have an actual interface open from the last session.

We simply guard the singleton/factory against `KeyError` and instantiate
a new interface object bound to an already present interface view. This,
of course, is very subtle and implicit programming style. But at least we don't 
change the caller side.

### DIFF
--- a/common/ui.py
+++ b/common/ui.py
@@ -31,13 +31,17 @@ class Interface():
         Search for intended interface in active window - if found, bring it
         to focus and return it instead of creating a new interface.
         """
-        window = sublime.active_window()
-        for view in window.views():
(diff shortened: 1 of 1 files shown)


randy3k:
LGTM, I didn’t test it though

stoivo:
Thanks for the PR, I have been testing this for a week and no issues, so I will merge it. I'll get back to you if there are any issues.

stoivo:
This fixes #934?

kaste:
Yep.

PR #939, opened 2018-05-07
Feature: URL bitbucket integration

asfaltboy:


randy3k:
Actually, it is quite confusing to see all three commands `github/gitlab/bitbucket: open file on remote`. Shall we combine them? Say, we could guess the domain from the remote address (assume that it is not an enterprise repo). If the domain cannot be determined, than a panel is shown asking user to choose the remote type (github/gitlab/bitbucket).

asfaltboy:
@randy3k yes. I was also thinking to perhaps we can add a setting that will determine which "remote integration" is enabled. Perhaps using the "github: set remote for integration" as a basis for this. And then document that this is a required step before integration is enabled?

Generally though, the integration code is repeating a lot. We should definitely DRY refactor these, and perhaps move them into a sub-package (e.g `providers`) instead of having them all in root.

asfaltboy:
Agree, it's more common and useful to use e.g open URL on remote immediately when detected.

We could guess it and, fallback to prompt in case of cases like multiple remotes set up or if we have another issue with detecting (e.g custom self-hosted domains).

kaste:
(Love the doctest thing. Will probably copy that to SublimeLinter.)

asfaltboy:
@randy3k can we merge this as is, and I'll work on repo/provider detection in a separate PR? Reason is selfish: I need this feature personally, so I keep rebasing my branch every time our upstream has cool new stuff :)

stoivo:
I think we can merge it. I did't read the code too carefully. Could you rebase on dev push again so CI passes?

asfaltboy:
I think my fancy tests broke the MacOS tests :/ any ideas?

I think it happens locally too, but don't have time to look now (just moved flats, too many 📦📦📦es!)

PR #947, opened 2018-05-30
Fix subprocess command crashing with OSError

joerchan:
subprocess.check_output crashes with message:
OSError: [WinError 6] The handle is invalid
The issue seems to be with stderr

I'm not very familiar with python or subprocess, I only tested a couple of things and this seems to work.


stoivo:
I rebased your branch on dev. 
Changes look good if it fixes your issue.

stoivo:
@joerchan, thanks for your contribution.

PR #948, opened 2018-05-30
Fix: forward stderr to stdout

randy3k:
`git --version` prints to stderr on windows


randy3k:
Didn't know that it was fixed.

PR #956, opened 2018-06-18
Show reverse diff from commit to workspace

### LINKED ISSUE #955: Show reverse diff for 'checkout file at commit'
For the new command `GsCheckoutCurrentFileAtCommitCommand` you currently show the standard commit info in the bottom panel. I think it would be nicer to show the reverse patch here. 

This is esp. obvious comparing HEAD to workspace. E.g. you would see which lines will be removed if you checkout that commit. (Basically `git diff -R <commit> -- <file>`.) Without the '-R' it's a bit confusing. 

Would you consider a PR here?

kaste:
Closes #955

### DIFF
--- a/core/commands/checkout.py
+++ b/core/commands/checkout.py
@@ -136,7 +136,7 @@ class GsCheckoutCurrentFileAtCommitCommand(LogMixin, WindowCommand, GitCommand):
         if commit:
             self.window.run_command('gs_show_file_diff', {
                 'commit_hash': commit,
-                'fpath': self.file_path
+                'file_path': self.file_path
             })
 
     @util.actions.destructive(description="discard uncommitted changes to file")
@@ -153,9 +153,9 @@ class GsCheckoutCurrentFileAtCommitCommand(LogMixin, WindowCommand, GitCommand):
 
 
 class GsShowFileDiffCommand(WindowCommand, GitCommand):
-    def run(self, commit_hash, fpath):
+    def run(self, commit_hash, file_path):
         self._commit_hash = commit_hash
-        self._fpath = fpath
+        self._file_path = file_path
         sublime.set_timeout_async(self.run_async)
 
     def run_async(self):
@@ -165,7 +165,7 @@ class GsShowFileDiffCommand(WindowCommand, GitCommand):
             "-R",
             self._commit_hash,
             '--',
-            self._fpath
+            self._file_path
         )
 
         output_view = self.window.create_output_panel("show_file_diff")
--- a/core/commands/checkout.py
+++ b/core/commands/checkout.py
@@ -126,9 +126,18 @@ class GsCheckoutCurrentFileAtCommitCommand(LogMixin, WindowCommand, GitCommand):
     Reset the current active file to a given commit.
     """
 
+    show_commit_info = False
+
     def run(self):
         if self.file_path:
             super().run(file_path=self.file_path)
+
+    def on_highlight(self, commit):
+        if commit:
+            self.window.run_command('gs_show_file_diff', {
+                'commit_hash': commit,
+                'fpath': self.file_path
+            })
 
     @util.actions.destructive(description="discard uncommitted changes to file")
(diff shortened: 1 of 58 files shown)


core/commands/checkout.py
```diff
@@ -126,10 +126,19 @@ class GsCheckoutCurrentFileAtCommitCommand(LogMixin, WindowCommand, GitCommand):
     Reset the current active file to a given commit.
     """
 
+    show_commit_info = False
```
stoivo: What is this for?

core/commands/checkout.py
```diff
@@ -126,10 +126,19 @@ class GsCheckoutCurrentFileAtCommitCommand(LogMixin, WindowCommand, GitCommand):
...
+
     def run(self):
         if self.file_path:
             super().run(file_path=self.file_path)
 
+    def on_highlight(self, commit):
+        if commit:
+            self.window.run_command('gs_show_file_diff', {
+                'commit_hash': commit,
+                'fpath': self.file_path
```
stoivo: use file_name

stoivo:
Thanks for the PR.
Can you check out my comments
I changes the base branch to dev and now you need to rebase :) Thanks

core/commands/checkout.py
kaste: LOL? It's your code.

If you don't set `show_commit_info` to `False` it will run a default action on highlight, namely `gs_show_commit_info`.

https://github.com/divmain/GitSavvy/blob/92bc8bfcd3eaff93a17a7bb3480e106bfd802e17/core/ui_mixins/quick_panel.py#L514-L526

core/commands/checkout.py
kaste: Beside, that I don't know which thing I should rename here. You use `file_path` throughout your code to refer to full paths, whereby `file_name` refers to a naked filename without its path. What I mean is, I did not invent this name here and I get `self.file_path` from the inherited code.

kaste:
Rebased

core/commands/checkout.py
stoivo: haha, just checking you :D 
Apparently isn't obvious why so refactored it a little for you (and me in 3 months)
https://github.com/divmain/GitSavvy/pull/957

core/commands/checkout.py
stoivo: Ops. I meant `file_path` instead of fpath.

core/commands/checkout.py
stoivo: Now it's not needed anymore. All you need to do is to overite on_highlight

kaste: No, that doesn't work. Now `on_highlight` is never called unless I implement `run_async` and pass my `on_highlight` down. Apparently #957 is not a refactoring that makes the program flow any clearer.

kaste:
To be very clear here @stoivo @randy3k I did the contribution, and I don't want to rebase/merge for stuff that gets actually later in. I certainly don't do the who gets into master first wins and the other guy can merge/rebase ad infinitum. That's a bad move in any team. And actually you could have tested your #957 against my feature.

randy3k:
@kaste with #957 it is not necessary to set `show_commit_info` to False if you override `on_highlight`.

randy3k:
Also, you need to respect the setting [log_show_more_commit_info](https://github.com/divmain/GitSavvy/blob/master/GitSavvy.sublime-settings#L228)

randy3k:
@kaste sorry about it, we'll fix it a.s.a.p.

cc @stoivo

randy3k:
@kaste I have pushed some commits to your PR which should work with the current `dev`. Actually, we could refactor `GsShowFileDiffCommand` into something more useful for the future. But I guess we could do the refactorization later.

randy3k:
#958 has merged, this should be ready to go too.

randy3k:
Just pulled and tested it locally. Merge now.

PR #957, opened 2018-06-20
Internal: Refactor away show_commit_info from LogPanel

stoivo:
It is a lot more convenient to just set on_highlight.

When you need a different on_highligh you just overwrite the current one. Either
by a new call or as a kwargs to show_log_panel

For #956 to go easier

### DIFF
--- a/core/commands/log.py
+++ b/core/commands/log.py
@@ -21,7 +21,6 @@ class LogMixin(object):
     but the subclass must also inherit fro GitCommand (for the `git()` method)
     """
 
-    show_commit_info = True
     selected_index = 0
 
     def run(self, *args, file_path=None, **kwargs):
@@ -32,13 +31,8 @@ class LogMixin(object):
         show_log_panel(
             self.log_generator(file_path=file_path, follow=follow, **kwargs),
             lambda commit: self.on_done(commit, file_path=file_path, **kwargs),
-            selected_index=self.selected_index,
-            on_highlight=self.on_highlight,
-            show_commit_info=self.show_commit_info
+            selected_index=self.selected_index
         )
-
-    def on_highlight(self, commit):
-        pass
 
     def on_done(self, commit, **kwargs):
         if commit:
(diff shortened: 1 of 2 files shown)


stoivo:
@randy3k @asfaltboy

randy3k:
In this way, we cannot disable `show_commit_info` by passing `on_highlight=None`.

stoivo:
You would need to give it a callable like `on_highlight=lambda c: None`

randy3k:
Maybe we could change line 426 to detect if kwargs https://github.com/divmain/GitSavvy/blob/c67bf8cc6ad981d5bd7e8f144f2dec4752f62177/core/ui_mixins/quick_panel.py#L360
has `on_highlight`

randy3k:
Your are probably right, but a more straightforward solution is to check `kwargs`.
```
if "highlight" is not in kwargs:
    # default highlight
else:
    if kwargs["highlight"]:
        # use this
    else:
        # disable highlight
```

stoivo:
What about 
```
def show_log_panel(entries, on_done, **kwargs):
    """
    Display log entries in quick panel with pagination, and execute on_done(commit)
    when item is selected. `entries` can be either a list or a generator of LogEnty.

    """
    _kwargs = {}
    for option in ['selected_index', 'limit']:
        if option in kwargs:
            _kwargs[option] = kwargs[option]

    for option in ['on_highlight']:
        if option in kwargs:
            _kwargs[option] = kwargs[option]
            if not callable(kwargs[option]):
                _kwargs[option] = lambda commit: None

    lp = LogPanel(entries, on_done, **_kwargs)
    lp.show()
    return lp

```

randy3k:
it also works (but the second for loop has only one item so it is not necessary)

stoivo:
I will commit is like 
```python
def show_log_panel(entries, on_done, **kwargs):
    """
    Display log entries in quick panel with pagination, and execute on_done(commit)
    when item is selected. `entries` can be either a list or a generator of LogEnty.

    """
    _kwargs = {}
    for option in ['selected_index', 'limit']:
        if option in kwargs:
            _kwargs[option] = kwargs[option]

    if not callable(kwargs['on_highlight']):
        _kwargs[option] = lambda commit: None

    lp = LogPanel(entries, on_done, **_kwargs)
    lp.show()
    return lp

```

PR #958, opened 2018-06-26
Fix: expose on_highlight to LogMixin

randy3k:
This should address the issue of #956 

@kaste @stoivo

### DIFF
--- a/core/commands/log.py
+++ b/core/commands/log.py
@@ -31,12 +31,26 @@ class LogMixin(object):
         show_log_panel(
             self.log_generator(file_path=file_path, follow=follow, **kwargs),
             lambda commit: self.on_done(commit, file_path=file_path, **kwargs),
-            selected_index=self.selected_index
+            selected_index=self.selected_index,
+            on_highlight=self.on_highlight
         )
 
     def on_done(self, commit, **kwargs):
+        sublime.active_window().run_command("hide_panel", {"panel": "output.show_commit_info"})
         if commit:
             self.do_action(commit, **kwargs)
+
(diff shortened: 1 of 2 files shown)


stoivo:
I think it should be done like 07d701fb029cda7742b2cdb3ad6a40bf86146679, What do you think?

I open a new PR #960 with the sum. I think that the cleanser solution. And this time I have tested that it works. 
cai3fek3mh

randy3k:
I like my code a little bit better because it allows user to override LogMaxin and use `super().on_highlight` to invoke the original `gs_show_commit_info`.

stoivo:
The original gs_show_commit_info  https://github.com/divmain/GitSavvy/pull/958/files#diff-6809e629d1384bcdee2aa618a5425cbfL517 does not do a lot so I think is nicer to call the command directly then calling super.

Let leave it for 24 hours, and see what we think then.

kaste:
@randy3k's version is better. You just need to count the `if`s. Basically it uses standard inheritance for late dispatch. It also simplifies the factory `show_log_panel`.

stoivo:
> In this way, we cannot disable show_commit_info by passing on_highlight=None because it is only overwriten if it is non-None
> 
> https://github.com/divmain/GitSavvy/pull/957#issuecomment-398650429

Here you cant call on_highlight with a none callable? That should be allowed to dissable any 
highlight

randy3k:
There is no default highlight anymore. The show commit info highlight has moved to LogMixin.

randy3k:
@stoivo do you feel better about this PR now?

Issue #954, opened 2018-06-15
Edit Project Settings is 'hostile'


kaste:
In the new GsEditProjectSettingsCommand you persist the project data like so

https://github.com/divmain/GitSavvy/blob/9ec1cc44455a11e1dfef89629fc7c4e8c01f6a57/common/global_events.py#L88

This is 'hostile' bc e.g. it removes all the comments from that file. (Essentially it rewrites/formats the file.) In my opinion, setting files in Sublime are not data files, but user documents, so you really shouldn't do that.

(Since I'm coming from SublimeLinter land: I noticed that you expect a top-level key 'GitSavvy' here. SublimeLinter will move away from that, and will instead expect flat key-value pairs like so 

```
settings:
	SublimeLinter.linters.flake8.disable: true
```  

We do this essentially in a #usetheplatform spirit. I _really_ would like a move in the Sublime community here, and would appreciate if GitSavvy would do the same. E.g. `GitSavvy.name_of_the_setting`. The namespacing here is important for automatic setting documentation and discovery.)

stoivo:
The Default settings, should those be prefixed with `GitSavvt` too?

It makes sense. We just recently introduce Project settings.

kaste:
Consider this

```
sublime.load_settings('GitSavvy.sublime-settings').get('git_path')
                       ^^^^^^^^ namespace
                                                        ^^^^^^^^ key

view.settings().get('GitSavvy.git_path')
                     ^^^^^^^^ namespace
                              ^^^^^^^^ key
```

For the first case, you're always namespaced bc Sublime merges all the settings from files with the same name. 

But, if you access the settings through a context object (the view), you're in the global namespace bc Sublime starts here with the generic 'Preferences' file. Since we now are in a public, shared key space, we must namespace our keys.

randy3k:
@kaste The usage of the `top` level key is actually inspired by `SublimeLinter`. I am happy to hear that `SublimeLinter` is moving to flat key-value pairs. However, for `GitSavvy`, I do prefer something like

```
	"settings":
	{
		"GitSavvy": {
                           # blablabla
                }
	}
```

It is because we essentially support every setting which could be specified in the `GitSavvy.sublime-settings`. It would be much easier for users to copy the keys from that file to their project file.

Regarding `set_project_data`, there are several command actually would change `sublime-project` files, e.g., the `Remove folder from project` and `Project: Add Folder` . I don't think it is a good hobby for any users to put comments into them.

[...]

PR #959, opened 2018-06-26
set_project_data only when key does not exist

randy3k:
partially address #954

### DIFF
--- a/common/global_events.py
+++ b/common/global_events.py
@@ -84,8 +84,7 @@ class GsEditProjectSettingsCommand(WindowCommand):
             project_data = {}
         if "GitSavvy" not in project_data:
             project_data["GitSavvy"] = {}
-
-        self.window.set_project_data(project_data)
+            self.window.set_project_data(project_data)
 
         sublime.set_timeout(lambda: self.window.run_command("edit_settings", {
             "user_file": project_file_name,


core/settings.py
```diff
@@ -7,11 +7,19 @@ def __init__(self, parent=None):
         self.global_settings = sublime.load_settings("GitSavvy.sublime-settings")
 
     def get(self, key, default=None):
-        project_data = sublime.active_window().project_data()
+        view = sublime.active_window()
+        project_savvy_settings = view.settings().get("GitSavvy", {})
```
stoivo: This change. Where do we set these type of settings?

stoivo:
You just need to explain something to me :)

core/settings.py
randy3k: From the `sublime-project` file. The settings in `sublime-project` will be imported to `view.settings()` when the file is loaded.

stoivo: Is that something we make of my default?

stoivo: What i meat was. Do we copy the project settings to the view every time a new view opens?
 if so where?

randy3k: It is done by SublimeText, not by us.

stoivo: Then you can merge it.

stoivo: All good

randy3k:
as discussed in #954 , `GsEditProjectSettingsCommand` will not alter the project file, instead, it now shows a popup of instruction.

randy3k:
@stoivo how do you think about this?

PR #964, opened 2018-07-04
Allow using ~/%HOME% when providing clone path, by expanding user home dir

asfaltboy:
I personally find this more useful than typing the path every time


stoivo:
I haven't tested but I you have

asfaltboy:
I haven't actually, let me try it out

asfaltboy:
Ok @stoivo , it works just fine 👍

PR #969, opened 2018-07-12
Release/2.17.4

randy3k:
Hooray, next release.

PR #970, opened 2018-07-13
Enhancement: better message when prompt_on_abort_commit

randy3k:
Additionally, no prompt will be made if the message is empty and the default
value of prompt_on_abort_commit is set as true


randy3k:
Should this also go to 2.17.4?

stoivo:
Yes, but I stil think the release can wait untill  mid August
https://github.com/divmain/GitSavvy/pull/969#issuecomment-404822829

PR #972, opened 2018-07-27
Fix: use fallback encoding to decode stderr

### LINKED ISSUE #966: git_command.py utf-8 error
I've started getting this error on Sublime Text 3176 (Portable Edition) with git 2.17.1 on Windows 7.

```
'utf-8' codec can't decode byte 0x8a in position 16: invalid start byte

  File "[...]Sublime\Data\Installed Packages\GitSavvy.sublime-package\core/git_command.py", line 215, in git
    stdout, stderr = self.decode_stdout(stdout), stderr.decode()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8a in position 16: invalid start byte
```

The error appears on loading Sublime and prevents using GitSavvy. It was running until recently with no notable changes and updates.

randy3k:
close #966

### DIFF
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -212,7 +212,7 @@ class GitCommand(StatusMixin,
                 stdout, stderr = p.communicate(stdin)
 
             if decode:
-                stdout, stderr = self.decode_stdout(stdout), stderr.decode()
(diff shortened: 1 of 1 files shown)


stoivo:
We should probably release this sooner since is it is a bugfix an not a improvement.

stoivo:
I am on vacation

PR #982, opened 2018-08-12
Enhancement: honor prepare_commit_msg and pre_commit hooks

### LINKED ISSUE #161: Properly support git-hooks or (at the very least) document which ones we do not support and why
Here is a list of the basic (non email patch) git hooks and whether or not we currently support them.

* [x] pre-commit
* [ ] prepare-commit-msg
* [x] commit-msg
* [x] post-commit
* [ ] pre-rebase
* [x] post-rewrite
* [x] post-checkout
* [x] post-merge
* [x] pre-push

If we kick off rebases using the `git rebase` command as I suggested then that would automatically add support for `pre-rebase` at that time.  That would leave only `prepare-commit-msg` as unsupported.  Rather than have all the hooks work except for one I'd love to see if we could support this.  It seems as simple as asking git to furnish the prepared commit msg (thereby running the `prepare-commit-msg` automatically in the process) and then plugging that into our commit dashboard work-flow.

It would also remove some of the hand coded variants we have currently contrived to work around the fact that we simply aren't leaning on git for it's own default behavior:

- having to figure out diff HEAD^ vs diff -cached for amend.  Asking git gives us that for free.
- having to manually query the previous commit msg for amend. Asking git gives us that for free.
- having to manually query and add sign-off info. Asking git gives us that for free.

Please consider supporting these hooks.  It just feels like when we're rebuilding basic git functionality that we're doing something wrong. I'm even game to do the legwork on `prepare-commit-msg` as it fits right in to the other commit workflow stuff I've been working on.

### LINKED ISSUE #980: Prevent opening commit window when there is nothing to commit
Is there an option to prevent commit message window from opening when there is nothing to commit? Like if you have not staged anything yet, but write `git commit`, it should not open window.

That is standard git behavior. 
Or is this intended with GitSavvy to not check if there is anything to commit or not.

I mean you can think that you staged something, then write commit message and try to commit it, but only then you will know there was nothing to commit in a first place, which could be misleading.

randy3k:
use the idea of https://github.com/SublimeGit/SublimeGit/issues/73#issuecomment-292795958 to honor the `prepare_commit_msg` and `pre_commit` hooks

close #161

also check "nothing to commit"

close #980

### DIFF
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -56,6 +56,10 @@ class GsCommitCommand(WindowCommand, GitCommand):
     def run_async(self, repo_path=None, include_unstaged=False, amend=False):
 
         repo_path = repo_path or self.repo_path
+
+        # check "nothing to commit"
+        self.dry_run_commit(repo_path, include_unstaged, amend)
+
         # run `pre-commit` and `prepare-commit-msg` hooks
         hooks_path = os.path.join(repo_path, ".git", "hooks")
         pre_commit = os.path.join(hooks_path, "pre-commit")
@@ -114,6 +118,16 @@ class GsCommitCommand(WindowCommand, GitCommand):
                 pass
             else:
                 raise GitSavvyError(e.args[0])
+
+    def dry_run_commit(self, repo_path, include_unstaged, amend):
+        show_panel_overrides = self.savvy_settings.get("show_panel_for")
+        self.git(
+            "commit",
+            "-q" if "commit" not in show_panel_overrides else None,
+            "-a" if include_unstaged else None,
+            "--amend" if amend else None,
+            "--dry-run"
+        )
 
 
 class GsCommitInitializeViewCommand(TextCommand, GitCommand):
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -7,6 +7,7 @@ from sublime_plugin import EventListener
 from ..git_command import GitCommand
 from ...common import util
 from ...core.settings import SettingsMixin
+from ..exceptions import GitSavvyError
 
 
 COMMIT_HELP_TEXT_EXTRA = """##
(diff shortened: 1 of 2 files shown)


oerp-odoo:
@randy3k seems to be working. It tiggers both pre-hook and warning about nothing to commit (though have not tested prepare-commit-msg).

oerp-odoo:
Don't know if its related with this fix, but I noticed getting this error, now when I open commit window:

```
> /usr/bin/git commit
error: There was a problem with the editor 'false'.
Please supply the message using either -m or -F option.
```
Though, I can proceed and commit successfully, even with this error.

randy3k:
Did you add `commit` to the `show_panel_for` settings?

oerp-odoo:
Yes. My custom settings are currently these:
```
{
    /*
        Add entries to this array (e.g. "pull" or "push") if you'd like the
        output of these Git commands to always be shown in a panel.
     */
    "show_panel_for": ["commit", "push", "pull"],
    "pedantic_commit_message_line_length": 72,
}
```

oerp-odoo:
So is this expected or should I change some settings?

randy3k:
I'd suggest removing `commit` from `show_panel_for` for now.

In fact, I don't understand why the global settings `show_panel_for` would override the `show_panel` setting of [`git`](https://github.com/divmain/GitSavvy/blob/79870b8ad4c70f393b7353eda79676e041d96521/core/git_command.py#L134-L134). So a possible solution is to allow `show_panel` to override the global setting instead of the other way round.

oerp-odoo:
Don't know if this is really related with this specific change, but it seems amending commit does not work (at least when using commit on close).

I have these settings in my key-binds (I disabled ctrl+s from commiting, because I like commit to be done on close only):
```
    // COMMIT VIEW (git savvy)
    {
        "keys": ["ctrl+s"],
        "command": "dummy",
        "context": [
            { "key": "setting.command_mode", "operator": "equal", "operand": false },
            { "key": "setting.git_savvy.get_long_text_view", "operator": "equal", "operand": true },
            { "key": "setting.git_savvy.commit_on_close", "operator": "equal", "operand": true }
        ]
    },
    {
        "keys": ["ctrl+enter"],
        "command": "gs_commit_view_sign",
        "context": [
            { "key": "setting.command_mode", "operator": "equal", "operand": false },
            { "key": "setting.git_savvy.get_long_text_view", "operator": "equal", "operand": true }
        ]
    },
```
And then I have these custom settings for gittsavy:
```
{
    /*
        Add entries to this array (e.g. "pull" or "push") if you'd like the
        output of these Git commands to always be shown in a panel.
     */
    "show_panel_for": ["commit", "push", "pull"],
    "pedantic_commit_message_line_length": 72,
    "show_stdin_in_output": true,
    "commit_on_close": true,
}
```
As you can see from video, amend triggers only if you do not modify anything. It does show in panel that I recommitted (well same thing). Now If I try to actually amend commit message, it ignores it and message is left unchanged. Though normal commit works (if I'm not amending, but do committing new one).

[gittsavy-commit-amend.ogv.zip](https://github.com/divmain/GitSavvy/files/2292951/gittsavy-commit-amend.ogv.zip)


P.S. Do note I still use `commit` for `show_panel_for` even though you recommended to not use it. But then I do not see any information regarding commit (when it is opened, when it is closed, etc). But I tested with that disabled, amend does not work with or without that.

randy3k:
That bug is not related to this PR. For some reasons, you need to keep the last line empty in order for `amend` to work.

randy3k:
Oh, I was wrong. It is this PR specific.

oerp-odoo:
You mean last line of commit message? When I started using gittsavy, I noticed that on default it opens commit window with two empty lines, instead of one. It looked odd, but I did not know that it is because of some limitation.

If it is hard to fix it (that you would not need that extra line), maybe it is a good idea to write some comment on commit, warning to leave that line (like there are comments explaining how to sign off etc)?

P.S. I guess that empty line is not committed? It is just ignored?

randy3k:
I am looking into it. It is definitely "fixable".

randy3k:
There are two sepearte issues.

1. It seems only affect merge commit. Regular commits can be amended correctly.

2. [This piece](https://github.com/divmain/GitSavvy/blob/dd7839ecb41f31180e120eb6df46cd03517f35da/core/commands/commit.py#L155-L166) of code needs to be reenginnered.

randy3k:
I don't have time to fix it now, will further look into it tonight.

oerp-odoo:
Cool. Thanks

randy3k:
#988 should fix the committing problem. And I'll rebase this PR onto #988 and fix the prepare commit message hook.

oerp-odoo:
Bump. When this PR is planned to be released?

randy3k:
@stoivo @asfaltboy free to review?

stoivo:
@randy3k, you can merge this but maybe you want to make a patch release from dev first?

randy3k:
Let's merge it to dev and test it more extensively.

@oerp-odoo The new feature will be available on the dev branch.

randy3k:
Thanks @stoivo 👍

PR #988, opened 2018-08-17
Help text

randy3k:
@oerp-odoo discovered a bug on committing/amending when the last line of the commit message is not empty (when `commit_on_close` is `true`). It is merely due to the fact that `help_text` can a leading "\n". I have also added some conditions for helping to detect the message text.

@stoivo I think we should also merge this to release 2.17.4. Please review it and we might want to delay the release of 2.17.4 for a few days for testing.


stoivo:
Then we wait with releaseing 2.17.4

stoivo:
Should this commit be cherrypicked into dev?

PR #1003, opened 2018-09-04
Clone git submodules

luv:
With this simple change, git submodules are cloned so they can be used right away.

### DIFF
--- a/core/commands/init.py
+++ b/core/commands/init.py
@@ -121,7 +121,7 @@ class GsClone(WindowCommand, GitCommand):
         self.window.status_message("Start cloning {}".format(self.git_url))
         self.git(
             "clone",
-            "--recursive" if self.recursive else "",
+            "--recursive" if self.recursive else None,
             self.git_url,
             self.suggested_git_root,
             working_dir='.')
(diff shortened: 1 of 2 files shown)


luv:
Sounds resonable. I can imagine that changing the current behaviour of git clone might surprise a few users.

What new command name would you recommend for clone with --recurse ?

randy3k:
Maybe `git clone recursively`?

luv:
I have updated the PR to use command "git: clone recursively".

randy3k:
Lgtm

core/commands/init.py
```diff
@@ -118,7 +119,10 @@ def on_enter_directory(self, path):
 
     def do_clone(self):
         self.window.status_message("Start cloning {}".format(self.git_url))
-        self.git("clone", self.git_url, self.suggested_git_root, working_dir='.')
```
stoivo: You can do to dry it up and nicer when we want to add more clone options
```python
self.git(
    "clone", 
    "--recursive" if self.recursive else None
    self.git_url, 
    self.suggested_git_root, 
    working_dir='.')
```

luv: sure

stoivo: Use `else None`, Everything else is good

luv: sure

stoivo:
Thanks

PR #1008, opened 2018-09-16
Enhancement: Add more complex git output scopes to diff syntax

deathaxe:
This commit addresses the following flaws of the diff.sublime-syntax

  1. missing rules for git output:
     - deleted file mode
     - rename from
     - rename to
     - similarity index

  2. the whole git output is scoped as as `punctuation`

  3. the linefeed \n is not included in the git output rules which
     causes bad results if a color scheme wants to highlight the
     background of the git output.

This commit therefore

  1. adds rules for the missing output messages
  2. introduces normal keyword, constant, string scopes for the output
     and encapsulates everything in a meta scope based on the existing
     `meta.diff.header` for easier color-scheme creation.
  3. optionally includes the \n? in any rule.

time.

The goal is to be able to create color schemes with nice looking git diff headers.

screenshot

### DIFF
--- a/syntax/diff.sublime-syntax
+++ b/syntax/diff.sublime-syntax
@@ -16,17 +16,60 @@ first_line_match: |-
       )
 contexts:
   main:
-    - match: ^diff.+$
-      scope: punctuation.definition.command.diff
+    # diff --git a/path/file.ext b/path/file.ext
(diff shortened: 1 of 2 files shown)


stoivo:
Cool @deathaxe

PR #1013, opened 2018-09-20
Fix: check the nullness of project_savvy_settings

### LINKED ISSUE #1007: NoneType is not iterable
- Sublime Text version number
3.1.1 build 3176

- Git version number
git version 2.18.0

- OS type and version
MacOS 10.13.6

- Console error output
Please report this error to GitSavvy:

argument of type 'NoneType' is not iterable

  File "/Users/diek/Library/Application Support/Sublime Text 3/Installed Packages/GitSavvy.sublime-package/core/git_command.py", line 215, in git
    stdout, stderr = self.decode_stdout(stdout), self.decode_stdout(stderr)
  File "/Users/diek/Library/Application Support/Sublime Text 3/Installed Packages/GitSavvy.sublime-package/core/git_command.py", line 266, in decode_stdout
    silent_fallback = self.savvy_settings.get("silent_fallback")
  File "/Users/diek/Library/Application Support/Sublime Text 3/Installed Packages/GitSavvy.sublime-package/core/settings.py", line 14, in get
    if key in project_savvy_settings:
TypeError: argument of type 'NoneType' is not iterable

You told me to report it :) 

- Steps to reproduce, if possible
On startup, and restart of SBT

- If the bug is caused by a failing command, [include a debug log](https://github.com/divmain/GitSavvy/blob/master/docs/debug.md#providing-a-debug-log).

randy3k:
close #1007

### DIFF
--- a/core/settings.py
+++ b/core/settings.py
@@ -9,7 +9,7 @@ class GitSavvySettings:
     def get(self, key, default=None):
         window = sublime.active_window()
         view = window.active_view()
-        project_savvy_settings = view.settings().get("GitSavvy", {})
+        project_savvy_settings = view.settings().get("GitSavvy", {}) or {}
 
         if key in project_savvy_settings:
             return project_savvy_settings[key]


core/settings.py
```diff
@@ -11,7 +11,7 @@ def get(self, key, default=None):
         view = window.active_view()
         project_savvy_settings = view.settings().get("GitSavvy", {})
 
-        if key in project_savvy_settings:
+        if project_savvy_settings and key in project_savvy_settings:
```
asfaltboy: Can also be fixed by doing this on line 12:

```
        project_savvy_settings = view.settings().get("GitSavvy") or {}
```

Not sure which style we prefer

stoivo: I like your suggustion asfaltboy

PR #1014, opened 2018-09-20
Enhancement: cleanup commit message

randy3k:
The cleanup flag controls how a commit message is cleaned up.
https://git-scm.com/docs/git-commit/2.0.0#git-commit---cleanupltmodegt

In command line, the default mode strips out comments started with `#` if an editor is opened interactively. However, GitSavvy only passes the message to the stdin, and the default mode only strips `whitespace`. This PR is trying to reflect the very same behavior on cli.

### DIFF
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -315,6 +315,12 @@
     "confirm_force_push": true,
 
     /*
+        This setting controls how commit messages are cleaned up in the commit view:
+        `git commit --cleanup=<mode>`. It is ignored if `git config commit.cleanup` is set.
+    */
+    "commit_cleanup_default_mode": "strip",
+
+    /*
         When set to `true`, closing the commit message window via keyboard will result
         in a commit action being taken, except in cases where the message is empty.
         The same is also true for amending commits.
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -155,7 +155,7 @@ class GsCommitInitializeViewCommand(TextCommand, GitCommand):
 
         if has_prepare_commit_msg_hook and os.path.exists(commit_editmsg_path):
             with util.file.safe_open(commit_editmsg_path, "r") as f:
-                initial_text = "\n" + f.read().rstrip() + help_text
+                initial_text = f.read().rstrip() + help_text
         elif option_amend:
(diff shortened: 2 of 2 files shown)


randy3k:
But it is the default clean up mode in editor. Perhaps we should use the syntax highlight to reflect this

kaste:
No! Try Git GUI, it will only strip whitespace. If you have a good editor the commit message is taken as-is (only strip).  (They use these modes for other automated things. Maybe getting commit messages out of emails etc. see ✂️)

randy3k:
However, it is the default behaviour if vim is used in git cli. Maybe the git gui does thing differently.

kaste:
`strip` is the default if `-e` is set. So it also applies if you set Sublime as EDITOR as well. Otherwise `whitespace` is used. 

But here with GitSavvy we have an UI. `whitespace` must be the default. Maybe automatically set it to `strip` if a pre commit commit message handler runs (`has_prepare_commit_msg_hook`) _and_ outputs something bc it is likely that such a handler basically adds comments to the commit message to be edited by the user.

This behavior should be reflected in the help text.

randy3k:
> This behavior should be reflected in the help text.
I agree, though I wonder how many users will actually read the text.

randy3k:
I vote for doing strip anyway but use syntax highlight to warn users that the corresponding lines will be stripped. Doing automatically detection will be a pain.

kaste:
I think syntax highlighting wouldn't work bc if the user has set whitespace or even verbatim we can't mark these lines as comments.

stoivo:
What about just substitjutint the comment away before we send it as std in to git commit. Since we know what the comment is. We added it. This will not work if the user changes the comment.

kaste:
This is how we start _by default_ a commit

image

We send everything above our comment as commit message. The *old* behavior was to not further mangle the message the user typed in this 'field'. This is also the behavior in GitGUI, GithubApp, Github Web, VSCode. 

The *new* behavior is to strip comments from this message, and e.g. reduce multiple empty lines to one new line.

To make it clear, this is *not* the default behavior of git, and imho it is not a good default for new and old users.

randy3k:
#1071 is going to revert this

PR #1015, opened 2018-09-21
Bugfix: avoid splitlines in blame

### LINKED ISSUE #1011: Git blame does not always work
- Sublime Text version number: 3.1.1 3176
- Git version number: 2.7.4
- OS type and version: Ubuntu 16.04.5 LTS
- Console error output:
```
  File "/opt/sublime_text/sublime_plugin.py", line 1072, in run_
    return self.run(edit)
  File "/home/oerp/.config/sublime-text-3/Installed Packages/GitSavvy.sublime-package/core/commands/blame.py", line 165, in run
  File "/home/oerp/.config/sublime-text-3/Installed Packages/GitSavvy.sublime-package/core/commands/blame.py", line 217, in get_content
  File "/home/oerp/.config/sublime-text-3/Installed Packages/GitSavvy.sublime-package/core/commands/blame.py", line 260, in parse_blame
AttributeError: 'NoneType' object has no attribute 'groups'
```
On one of the source files, I tried to run git blame, but it just opens empty git blame. By looking at console output, I can see something went wrong. If I run git blame on same file with raw git, it does open it normally.
- Steps to reproduce, if possible
I don't know what is the difference on that specific file, but other files seem to run fine with git blame except the one I found the issue with.
- If the bug is caused by a failing command, [include a debug log](https://github.com/divmain/GitSavvy/blob/master/docs/debug.md#providing-a-debug-log).

[gitsavvy-git-blame-debug.json.zip](https://github.com/divmain/GitSavvy/files/2404222/gitsavvy-git-blame-debug.json.zip)

asfaltboy:
This avoid additional "line break" symbols in source code from being considered as line breaks in our parser. Git seems to only ever `'\n'` as a newline character, but python's `splitlines()` method considers many others by default.

Refs:
https://github.com/python/cpython/blob/3.4/Objects/unicodetype_db.h#L4317-L4338
https://github.com/git/git/blob/master/blame.c#L1633-L1637

Fixes #1011  (more info there)

### DIFF
--- a/core/commands/blame.py
+++ b/core/commands/blame.py
@@ -250,6 +250,9 @@ class GsBlameRefreshCommand(BlameMixin, TextCommand, GitCommand):
         return spacer.join(partitions_with_commits_iter)
 
     def parse_blame(self, blame_porcelain):
+        if blame_porcelain[-1] == '':
+            blame_porcelain = blame_porcelain[:-1]
+
         lines_iter = iter(blame_porcelain)
 
         blamed_lines = []
--- a/core/commands/blame.py
+++ b/core/commands/blame.py
@@ -214,7 +214,7 @@ class GsBlameRefreshCommand(BlameMixin, TextCommand, GitCommand):
             commit_hash, "--", filename_at_commit
         )
         blame_porcelain = unicodedata.normalize('NFC', blame_porcelain)
-        blamed_lines, commits = self.parse_blame(blame_porcelain.splitlines())
+        blamed_lines, commits = self.parse_blame(blame_porcelain.split('\n'))
 
         commit_infos = {
             commit_hash: self.short_commit_info(commit)


Default (OSX).sublime-keymap
```diff
@@ -25,7 +25,6 @@
         "keys": ["super+w"],
         "command": "gs_commit_view_close",
         "context": [
-            { "key": "setting.command_mode", "operator": "equal", "operand": false },
```
randy3k: how about windows/linux keybind?

randy3k:
how about windows/linux keybind?

asfaltboy:
Oops I didn't mean to commit that file here ... shall I remove it or just add the other 2 OS files?

randy3k:
maybe we should remove the keybind file for this PR?

asfaltboy:
I just tested it, this change seems to breaks blame command for all files :/

asfaltboy:
Sorry for the holdup, should be good to go now

Issue #1010, opened 2018-09-18
commit_on_close: false doesn't show prompt if using command mode


asfaltboy:
If we want the prompt shown, then it should show regardless of "vim mode" IMHO.

Please include the following information when filing a bug:

- Steps to reproduce: 1. make change and enter commit view, 2. add some words and exit editing mode (e.g hit <kbd>esc</kbd>), 3. hit "super+w" to close tab, it closes without any prompt.
- Sublime Text version number: dev branch (6ac8f3e)
- Git version number: 2.16
- OS type and version: MacOS 10.13.6
- Console error output: -

I worked around this by modifying my keybindings to remove the `setting.command_mode` condition, here's a sample diff:

```diff
 Default (OSX).sublime-keymap | 1 -
 1 file changed, 1 deletion(-)

diff --git a/Default (OSX).sublime-keymap b/Default (OSX).sublime-keymap
index 2b58686..e86cdef 100644
--- a/Default (OSX).sublime-keymap	
+++ b/Default (OSX).sublime-keymap	
@@ -25,7 +25,6 @@
         "keys": ["super+w"],
         "command": "gs_commit_view_close",
         "context": [
-            { "key": "setting.command_mode", "operator": "equal", "operand": false },
             { "key": "setting.git_savvy.commit_view", "operator": "equal", "operand": true }
         ]
     },
```

Can submit a PR if it makes sense to everyone...

randy3k:
Should we override the `ctrl+w` on Windows/Linux too? I am not a vintage/vintageous user and I can't tell.

stoivo:
Do you usesaly close buffers in vim mode with super+w? if so I think this patch is good

asfaltboy:
@stoivo I alternate between vim modes and occasionally hit super+w without thinking which mode I am in, and I'm in command_mode and it makes me lose my commit message (since the buffer is discardable and our `gs_commit_view_close` command doesn't fire).

@randy3k we should definitely update the other OS keybinding files too.

One other thing that happens to me a lot, is that I would have a commit tab open, and then I find that I have too many tabs, want to declutter to read through the changes again before committing, and I use "close all tabs" , which doesn't launch our `commit_view_close` command once again 😞 , though not sure there's a way to solve this without writing my own close all command that wraps around these two (depending on some information about the tab)

randy3k:
Maybe we could add an eventlisenter to emit `commit_view_close` when the view is closing and `commit_view_close` has not yet be executed.

[...]

PR #1016, opened 2018-09-21
Enhancement: better close commit handler

asfaltboy:
As proposed in #1010 , this commit adds an `on_pre_close` handler that fires whenever sublime closes the commit view (identified by the view setting).

Unfortunately, the handler cannot prevent the view from closing, so this change adds a clone method that is called when we want to keep the commit window open. Moreover, since close also occurs when we actually commit, we need to make sure we allow the close to go through, even if `prompt_on_abort_commit` is turned on. This is achieved by setting `git_savvy.commit_view` to `False` after completing the commit.

Would appreciate if you could review and test this PR, as it's the first time I've written an event handler in sublime.

### DIFF
--- a/Default (Linux).sublime-keymap
+++ b/Default (Linux).sublime-keymap
@@ -21,14 +21,6 @@
             { "key": "setting.git_savvy.get_long_text_view", "operator": "equal", "operand": true }
         ]
     },
-    {
-        "keys": ["ctrl+w"],
-        "command": "gs_commit_view_close",
-        "context": [
-            { "key": "setting.command_mode", "operator": "equal", "operand": false },
-            { "key": "setting.git_savvy.commit_view", "operator": "equal", "operand": true }
-        ]
-    },
 
     ///////////////
     // DIFF VIEW //
--- a/Default (OSX).sublime-keymap
+++ b/Default (OSX).sublime-keymap
@@ -21,14 +21,6 @@
             { "key": "setting.git_savvy.get_long_text_view", "operator": "equal", "operand": true }
         ]
     },
-    {
-        "keys": ["super+w"],
-        "command": "gs_commit_view_close",
-        "context": [
-            { "key": "setting.command_mode", "operator": "equal", "operand": false },
-            { "key": "setting.git_savvy.commit_view", "operator": "equal", "operand": true }
-        ]
-    },
 
     ///////////////
     // DIFF VIEW //
--- a/Default (Windows).sublime-keymap
+++ b/Default (Windows).sublime-keymap
@@ -21,14 +21,6 @@
             { "key": "setting.git_savvy.get_long_text_view", "operator": "equal", "operand": true }
         ]
     },
-    {
-        "keys": ["ctrl+w"],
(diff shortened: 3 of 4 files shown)


randy3k:
I will take a look later today. Thank you for working on it.

randy3k:
Two issues:

1. With `commit_on_close` off; `git: commit` -> hit `c-w` -> click cancel -> hit `c-w` again and no prompt any more.

2. With `commit_on_close` on; `git: commit` -> hit `c-w` -> the view is mysteriously cloned (is it intended?)

asfaltboy:
Thanks @randy3k 

1. I tried to reproduce this but couldn't. Can you confirm your GitSavvy settings are:
```
    "commit_on_close": false,
    "prompt_on_abort_commit": true,
```

2. This is intended, since I couldn't prevent the view from being closed in the event handler; and, `GsCommitViewDoCommitCommand` takes a view. But, that last command should normally close the view. Is there any reason that can't happen in your environment (any errors or other conditions)?

P.s: are you sure you restarted sublime after checking out the branch?

randy3k:
My settings are
```
   "commit_on_close": false,
   "prompt_on_abort_commit": true,
```

A more detail reproduction steps.

1. Make some changes to a file and stage the file
2. `git: commit`
3. type a random message
4. `cmd + w` in the commit view
5. click "cancel" in the display dialogue
6. `cmd + w` again in the commit view
7. dialogue doesn't show and the view is closed

PS: restarted Sublime Text to confirm, I am on MacOS 10.13 if it matters.

asfaltboy:
That's odd, I'm seeing a different result. Sublime 3.1.1 build 3176, MacOS 10.14 (Mojave). Here's a screen capture: https://imgur.com/a/qRUuSoH  and my console shows these message:

```
GS [debug] about to close  <sublime.View object at 0x10d090a10> COMMIT: GitSavvy
GS [debug] Running gs_commit_view_close command in view <sublime.View object at 0x10c75e110> COMMIT: GitSavvy
GS [debug] about to close  <sublime.View object at 0x10d077650> COMMIT: GitSavvy
GS [debug] Running gs_commit_view_close command in view <sublime.View object at 0x10b20f550> COMMIT: GitSavvy
```

I wonder what difference is there more 🤔

randy3k:
Odd, my second `cmd+w` closes the view.

```
>>> sublime.log_commands(True)
command: drag_select {"event": {"button": 1, "x": 626.4921875, "y": 222.390625}}
command: gs_status_commit
command: close
GS [debug] about to close  <sublime.View object at 0x103ba75d0> COMMIT: GitSavvy
GS [debug] Running gs_commit_view_close command in view <sublime.View object at 0x103bc7650> COMMIT: GitSavvy
command: close
```

Update: I just managed to get something different: the second `cmd+w` didn't close the view. I tried again, the third `cmd+w` closed it.

It seems to me that there is a race condition somewhere.

asfaltboy:
I'm going to abort this effort for now as I don't have too much free time to catch up on this. I'll keep the branch around If we ever want to try this again in the future.

Issue #596, opened 2017-02-08
Signing fails via GPG 


franciscolourenco:
Related to #581 but not on windows.

```
`/usr/bin/git commit -q -F -` failed with following output:

gpg: cannot open `/dev/tty': Device not configured
error: gpg failed to sign the data
fatal: failed to write commit object
```

On the terminal it works, and no password is asked. I guess it is cached or handled automatically with 
```
[credential]
	helper = osxkeychain
```

Since no interactin is actually required, is there any configuration possible which allows this to work?

aristidesfl:
Any solutions for this? Right now is not possible to use Gitsavvy with signed commits at all.

stoivo:
@aristidesfl, sorry for not commenting anything. 
@asfaltboy @randy3k any experience here? I dont.

divmain:
@aristidesfl have you seen the docs for [signing commits with GPG](https://github.com/divmain/GitSavvy/blob/master/docs/commit.md#signing-your-commits-with-gpg)?  This method should work on Windows.

aristidesfl:
@divmain Hadn't seen them and I was missing 
 ```
[core]
   askpass = git-gui--askpass
```

but the error persists, and in the command line it still works like usually.

divmain:
Are you prompted for the password via `gitgui` when committing at the terminal?  Or are you prompted for password in the terminal?

aristidesfl:
@divmain at the moment I'm not prompted at all, I believe because of `gpg-agent`.

aristidesfl:
Isn't gpg is responsible for asking for the private key password? are you sure that `git-gui--askpass` is involved in asking in commit signing? - as opposed to ssh key and http auth passwords

divmain:
Can you check to see if [setting `no-tty` in `~/.gnupg/gpg-agent.conf`](https://github.com/kemayo/sublime-text-git/issues/475) helps?  If it does, we should be able to set the appropriate environment config from Sublime so that you don't have to set this globally.

aristidesfl:
@divmain that solves the issue. that last comment makes a lot of sense. Possible to streamline with Gitsavvy?

divmain:
Sweet!  I'll leave this open until we push a fix into master.  Thanks for confirming!

stoivo:
@divmain, did you see anywhere if it is possible to override that with a environment variable?

aristidesfl:
[...]

PR #1018, opened 2018-09-23
GPG no-tty helper script and options

asfaltboy:
As in #596, this change documents a proper GPG setup on tty-enabled OS (Linux/MacOS), and correct credentials docs. This commit also adds the `global_pre_flags` setting to allow specifying options `BEFORE` the command, such as [the "-c" git per-command config option](https://git-scm.com/docs/git#git--cltnamegtltvaluegt).

The GPG prompt will only work in Sublime if user sets up `git config` with the options:
```
gpg.program=./scripts/stgpg.sh
commit.gpgsign=true
```

or instead of last one, passes the `-S` flag to commit command. So, I added the options in the GitSavvy settings comments.

TODO / Discuss:
- [ ] Figure how to use the script file from the packaged zip. Maybe create the file in the User folder on launch?
- [ ] Should we do this by default? Or only if user specifies an option? Or leave it in settings / docs?
- [ ] gpg-agent could freeze (happened to me while importing a key), breaking the whole commit flow, until it's killed. Should we take responsibility for this and warn users?

### DIFF
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -184,12 +184,27 @@
 
     /*
         For each command specified, always include the command line flags
-        indicated in the global_flags hash.
+        indicated in the global_flags option AFTER the command.
      */
     "global_flags": {
         // --no-columns is not supported in Git versions <1.7.11.  If Git is configured
         //   to use columns globally, --no-columns should be added here.
         // "branch": ["--no-columns"]
+        //
+        // or, configure a GPG key to sign commits with a given key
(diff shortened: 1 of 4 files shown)


randy3k:
I guess we could merge this PR first and evaluate its usefulness.

randy3k:
For the record, I use MacGPG and its `gpg` works in non tty software.

asfaltboy:
I think without at least documenting it, it won't be of much use. Sorry I didn't get a chance to

Issue #653, opened 2017-05-24
Performance issue


max-mykhailenko:
Hi! I'm using gitsavvy almost from first releases and every new release decrease plugin performance. I can't fast add to index files on status dashboard, multiple pressing `s` will cause errors. Switching between status dashboard and branches also works slow.

How I can help with debug this problem?
How I can improve performance?

Used on iMac 2015
3.2 GHz Intel Core i5
16 GB 1867 MHz DDR3
AMD Radeon R9 M390 2048 MB

stoivo:
I have been waiting to see an issue like this.

I don't know what's slow. I think the issue is that we make to many calls to git. We can see all calls to git in GitSavvy logging. Possibly I should improve it to add the duration. 

Can you record some log now with the current release and create a gist with it?

max-mykhailenko:
Yes, no problem

max-mykhailenko:
So I can't repeat errors in normal use right now, but if I try to stage/unstage/stage fast I got error about locked index.
`Unable to create '/.../.git/index.lock`
Same error also happens in normal using, I'll log it next time.

Log with "forced" bug is here.
https://gist.github.com/max-mykhailenko/9a61fe5ac5020959da075a98801f87e0

max-mykhailenko:
Maybe it happens in larger repositories more often, I'll check it.

stoivo:
I just looked at the log, this si all the called made aggregated
```
times comand
 4 git add -A
 3 git reset
 4 git rev-parse --show-toplevel
10 git log -n 1 --pretty=format:%h %s --abbrev-commit
10 git stash list
10 git status --porcelain -z
10 git status -b --porcelain
```
This doesn't mean much. I will add time.

stoivo:
If you are running it GitSavvy from git repo you can check out my branch to include seconds. 

Otherwise, we will merge it to master in a couple of days. but not make a release before next month.

max-mykhailenko:
I'll be waiting release, too many plans on this weekends

stoivo:
We have made a minor release, can you record some log and see that includes seconds and post it here?

max-mykhailenko:
Sometimes happens when closing files during pull.
https://gist.github.com/max-mykhailenko/ac4dad3a93096de0fa5ef9e8fc893e2f

max-mykhailenko:
One more log with errors. Happens don't know when.
[...]

PR #1022, opened 2018-10-01
sanitize ansi sequences

### LINKED ISSUE #1020: Translate/strip color codes from push log
When pushing to a remote that responds with some build information (such as Heroku), color in the output is not displayed, but rather the raw color codes, e.g:
```
remote: <0x1b>[1;33m !! <0x1b>[0m
```
Only `!!` is the actual output here.

If you could either strip such data or use it to display actual colors, that would be great.

ST: 3176
Git: 2.19.0
OS: Arch, kernel 4.18.5

randy3k:
close #1020

as title, sanitize ansi sequences

### DIFF
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -55,6 +55,9 @@ GIT_REQUIRE_MINOR = 9
 GIT_REQUIRE_PATCH = 0
 
 
+ANSI_ESCAPE = re.compile(r'\x1B\[[0-?]*[ -/]*[@-~]')
+
+
 class LoggingProcessWrapper(object):
 
     """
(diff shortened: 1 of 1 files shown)


stoivo:
Do you think it make sense to do this for every command Gitsavvy called to git. I am thinking this may not be the bet this to do performance wise. I know is't a small thing but executed very frequent.

randy3k:
This pr sanitizes ansi sequences for every command called by `self.git(...)`.

stoivo:
Yes and it is wasiting resouces for 99.99% of all commands. Thinks works fine for you and me but we should't add uneccessary load.

randy3k:
I believe that regular expression searches are quite efficient though.

stoivo:
quite efficient but unneccearry, espcelly when we have a open issue about performance. https://github.com/divmain/GitSavvy/issues/653

randy3k:
The slow performance is usually associated to slow disk access. I am not too worried about the performance hit here.

stoivo:
It's probaly fine so we can merge it

Issue #1019, opened 2018-09-28
Attempt to add file change age to status


lttlrck:
I need some help. I want to add the file change age to the status and although I got it working it broke shortcuts such as open, stage single file etc.

This is what it looks like:

![image](https://user-images.githubusercontent.com/779029/46237521-da6e1600-c339-11e8-9eb5-8da3ffe0f558.png)

The change afters several slices e.g. `line[:4]` and `line[4:]` throughout `core/interfaces/status.py` and although I tried correcting them, it didn't fix the problems.

Have I gone about this the wrong way? Am I missing something in additional to these slices? I'm a little confused about regions, and how exactly the file selection works. I rarely touch Python.

This the commit that has the changes I made (minus any slice adjustments as they didn't solve anything):

https://github.com/lttlrck/GitSavvy/commit/1e9b3b6c5b15b40491da08ce4018ffc3baec2052

Any help would be greatly appreciated, this feature it pretty useful for finding associated commits (though branches are better, sometimes things get out of hand :smiley: )

asfaltboy:
Thats a nice feature @lttlrck , thanks for working on it! How about submitting a PR and we can discuss the development in detail?

Now, about the keyboard bindings; these are specified in the `.sublime-keymap` files, like this one:
https://github.com/divmain/GitSavvy/blob/de9caff3222924ce4fd6a71a1f4f2804126f2c84/Default.sublime-keymap#L107-L119

As you can see the context selector uses the `git-savvy.status` to specify which view the bindings are valid on. The `status.sublime-syntax` file specifies how scope for each row is matched:
https://github.com/divmain/GitSavvy/blob/9ec1cc44455a11e1dfef89629fc7c4e8c01f6a57/syntax/status.sublime-syntax#L46-L58

I guess the number of spaces there doesn't match with the date added.

lttlrck:
thanks for help, looks much better. I made a PR.

PR #1023, opened 2018-10-01
add file change age to status view

lttlrck:
This add file change age to the status view.

The line slices seem a little fragile, maybe they should be turned into function calls.

Only cursory testing, it seems ok, but I only really tested with what I use on a daily basis.

### DIFF
--- a/core/interfaces/status.py
+++ b/core/interfaces/status.py
@@ -1,5 +1,6 @@
 import os
-
+import calendar
+import time
 import sublime
 from sublime_plugin import WindowCommand, TextCommand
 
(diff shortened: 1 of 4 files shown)


asfaltboy:
More details in ticket #1019

core/interfaces/status.py
```diff
@@ -235,7 +256,7 @@ class GsStatusOpenFileCommand(TextCommand, GitCommand):
 
     def run(self, edit):
         lines = util.view.get_lines_from_regions(self.view, self.view.sel())
-        file_paths = (line.strip() for line in lines if line[:4] == "    ")
+        file_paths = (line[5:].strip() for line in lines if line[5:8] == "   ")
```
asfaltboy: I feel we should take this opportunity to clarify what these slices mean and, possibly, replace with some sort of regex/syntax-scope matching instead?

lttlrck: yes, especially as I got it wrong... it should be `line[7:].strip()`

randy3k:
@lttlrck I have changed the merge base to `dev`. It is our practice to merge the new features into `dev` for evaluation first.

lttlrck:
autostash breaks it because it resets the file timestamps on pull. Branching is another problem, I had that in the back of my mind but wasn't too concerned about that case, but autostash is a bit of a deal breaker for me. It'll need some research.

core/interfaces/status.py
randy3k: it seems to break too when a file was modified 100 weeks ago.

randy3k:
@lttlrck could you also fix the [tests](https://github.com/divmain/GitSavvy/blob/master/syntax/test/syntax_test_status.txt) for the status dashboard?

stoivo:
What's the status on this?

lttlrck:
> What's the status on this?

well I'm using it and it's useful despite the timestamps being reset on branch changes and pull (with auto-stash), stashing etc.

However I would like to colorize the the date, so green (recent) -> red (old) to help pick out the recent/old changes. But that means diving into themes which I don't have time for at the moment.

lttlrck:
> @lttlrck could you also fix the [tests](https://github.com/divmain/GitSavvy/blob/master/syntax/test/syntax_test_status.txt) for the status dashboard?

Not sure what I'm doing wrong, I thought that last commit would be ok, but it looks like it's tripping up on the deleted file test. I'll have to take a look later. Also, apologies I did I push force - making numerous commits to figure out the test failure problem is nasty.  I assume that's ok though? I can't run the tests locally AFAIK,

PR #1031, opened 2018-11-03
Make live panel output optionally sync

### LINKED ISSUE #1021: Live push logs
When pushing to a remote that responds with some build information (such as Heroku), it will stream the response to the client.
However, when pushing with GitSavvy, I only get the complete log _once the push has finished_, and not streamed like in a terminal.

Not sure if there are limits with the ST API, but support for this would be nice.
(The `[Done in XXs]` at the end is more than sufficient to indicate the process completion).

ST: 3176
Git: 2.19.0
OS: Arch, kernel 4.18.5

asfaltboy:
to fix live_panel_output not showing panel until command is complete

fix #1021

### DIFF
--- a/common/util/log.py
+++ b/common/util/log.py
@@ -1,13 +1,23 @@
 import sublime
 
 
-def panel(*msgs):
+def panel(*msgs, run_async=True):
     msg = "\n".join(str(msg) for msg in msgs)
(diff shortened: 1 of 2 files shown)


randy3k:
the macOS failures are not relevant. I will fix it at some point.

PR #1033, opened 2018-11-07
Optimize status interface

### LINKED ISSUE #1032: Optimize Status Interface
In general I want to see UI performance improvements. Since everything around the current repo status is probably used most often, I would start here.

- Just opening the Status interface should be fast
- Staging/Unstaging files should ideally be optimistic, but in general hitting `s s s` (in a meaningful typing speed) should stage three files

Some observations:

- `git status` is called very often already for the statusbar, we could use that information in the Status interface.
- `git stash list` is very slow (>1 second). Waiting for this call seems unnecessary bc most often this call will probably return nothing at all, or the same output as before. (E.g. staging/unstaging/discarding does not change the stash list)
- The file statuses are generated via a call to `git status -z`, the branch gets computed via `git status -b`. These two calls can be merged into one call.

kaste:
Fixes #1032 

DONE: 

- short-circuit staging/unstaging files (still not optimistic)
- render view as data arrives, whereby `git stash list` has the lowest priority
- optimize `git status` calls
- only draw when view changed

TODO:

- ~~Optimize `git status` calls~~
- ~~User looses selections / not multiple cursors but selected ranges collapse after each render~~ (This is actually implemented in `GsNewContentAndRegionsCommand` so I don't touch that) 
- ~~Only re-render if state actually changes / this should be trivial now since the actual render is a pure function on `state`~~

### DIFF
--- a/common/commands/view_manipulation.py
+++ b/common/commands/view_manipulation.py
@@ -69,6 +69,7 @@ class GsHandleVintageousCommand(TextCommand):
         if savvy_settings.get("vintageous_friendly", False) is True:
             self.view.settings().set("git_savvy.vintageous_friendly", True)
             if savvy_settings.get("vintageous_enter_insert_mode", False) is True:
+                self.view.settings().set("vintageous_reset_mode_when_switching_tabs", False)
                 self.view.run_command("_enter_insert_mode")
 
 
--- a/core/git_mixins/active_branch.py
+++ b/core/git_mixins/active_branch.py
@@ -15,7 +15,7 @@ class ActiveBranchMixin():
         except StopIteration:
             return None
 
-    def _get_branch_status_components(self):
+    def _get_branch_status_components(self, lines):
         """
         Return a tuple of:
 
@@ -28,9 +28,8 @@ class ActiveBranchMixin():
           6) # commits behind of remote
           7) boolean indicating whether the remote branch is gone
         """
-        stdout = self.git("status", "-b", "--porcelain").strip()
 
-        first_line, *addl_lines = stdout.split("\n", 2)
+        first_line, *addl_lines = lines
         # Any additional lines will mean files have changed or are untracked.
         clean = len(addl_lines) == 0
 
@@ -67,8 +66,12 @@ class ActiveBranchMixin():
         If a delimeter is provided, join tuple components with it, and return
         that value.
         """
-        detached, initial, branch, remote, clean, ahead, behind, gone = \
-            self._get_branch_status_components()
+        lines = self._get_status()
+        branch_status = self._get_branch_status_components(lines)
+        return self._format_branch_status(branch_status, delim)
+
+    def _format_branch_status(self, branch_status, delim=None):
+        detached, initial, branch, remote, clean, ahead, behind, gone = branch_status
 
         secondary = []
 
@@ -102,14 +105,15 @@ class ActiveBranchMixin():
         return status, secondary
 
     def get_branch_status_short(self):
-
         if self.in_rebase():
             return "(no branch, rebasing {})".format(self.rebase_branch_name())
 
-        merge_head = self.merge_head() if self.in_merge() else ""
+        lines = self._get_status()
+        branch_status = self._get_branch_status_components(lines)
+        return self._format_branch_status_short(branch_status)
 
-        detached, initial, branch, remote, clean, ahead, behind, gone = \
-            self._get_branch_status_components()
+    def _format_branch_status_short(self, branch_status):
+        detached, initial, branch, remote, clean, ahead, behind, gone = branch_status
 
         dirty = "" if clean else "*"
 
@@ -123,6 +127,7 @@ class ActiveBranchMixin():
         if behind:
             output += "-" + behind
 
+        merge_head = self.merge_head() if self.in_merge() else ""
         return output if not merge_head else output + " (merging {})".format(merge_head)
 
     def get_commit_hash_for_head(self):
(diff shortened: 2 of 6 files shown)


randy3k:
Thanks for your contribution. I guess this PR may take some time to review.

kaste:
The first 9 commits are actually good to read, I could also add comments. The 10th commit is too complex, but splitting a commit after the fact is also super hairy.

asfaltboy:
@kaste that's an inspiring effort, thanks for contributing! I think the PR makes a lot of sense, and the state caching is something I had long been wishing for. To make the PR easier to read, could you kindly write a few words (e.g in docstring) to help walk us through the logical flow of status interface state / render changes?

 I have not tested it yet, I'll check it out now and report back if any issues pop up.

kaste:
I actually saw that render has this side-effect but assumed it would be a like a post process only necessary once per render. 

Now this essentially comes down to invoking just another command 'enter_insert_mode' iirc. Do we have to call this actually on (every) focus? Is there somewhere the reverse action leave_insert_mode?

asfaltboy:
I think this is the default of Vintagous/NeoVintageous to switch to normal mode on tab focus:
https://github.com/NeoVintageous/NeoVintageous/blob/master/res/doc/neovintageous.txt#L528

Adding this setting `"vintageous_reset_mode_when_switching_tabs": false` to my settings seems to work well. Maybe we should point to this in our release notes / README / documentation?

kaste:
Ah okay, then the solution would be to just ... well 

If you look here 

https://github.com/divmain/GitSavvy/blob/770f91523655c19a0486555989790e1b398d6357/common/commands/view_manipulation.py#L60-L85

it just looks like we could do this as a one time side-effect:

```
if savvy_settings.get("vintageous_friendly", False) is True:
    self.view.settings().set("git_savvy.vintageous_friendly", True)
    if savvy_settings.get("vintageous_enter_insert_mode", False) is True:
        self.view.settings().set("vintageous_reset_mode_when_switching_tabs", False)  # <--
        self.view.run_command("_enter_insert_mode")

if savvy_settings.get("arrow_keys_navigation", False) is True:
    self.view.settings().set("git_savvy.arrow_keys_navigation", True)
```

kaste:
Hm, from reading the code I don't think the inline diff view calls 'enter_insert_mode' on refresh/activate only after the first run. Can you check this? Does it switch into insert mode if you switch tabs?

asfaltboy:
You're right about the _inline diff_ view, and your suggestion to use `view.settings.set` fixes it for this view as well. I'll this single-line commit to your PR if you don't object, because otherwise `vintageous_enter_insert_mode` setting would appear broken to users.

kaste:
Ok, that was an easy fix. 

Another bug: You can't toggle the help bc this state is set in the view settings (and then read like so `self.view.settings().get("git_savvy.help_hidden")`). Toggling code is:

https://github.com/divmain/GitSavvy/blob/50c68a323006ed00f76afda62c662913fc29e403/common/ui.py#L299-L312

You see that it requests a complete refresh. A typical solution here could be (in the toggle command just shown)

```
interface = get_interface(self.view.id())
if interface:
	interface.toggle_help()
```

We then define on the base class 

```
def toggle_help(self):
	current_help = bool(self.view.settings().get("git_savvy.help_hidden"))
    self.view.settings().set("git_savvy.help_hidden", not current_help)
	self.render()
```

which we override for the status dashboard

```
def toggle_help(self):
	self.state.update({'show_help': not self.state['show_help']})
	self.just_render()
```

Since both bugs are due to `@distinct_until_state_changed` we could also just drop this as premature optimization. But this is also lame, a complete refresh bc we toggle the help text is a waste we now see this and can tackle it.

asfaltboy:
Shall we tackle "toggle help requires full refresh" in another ticket / PR? This one is already reviewed and as far as I'm concerned it's ready to be merged.

stoivo:
Look good tested it locally and I can feel is't snappier. I added a syntax test.

kaste:
* Added docstrings, 
* `call_then_render` was an ad-hoc method to make it work. Wrote it now using a `update_state` fn
* Basically fix 'toggle help' without making the solution pointed out above. I mean: Just fix it.

kaste:
Note:

* `self.state.update(data)` this one line ... we probably need a lock here bc we call it from two threads. 
* Same:

```
        self.clear_regions()
        rendered = self._render_template()
```

in `just_render`. It would be better if `_render_template(state) -> (content, regions)`. We should ensure that regions contains all possible keys (even empty ones) bc then we don't need to erase unused regions, we just draw them empty. (I would prefer to not use regions here at all, and just search for scopes in the actions. Less state to manage, we just draw the view, and extract information out of it using syntax selectors or scopes.)

kaste:
LGTM

asfaltboy:
The lock and state related functionality looks legit; next step would be to extract these into a reusable Mixin and start converting more interface views! I'm gonna merge this into dev as I've been using this for a while without noticeable issues. Thanks again @kaste !

@randy3k @stoivo shall we ready the next release (last one was in August) ?

kaste:
Hooray!

@asfaltboy a tip here: Do not build abstractions or mixins over one case. If you did something three or four times it's much easier to see how it all fits together. We only have three more dashboard views but on the other side we have other fancy views like `graph`, `diff`, `commit_prepare` for which we currently have a completely different abstraction, although there are also just views with state.  It's like in TDD, *first* make it work in as few, but sane steps as possible *then* (maybe) do the refactoring.

PR #1034, opened 2018-11-07
Fix travis osx builds

kaste:
The (currently failing) `brew update` dance is not necessary anymore. We just skip it, they ship py36 which is enough.


randy3k:
Thanks

PR #1037, opened 2018-11-08
Fix and improve diff view

kaste:
Fixes: Basically undefined behavior if you (hH)unk being on the top of the file. See the commit message https://github.com/divmain/GitSavvy/commit/61e442a24515a612a3e70c56a98bfd66ef7ae694

Fixes: After (hH)unking the cursor was misplaced. I just invoke the navigate command and I think it's 👌

### DIFF
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -70,8 +70,7 @@ class GsDiffCommand(WindowCommand, GitCommand):
 
         self.window.focus_view(diff_view)
         diff_view.sel().clear()
-        diff_view.run_command("gs_diff_refresh")
-        diff_view.run_command("gs_diff_navigate")
+        diff_view.run_command("gs_diff_refresh", {'navigate_to_next_hunk': True})
         diff_view.run_command("gs_handle_vintageous")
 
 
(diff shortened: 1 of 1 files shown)


randy3k:
Does it fix #1040 ?

kaste:
@randy3k No, these are just two random fixes for bugs I never reported. Esp. the cursor jumping was annoying, and today was the day to just do it. 

I can add the mouse-clicking thing bc it is just a three-liner. 

The jumping from file to hunk in diff view is a bit more code, so at some point this gets trickier to review.

randy3k:
I don't think I have to staimnia to review it (I am not very familar with this part of the codebase). Though, I think that the changes are legitimate.

core/commands/diff.py
```diff
@@ -221,18 +221,24 @@ def get_hunk_diff(self, pt):
...
+        else:
+            window = self.view.window()
+            if window:
+                window.status_message('Not within a hunk')
+            return  # Error!
+
+        header_start, header_end = max(
+            (header_start, header_end)
+            for header_start, header_end in zip(self.diff_starts, self.diff_header_ends)
+            if (header_start, header_end) < (hunk_start, hunk_end)
```
stoivo: How does it work when comparing a tuple to an other tuple?
Not sure how the bisect works since it's is a long time since I looked at it

stoivo:
Other than that it looks good.

core/commands/diff.py
stoivo: Thanks @kaste

PR #1047, opened 2018-11-14
Show toggle remotes help message only if there are any remotes

### LINKED ISSUE #1046: `NO_REMOTES_MESSAGE` never shows in tags dashboard
In the tags dashboard, if there are no remotes configured, you promise to show this const

```
NO_REMOTES_MESSAGE = "You have not configured any remotes."
```

but this does not happen.

kaste:
Fixes #1046 

Because showing `** Press [e] to toggle display of remote branches. **` is lame when no remotes have been configured, we just suppress this help message in such cases.


randy3k:
LGTM

PR #1051, opened 2018-11-15
Tag the line containing HEAD for styling purpopses

kaste:
I'm not super smart in this area. 

image

I just wanted to highlight the branch/tag thing in the graph view. After the fact, I thought maybe it's 'better' to mark the whole line. But I couldn't get that to work ~without spending too much time~ at all, I basically have no idea.

### DIFF
--- a/syntax/graph.sublime-syntax
+++ b/syntax/graph.sublime-syntax
@@ -18,6 +18,21 @@ contexts:
       comment: SHA
       scope: constant.numeric.graph.commit-hash.git-savvy
 
+    - match: (?=\s*\(HEAD)
+      push:
+        - meta_scope: meta.graph.graph-line.head.git-savvy
+        - include: branch-name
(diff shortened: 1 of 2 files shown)


kaste:
Would be nice to have it for the whole line, so I could mark the commit dot at the beginning of the line. 🤔

randy3k:
Highlighting the whole line is possible, but it will be some work. The basic idea is to use `view.add_regions` to add a specific scope to the region (and erase it when the view is updated.)

randy3k:
It would say it is good for this purpose now. Let's merge it.

kaste:
Well, I thought I can just do a lookahead for '(HEAD ' and then mark that line using a meta-scope, and pop on '$'. Now, I already do such a look-ahead here but in the wrong place.

kaste:
Sometimes you just have to write down the problem and then you get #1067

PR #1056, opened 2018-11-16
Improve graph view

### LINKED ISSUE #1052: Activating a graph view blocks the main UI
This is seriously notable. If you activate using the mouse, click while moving the mouse, you often select text. You can also see that cursor appears only after a delay, or `ctrl+tab` is laggy and not instant at all.

### LINKED ISSUE #1053: In the `graph view` clicking on a row should update the commit panel
Currently, you must use `,.` special keys to navigate through the graph. Technically, the view should respond to `on_selection_modified_async`-events.

### LINKED ISSUE #1054: Changing `git_graph_args` should have immediate effect
Currently, if you change `git_graph_args` in the settings, these new settings are not applied to already open graph views. Even hitting `r` (refresh) has no effect. You actually have to open a new `git: graph` to see the changes. 

Technically, we cache `git_graph_args` when creating the view. This is hopefully not necessary.

kaste:
Fixes #1052 
Fixes #1053 
Fixes #1054 

PLUS:

- Commit info panel can be hidden via `<ESC>` or mouse
- Commit info panel can be shown via mouse
- Don't fight with other panels the user might open. E.g. find or console
- Cache `git show <commit>` calls

Let's see how this one goes. `commit_compare` is actually the same as graph, that was super confusing first. 🍭 

🚶 🍷  🛌 

~Needs https://github.com/divmain/GitSavvy/pull/1033/commits/50c68a323006ed00f76afda62c662913fc29e403 which is included in #1033 for vintagous~ #1033 has been merged

### DIFF
--- a/common/commands/view_manipulation.py
+++ b/common/commands/view_manipulation.py
@@ -30,8 +30,13 @@ class GsReplaceViewTextCommand(TextCommand):
     a single cursor at the start of the file.
     """
 
-    def run(self, edit, text, nuke_cursors=False):
+    def run(self, edit, text, nuke_cursors=False, restore_cursors=False):
         cursors_num = len(self.view.sel())
+
+        if restore_cursors:
+            save_cursors = [self.view.rowcol(s.a) for s in self.view.sel()]
+            self.view.sel().clear()
+
         is_read_only = self.view.is_read_only()
         self.view.set_read_only(False)
         self.view.replace(edit, sublime.Region(0, self.view.size()), text)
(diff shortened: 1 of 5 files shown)


kaste:
@randy3k It seems that *all* code we're running on the worker has no coverage which makes me 😭 See: https://coveralls.io/builds/20322818/source?filename=core%2Fcommands%2Flog_graph.py#L34   I do functional tests here, and sure I painted the view. The only thing I did not cover here is the actual execution of the git command. (I mock `self.git`.)   😞 Do you know anything about this?

I actually don't want to do unit tests here. a) how we realize responsiveness (here: Sublime worker thread) is an implementation detail, b) it would break if we refactor to a different architecture e.g. if the drawing/refreshing is not implemented as a `TextCommand` but a normal function.

kaste:
Can you please give me some feedback here about the functionality/UX. I do the refactoring after the functionality so a code review is not necessary.

I would like it if the commit detail panel would close/open automatically if I switch back and forth to other views. I didn't open an issue for that. But that sounds 👍, no?

stoivo:
Can you rebase, merged some of the other PRs

randy3k:
I love it. Thanks @kaste

PR #1058, opened 2018-11-17
Support clicking on filenames in the status view

kaste:
The typical double-click we know from "Find in files" can be used to navigate
to files from the status dashboard via mouse.

🎆 🌃 🕺

### DIFF
--- a/core/interfaces/status.py
+++ b/core/interfaces/status.py
@@ -255,6 +255,27 @@ class StatusInterface(ui.Interface, GitCommand):
         """
         self.update_state(self.fetch_repo_status, self.just_render)
 
+    def after_view_creation(self, view):
+        # Expected
+        #  - common/commands/view_manipulation.py
+        #    common/ui.py
+        #    core/commands/commit_compare.py -> core/commands/commit_compare_foo.py
+        # But do not match our stashes or anything from our help
+        #    (1) log git start/stop
+        #           [t] create stash
+        file_regex = r"^(?:    .+ -> |  [ -] (?!\(\d+\) ))(\S.*)$"
+        #                  ^ leading 4 spaces
+        #                      ^ a filename
+        #                         ^ marker indicating a rename/move
+        #                            ^ OR
+        #                             ^ leading 4 spaces or two spaces and our deleted marker
+        #                                    ^^^^^^^^^^^ but be aware to not match stashes
+        #                                                ^^^^^^ the actual filename matcher
+        # Note: A filename cannot start with a space bc otherwise our naive `.*` could consume
+        # only whitespace.
+        view.settings().set("result_file_regex", file_regex)
+        view.settings().set("result_base_dir", self.repo_path)
+
     def on_new_dashboard(self):
         self.view.run_command("gs_status_navigate_file")
 
--- a/core/interfaces/status.py
+++ b/core/interfaces/status.py
@@ -11,6 +11,31 @@ from ..git_command import GitCommand
 from ...common import util
 
 
+# Expected
+#  - common/commands/view_manipulation.py
+#    common/ui.py
+#    core/commands/commit_compare.py -> core/commands/commit_compare_foo.py
+# But do not match our stashes or anything from our help
+#    (1) log git start/stop
+#           [t] create stash
+EXTRACT_FILENAME_RE = (
+    r"^(?:    .+ -> |  [ -] (?!\(\d+\) ))"
+    #     ^ leading 4 spaces
+    #         ^ a filename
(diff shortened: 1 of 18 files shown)


kaste:
Tests https://regex101.com/r/PrPl7q/2/

core/interfaces/status.py
```diff
@@ -137,6 +137,27 @@ def pre_render(self):
          self.untracked_entries,
          self.conflict_entries) = self.sort_status_entries(self.get_status())
 
+    def after_view_creation(self, view):
```
stoivo: Did we have a fansy way to add these doc test to be executable?

kaste:
Test coverage increases by 2%. So we have 34%.

asfaltboy:
Awesome idea, and kudos for adding more tests! I have only one concern: could we include mockito as a [package-control dependency][1] instead of copy-pasted ?

A cool enhancement of this would be to implement a stash peek command (basically output `git stash show -p` into a panel), and be able to trigger it on click!

[1]: https://packagecontrol.io/docs/dependencies

kaste:
@asfaltboy mockito and parameterized are only needed for tests. So instead of installing it via PC as a dependency we should actually mark them 

```
# .gitattributes
/tests/             export-ignore
```

so that normal users do not have to download these libs.

For the click here we just enable what Sublime can do. We cannot run arbitrary code on double-click, Sublime only supports opening files (optionally at a `<line:col>`).

asfaltboy:
Good idea about excluding tests from users. Feel free to add the gitignore as suggested!

And, for our sake, can we maybe include the mockito package (and any other dev dependencies) in a `test-reqs.txt` or `dev-reqs.txt or something in the style? We could then explain how to install dependencies in our README/wiki, under the contributing section

kaste:
- Added `.gitattributes`
- Updated test to match new async dashboard and optimized `git status` calls
- Fix: Handle edge case when working dir is clean

What you want with `test-reqs.txt` is unfortunately not clear to me. We need installed dependencies for travis et.al. (I also use a dev version of mockito btw. 🙄) A simple `pip` will not work here, this will probably install a global package but we need it inside the package, inside Sublime. I actually used `pip -t ...` and then just removed the metadata, there is also nothing inherently bad with checking in your dependencies. You get nice diffs if you update a package, you can just fix any edge case.

kaste:
Now, you always have to maintain your dependencies in some way. 

We just check them in to have full control over everything. We actually *don't* want to auto update them esp. since python33 is basically dead. If you want to update them, you just `pip -t GitSavvy/tests`. 

Just keep it simple, man. For example, if we `pip` in travis, we already get a python36 version on osx iirc bc travis support for py33 is getting low already.

kaste:
@asfaltboy I just had `test_json` throwing (because I copied an unrelated json config file into the project). Seems like we copy-pasted a json linter here, slightly patched it. This linter is out-of-date so it blew away instead of producing a nice message. Beside that you probably can use `eslint` from within _travis_, how could we update the linter? No chance, we have to maintain it ourself. Honestly IMO, this is a magnitude dirtier than what this PR does.

Now, what we could try is to install the dev dependencies into 'GitSavvy/tests/lib' (lib or _vendor_) or even top-level `GitSavvy/vendor` (or _lib_) and keep all the metadata pip normally produces. Still checking in bc contributors don't have python33 with pip installed, and getting 'fresh' dependencies for py33 is bascially not a thing anymore. You already can't get a recent flake8 or pytest for it. (Since I know the maintainer of mockito 🙄 we probably _never_ get a fresher version of it supporting 33 anyway.)

kaste:
How can we move forward here? All my PR's have these test dependencies, so I'm feeling a bit unsecure if this is a general blocker for you. I personally hate manual testing after each code change bc I loose my focus. 

I actually share the concerns @asfaltboy has, but at the same time think it is a bit out scope of feature driven PR's. We're in Sublime land here, and python33 is basically history. If someone has a nifty idea that works on CI and is easy for contributors then 👍 just do it.

asfaltboy:
Sorry for my lack of feedback. I think for this feature sake, and having more tests is a _divine_ feature, we should merge as is and discuss externalising dependencies in a different time. Perhaps when sublime have moved to a newer, better supported, version of python or maybe when we've implemented a [better dependency control][1] mechanism in package-control.

@randy3k @stoivo if either of you agrees then LGTM.

[1]: https://github.com/wbond/package_control/issues/1183#issuecomment-440813158

kaste:
(Cough) @randy3k approved this PR already

kaste:
That's a great feature. Love clicking around.

PR #1059, opened 2018-11-18
Fix: Allow closing the commit view if `prompt_on_abort_commit=false`

### LINKED ISSUE #1055: `"prompt_on_abort_commit": false` broken
With the following settings

```
  "prompt_on_abort_commit": false,
  "commit_on_close": false,  # this is also the default
```

* Make a change
* Open status dashboard, `C` to commit changes

- Commit view opens with empty message
- Press `ctrl+w`, nothing happens

* Enter a message for the commit
* Press `ctrl+w`, nothing happens

Pressing the `x` button on the tab or in the sidebar works.

Regression from #970

kaste:
Fixes #1055

Regardless of what #1016 does or eventually will do, this needs a hotfix bc right now, the user can't close the commit view using `ctrl+w` if `prompt_on_abort_commit` is false.

### DIFF
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -402,3 +402,5 @@ class GsCommitViewCloseCommand(TextCommand, GitCommand):
 
             if ok:
                 self.view.close()
+        else:
+            self.view.close()


randy3k:
Thanks

Issue #856, opened 2018-02-13
Add keybindings to switch to cached diff and back


3v1n0:
When un diff mode to stage lines it could happen that you wrongly (un)stage something you didn't want, and then you need to go back to status and open the (un)cached version of the same diff.

It would be convenient to have keybindings to quickly switch views.

stoivo:
Agree, do you use the diff of the inline diff?
I think this should exist for both but It is interesting to know what you are using the most.

3v1n0:
I use them both... And, yes. It should be added to the two :)

PR #1061, opened 2018-11-20
Iterate on diff view

kaste:
This is pretty much to get early feedback. I saw #856 and implemented it. The keybinding I used is `<tab>` to quickly switch between *cached* and *non-cached* (in git parlance) mode. 

`<tab>` is convenient, however, if we ever want to use 'diff all files' as a dashboard view it would clash with our 'cycle views' command.

I also added a basic header/prelude to the diff view. This is similar to the prelude we have e.g. in the status dashboard. And I think it's really easier to grasp what the view is currently showing. (Esp. with the implemented fast toggle.) It currently shows the file you're diffing if any, the two commits you're comparing if any. For the typical `git diff` it tells you either 'UNSTAGED CHANGES' or 'STAGED CHANGES (Will commit)'. This is the exact wording git GUI uses, and avoids the more technically, usually confusing 'cached' terminology.  

I added a LOL-naive syntax. (Honestly, it did not click yet how these syntax files work, and so it is worse than writing CSS for me.)

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -588,6 +588,15 @@
         ]
     },
     {
+        "keys": ["tab"],
+        "command": "gs_diff_toggle_setting",
+        "args": { "setting": "in_cached_mode" },
+        "context": [
+            { "key": "setting.command_mode", "operator": "equal", "operand": false },
+            { "key": "setting.git_savvy.diff_view", "operator": "equal", "operand": true }
+        ]
+    },
+    {
         "keys": ["s"],
         "command": "gs_diff_toggle_setting",
         "args": { "setting": "ignore_whitespace" },
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -542,6 +542,15 @@
         ]
     },
     {
+        "keys": ["c"],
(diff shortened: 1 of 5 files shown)


kaste:
Hm 🤔 quickly switching between staged and unstaged view makes you want to invoke 'commit' (`[c]`?) directly, just saying. It feels right.

We should improve on cursor jumping though.

randy3k:
Thank you for the contribution and sorry for the slow review. 

It looks awesome. Just a few comments

1. It would be nice to have the cursor location saved while switching between the cached and uncached diff.

2. We also need to put the `[tab]` shortcut in the popup doc: https://github.com/divmain/GitSavvy/blob/master/popups/diff_view.html

3. I also vote for the `[c]` for commit idea.

kaste:
The h/H was merely about the static help. The actual commands are safe.

kaste:
This feels great. The [c] thing is 💋

kaste:
Let's scroll.

kaste:
Also added '[C]'

randy3k:
Thanks again.

Issue #1040, opened 2018-11-08
Add jump from file to hunk in diff view 


kaste:
In a view I see the gutter marks from Sublime or GitGutter and want to jump to the mostly accurate hunk of the diff. 

(Edge case: Where do we jump to when the user is not even nearly on a changed line.)

PR #1064, opened 2018-11-22
 Make filenames in the diff view clickable

kaste:
This is similar to #1058 but for the diff view.  

I mentioned this in #1040.


randy3k:
Thank you.

kaste:
For other testers, maybe we shouldn't enable clickable filenames for older, 'historical' diffs bc it is likely that the line numbers are wrong etc. We will see.

PR #1067, opened 2018-12-20
Mark the whole line containing 'HEAD' as headline

kaste:
Improves #1051

- Mark the whole line so that we can style the dot

image

- Fixup regex for detached heads
- Really mark the literal 'HEAD' 

image


kaste:
I didn't match the literal 'HEAD' but '(HEAD '. But this didn't work work for detached heads. Two commits later.

kaste:
For styling stuff, users always have to edit/change their theme/scheme.

PR #1069, opened 2018-12-24
Refresh graph view after checkout

kaste:
Now, `log_graph_view` is basically a irregular 'interface' so it should be here. Note that `refresh_gitsavvy_interfaces` is only called after checkouts, otherwise the similar `refresh_gitsavvy` is used. 

That doesn't seem correct, and a picky differentiation. We *should* always call `refresh_gitsavvy_interfaces`, but rename it to `refresh_gitsavvy`.

### DIFF
--- a/common/util/view.py
+++ b/common/util/view.py
@@ -82,6 +82,9 @@ def refresh_gitsavvy_interfaces(window,
         if view.settings().get("git_savvy.interface") is not None:
             view.run_command("gs_interface_refresh", {"nuke_cursors": interface_reset_cursor})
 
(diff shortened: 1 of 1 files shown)


randy3k:
`refresh_gitsavvy` was there for a long time and `refresh_gitsavvy_interfaces` was introduced much later to fix a checkout issue.

I agree that there should exist only one of such functions.

randy3k:
I guess we should resolve the `refresh_gitsavvy_interfaces` vs `refresh_gitsavvy` issue later. Let's merge this for the next release.

Issue #1060, opened 2018-11-19
Performance regression for commit view


kaste:
This is a typical workflow, just on a big repo. Edit file, open status dashboard, (s)tage file, enter (c)ommit view, enter message and commit. (All git status calls are super expensive, and we have probably to much of them, but #1033 reduces some of them.)

A huge problem though is the commit dry-run. I don't think this is strictly necessary, especially if we run (the much faster) git diff after that. If the diff returns empty, there is nothing staged. And we can just tell the user. 

```
 [4168ms] $ git status -b --porcelain
 [3477ms] $ git status --porcelain -z
 [ 78ms] $ git log -n 1 --pretty=format:%h %s --abbrev-commit
 [142ms] $ git stash list
 [3473ms] $ git status -b --porcelain
 [3556ms] $ git status -b --porcelain
 [585ms] $ git add --all -- README.md                    # <= adding the file
 [3519ms] $ git status --porcelain -z
 [ 96ms] $ git log -n 1 --pretty=format:%h %s --abbrev-commit
 [109ms] $ git stash list
 [3514ms] $ git status -b --porcelain
 [3701ms] $ git status -b --porcelain
 [10831ms] $ git commit -q --dry-run                     # <= new dry-run, why?
 [214ms] $ git diff --no-color --patch --stat --cached   # <= pretty fast diff 
 [3498ms] $ git status -b --porcelain
 [3346ms] $ git status -b --porcelain
 [ 94ms] $ git config commit.cleanup                     # ?
 [3209ms] $ git commit -q --cleanup=strip -F -           # <= actual commit
 [3477ms] $ git status --porcelain -z
 [ 93ms] $ git log -n 1 --pretty=format:%h %s --abbrev-commit
 [125ms] $ git stash list
 [3471ms] $ git status -b --porcelain
 [3500ms] $ git status -b --porcelain
 [3555ms] $ git status --porcelain -z
 [102ms] $ git log -n 1 --pretty=format:%h %s --abbrev-commit
 [125ms] $ git stash list
 [3372ms] $ git status -b --porcelain
 [3641ms] $ git status -b --porcelain
```

The same on the tiny GitSavvy repo.

```
 [ 97ms] $ git status -b --porcelain
 [113ms] $ git status --porcelain -z
 [ 78ms] $ git log -n 1 --pretty=format:%h %s --abbrev-commit
 [126ms] $ git stash list
 [ 94ms] $ git status -b --porcelain
[...]

PR #1070, opened 2018-12-24
Release/2.18.0

randy3k:
Let's freeze the dev branch and tag a release. (I don't have time yet to improve the release message.)

A Christmas present to every GitSavvy users.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# Files and directories with the attribute export-ignore won’t be added to
+# archive files. See http://git-scm.com/docs/gitattributes for details.
+
+/tests/             export-ignore
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -8,4 +8,4 @@ Please include the following information when filing a bug:
 - Console error output
 - A description of the problem.
 - Steps to reproduce, if possible
-- If the bug is caused by a failing command, [include a debug log](docs/debug.md#providing-a-debug-log).
+- If the bug is caused by a failing command, [include a debug log](https://github.com/divmain/GitSavvy/blob/master/docs/debug.md#providing-a-debug-log).
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ matrix:
           language: generic
 
 before_install:
-    - curl -OL https://raw.githubusercontent.com/randy3k/UnitTesting/master/sbin/travis.sh
+    - curl -OL https://raw.githubusercontent.com/SublimeText/UnitTesting/master/sbin/travis.sh
     - if [ "$TRAVIS_OS_NAME" == "linux"  ]; then
             export DISPLAY=:99.0;
             sh -e /etc/init.d/xvfb start;
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,12 @@ env:
 
 matrix:
     include:
-        - os: linux
+        - name: Linux
+          os: linux
           language: python
           python: 3.3
-        - os: osx
+        - name: OSX
+          os: osx
           language: generic
 
 before_install:
@@ -26,10 +28,6 @@ before_install:
 install:
     - sh travis.sh bootstrap
     - sh travis.sh install_package_control
-    - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-            brew update;
-            brew install python || brew upgrade python;
-      fi
     - pip3 install python-coveralls pycodestyle;
 
 script:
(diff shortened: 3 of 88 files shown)


kaste:
I'm objecting 'cleanup commit message' and 'check "nothing to commit"' as implemented. I know why they're there and roughly the use-case but they are poor/cheap implemented. 

'cleanup commit message' has the wrong default. It is destructive surprising for new users. T.i. a default setting can now lead to data loss for the end-user. It is a breaking change. It is costly because it adds a git call to every commit esp. as preparation before the commit view actually appears.

'check "nothing to commit"' is just cheap implemented. It is very costly in that it adds a (usually unnecessary) git call to the preparation phase before the commit view actually appears. I showed in #1060 that a '--dry-run' is always more expensive than just a 'diff'. Unfortunately, the 'dry-run' doesn't scale (at least on Windows). So the bigger the repo the slower the call (also in #1060). The 'dry-run' is not necessary if we do a 'diff' call anyway. And we do the 'diff' call by default. If the user doesn't want the diff call we can still do the 'diff' because it is at least faster than the 'dry-run'. Lastly, the UX here is bad. If you do now attempt a commit with "nothing to commit" it basically throws. T.i. it shows the typical error panel from a failing git command. At that point the user is not aware that she issued a commit run, or any other git command. Instead we should just open the usual commit view, and show "Nothing to commit" where we otherwise show the diff. The user then 'ctrl+w' to just close the view, or just stages something. Please note that such a mistake - t.i. attempting to commit when nothing is staged - is a daily routine for any user so we really shouldn't throw here. We need a frictionless experience.   Ideally we allow writing the commit message before any staging and update the diff below our comment as we stage/unstage hunks. (Please note at least, that I personally often write the long commit messages before staging anything. T.i. when writing the commit message takes half an hour itself, staging is not the main problem I'm facing.)

randy3k:
Okay, perhaps we should revert the corresponding PRs and discuss them in the next release.

stoivo:
I am experiencing double output in the live console
```
> /usr/local/bin/git push --set-upstream origin cash_register_ui:cash_register_ui
remote: 
remote: To create a merge request for cash_register_ui, visit:        
remote:   https://gitlab.xx.no/xx/xx/merge_requests/new?merge_request%5Bsource_branch%5D=cash_register_ui        
remote: 
To gitlab.xx.no:xx/xx.git
 * [new branch]          cash_register_ui -> cash_register_ui
Branch 'cash_register_ui' set up to track remote branch 'cash_register_ui' from 'origin'.
remote: 
remote: To create a merge request for cash_register_ui, visit:        
remote:   https://gitlab.xx.no/xx/xx/merge_requests/new?merge_request%5Bsource_branch%5D=cash_register_ui        
remote: 
To gitlab.xx.no:xx/xx.git
 * [new branch]          cash_register_ui -> cash_register_ui
Branch 'cash_register_ui' set up to track remote branch 'cash_register_ui' from 'origin'.

[Done in 2.39s]
```


These are my config
```
{
  "dev_mode": true,
	"vintageous_friendly": false,
	"git_graph_args":
	[
		"log",
		"--pretty=format:%h%d (%ar) by `%an` %s",
		"--graph",
    "--no-color"
	],
	"show_commit_diff": "full",
	"show_panel_for":
	[
		"_log",
		"commit",
		"apply",
		"pull",
		"push"
	],
	"live_panel_output": true,
	"tab_order":
	[
		"status",
		"branch",
		"graph",
		"rebase"
	],
	"show_stdin_in_output": true,
	"pedantic_commit": true,
  "show_full_commit_info": true,
}
```

randy3k:
I also have the double output issue when `live_panel_output` is on.

randy3k:
Are we ready to make a release?

kaste:
Iirc the diffview has too many bugs in dev and is annoying to use. Especially cursor jumps after hunking.

I would recommend you take everything up to 49a82d51cc730a5737304128c584edfe6def8f06 from my PR. This should be uncontroversial bc it does not contain any bigger refactorings. It mostly fixes cursor jumping, adds undo.

randy3k:
Perhaps, you should remove the commits after 49a82d5 from #1075 and prepare a separate PR for the rest. We could then review the trimmed PR and merge it into the release.

kaste:
That's probably a good idea, @randy3k

randy3k:
This release keeps rolling bigger because of the bugfixes @kaste pushes every day. 😄

asfaltboy:
I would release it as is right now, to give the pending fixes to users and allow more wide-spread testing.

randy3k:
@asfaltboy I will make a release once I get an approval

PR #1073, opened 2018-12-25
Fix: branch graph for selected

### LINKED ISSUE #1072: Bug: branch interface doesn't open graph for selected branch
When in branch interface, hit `[g]` to open graph under for the branch under the cursor. Instead, the graph view opens with the checked out branch, not the branch under cursor.

- Sublime Text 3 3184
- Git 2.19.1
- OS type and version 10.14.1 (18B75)
- Console error output -nothing

Git bisect points at 6a6c612 as the culprit, and we can see the `LogGraphMixin` command no longer uses `get_graph_args`, and users of the mixin need to use `prepare_target_view` to set additional graph arguments as view settings.

asfaltboy:
This regression was caused by a refactor to LogGraphMixin (more details in related ticket)

fix #1072

### DIFF
--- a/core/interfaces/branch.py
+++ b/core/interfaces/branch.py
@@ -658,7 +658,5 @@ class GsBranchesLogGraphCommand(LogGraphMixin, TextCommand, GitCommand):
         self._file_path = None
         super().run_async()
 
-    def get_graph_args(self):
-        args = super().get_graph_args()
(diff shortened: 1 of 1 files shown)


randy3k:
Thanks

randy3k:
This will be also cherry picked to the next release branch.

PR #1075, opened 2018-12-28
Implement Undo for diff-view

### LINKED ISSUE #1036: Add undo for diff view
We have it for inline-diff. The same for the diff view should be 👍, no?

### LINKED ISSUE #1074: Two navigation issues of the diff view [2.18.0]
I have been testing with the next release (2.18.0) branch and note that there are two minor navigation issues of the new diff view.

ezgif-1-59060baee699

1. Just note that when `cmd+enter` was hit over the first hunk, the cursor jumps to the third hunk rather to the second hunk.

2. When `tab` was hit to enter the cached diff, the cursor was not positioned at at the first hunk.

cc: @kaste

kaste:
Fixes #1036
Fixes #1074 

Bugs/Observations: 

- [ ] ~~can't double click into a new file ('@@ -16,7 +16 @@ ...')~~ Now in #1091 
- [x] just checking the hunk start ('@@ ...') for a good jump position isn't good enough. This line is just not stable enough. I will instead search for the actual hunk content.
- [x] cursor can still stay in meaningless land if the diff was actually empty before
- [x] cursor can land at eof which does not count as hunk so the user can't [h]unk it
- [ ] jump to file from hunk is not 100% implemented
- [ ] hunking with multiple cursor is cheaply implemented. If you have two cursors in the same hunk it throws (of course 🙄), otherwise it calls git n times.
- [ ] [H]unk complete file, or from this hunk to end of file functionality is clearly missing
- [ ] Also: [D]iscard complete file
- [ ] Maybe: Navigate to next/previous [f]ile
- [ ] If you staged, and then [c]ommit the commit view closes, and the diff view gains focus. It should probably close as well iff there is nothing left to stage
- [ ] ~~Prelude A..B shows target..base instead of base..target~~ Now in #1090 
- [ ] Ensure cursor stays in hunk position after hunking last hunk of a file and then therefore going to the next hunk in the next file is required
- [ ] Collapse multiple cursors to one after hunking
- [ ] Allow fast opening of `gs_diff`'s for current file (e.g. from the status file) which is broken right now

### DIFF
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -299,9 +299,7 @@ class GsDiffStageOrResetHunkCommand(TextCommand, GitCommand):
                 stdin=hunk_diff
             )
 
-        sublime.set_timeout_async(
-            lambda: self.view.run_command("gs_diff_refresh", {'navigate_to_next_hunk': True})
-        )
+        sublime.set_timeout_async(lambda: self.view.run_command("gs_diff_refresh"))
 
     def get_hunk_diff(self, pt):
         """
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -235,11 +235,10 @@ class GsDiffToggleCachedMode(TextCommand):
 
         self.view.run_command("gs_diff_refresh")
 
-        if self.view.settings().get("git_savvy.diff_view.just_hunked"):
-            self.view.settings().set("git_savvy.diff_view.just_hunked", False)
-            history = self.view.settings().get("git_savvy.diff_view.history")
-            _, stdin, _ = history[-1]
-            match = re.search(r'\n(@@ .+)\n', stdin)
+        just_hunked = self.view.settings().get("git_savvy.diff_view.just_hunked")
+        if just_hunked:
+            self.view.settings().set("git_savvy.diff_view.just_hunked", "")
+            match = re.search(r'\n(@@ .+)\n', just_hunked)
             if match is not None:
                 expected_content = match.group(1)
                 region = self.view.find(expected_content, 0, sublime.LITERAL)
@@ -343,7 +342,7 @@ class GsDiffStageOrResetHunkCommand(TextCommand, GitCommand):
             history = self.view.settings().get("git_savvy.diff_view.history") or []
             history.append((args, hunk_diff, pt))
             self.view.settings().set("git_savvy.diff_view.history", history)
-            self.view.settings().set("git_savvy.diff_view.just_hunked", True)
+            self.view.settings().set("git_savvy.diff_view.just_hunked", hunk_diff)
 
         sublime.set_timeout_async(lambda: self.view.run_command("gs_diff_refresh"))
 
@@ -469,6 +468,7 @@ class GsDiffUndo(TextCommand, GitCommand):
 
         self.git(*args, stdin=stdin)
         self.view.settings().set("git_savvy.diff_view.history", history)
+        self.view.settings().set("git_savvy.diff_view.just_hunked", stdin)
 
         self.view.run_command("gs_diff_refresh")
         self.view.sel().clear()
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -3,6 +3,7 @@ Implements a special view to visualize and stage pieces of a project's
 current diff.
 """
 
+from contextlib import contextmanager
 import os
 import re
 import bisect
@@ -255,9 +256,24 @@ class GsDiffToggleCachedMode(TextCommand):
             sel.clear()
             for (a, b) in last_cursors:
                 sel.add(sublime.Region(a, b))
-            self.view.show(sel)
+
+            # The 'flipping' between the two states should be as fast as possible and
+            # without visual clutter.
+            with no_animations():
+                self.view.show(sel)
         else:
             self.view.run_command("gs_diff_navigate")
+
+
+@contextmanager
+def no_animations():
+    pref = sublime.load_settings("Preferences.sublime-settings")
+    current = pref.get("animation_enabled")
+    pref.set("animation_enabled", False)
+    try:
+        yield
+    finally:
+        pref.set("animation_enabled", current)
 
 
 class GsDiffFocusEventListener(EventListener):
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -207,6 +207,7 @@ class GsDiffToggleCachedMode(TextCommand):
     Toggle `in_cached_mode`.
     """
 
+    # NOTE: MUST NOT be async, otherwise `view.show` will not update the view 100%!
     def run(self, edit):
         setting = 'in_cached_mode'
 
@@ -453,10 +454,8 @@ class GsDiffUndo(TextCommand, GitCommand):
     Undo the last action taken in the diff view, if possible.
     """
 
+    # NOTE: MUST NOT be async, otherwise `view.show` will not update the view 100%!
     def run(self, edit):
-        sublime.set_timeout_async(self.run_async)
-
-    def run_async(self):
         history = self.view.settings().get("git_savvy.diff_view.history") or []
         if not history:
             window = self.view.window()
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -4,6 +4,7 @@ current diff.
 """
 
 from contextlib import contextmanager
+from itertools import dropwhile, takewhile
 import os
 import re
 import bisect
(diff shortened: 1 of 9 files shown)


kaste:
Basically just to make debugging #1074 easier. 😁

kaste:
Added three commits that handle cursor positioning a) after hunking, b) after undoing, and after switching between stage/unstage areas. 

This is merely iteration code since (at least I) don't exactly know what 'good, expected behavior' looks and feels like.

asfaltboy:
Works for me, but I just have to ask you, why do we have both inline-diff and diff views implementing hunk stage/reset shortcuts? Is the inline diff only useful for line-by-line staging?

core/commands/diff.py
```diff
@@ -213,12 +213,29 @@ def run(self, edit, setting):
...
 
         self.view.run_command("gs_diff_refresh")
-        if last_cursors:
-            sel = self.view.sel()
-            sel.clear()
-            for (a, b) in last_cursors:
-                sel.add(sublime.Region(a, b))
-            self.view.show(sel)
+        if setting == 'in_cached_mode':
+            if self.view.settings().get("git_savvy.diff_view.just_hunked"):
```
asfaltboy: Just from a stylistic viewpoint, seems like this behaviour along with `last_cursors` getter/setter on lines 207-209 belong together. For readability, I would recommend moving these together into their own method (e.g `move_to_hunk`).

randy3k:
To be honest, I seldom use the inline mode. No comments

core/commands/diff.py
kaste: We have the `if setting == 'in_cached_mode':` three times in this function so it is better to extract this functionality https://github.com/divmain/GitSavvy/pull/1075/commits/a6c67faa56bbb57acefb1a70957594523f32e428 to actually reduce complexity.

kaste:
Do you guys actually do code reviews, 😀. 

- You couldn't undo if you used multiple cursors. https://github.com/divmain/GitSavvy/pull/1075/commits/e34b58d7d205f3d234dc8bef261e0af415b6afa6

- <hunk, undo, tab> crashed https://github.com/divmain/GitSavvy/pull/1075/commits/9701c9f98ae44a6ddc1af5ff0053564294f5374f

randy3k:
Ha...I only tested the basic operations..

asfaltboy:
Same here, did one stage-hunk and one undo ☺️, good catch @kaste

kaste:
Don't forget to remind me about updating the tooltip help. 🙄

core/commands/diff.py
```diff
@@ -395,10 +453,37 @@ class GsDiffNavigateCommand(GsNavigate):
...
+        args[1] = "-R" if not args[1] else None
+
+        self.git(*args, stdin=stdin)
+        self.view.settings().set("git_savvy.diff_view.history", history)
+        self.view.settings().set("git_savvy.diff_view.just_hunked", stdin)
+
+        self.view.run_command("gs_diff_refresh")
+        self.view.sel().clear()
+        self.view.sel().add(cursor)
+        self.view.show(cursor)
```
kaste: FIXME: `cursor` is only valid if the user is still on the same staging area, t.i. did not '<tab>'.

kaste:
I have visual bugs right now. The actual selections (cursors) are correct but the visual cursor is on the wrong line and/or the highlighted current line is not in sync with the visible cursor. (On cursor move up/down everything jumps to the correct position.) Sublime bug 🐛 probably.

kaste:
Okay, you cannot do `view.show` on the worker, so either the undo command must be sync or we invent a separate command `gs_move_cursor` which will just update the cursors. (We do the latter in SublimeLinter.)

kaste:
Another 🐛 : If you hit 'h' fast enough I get a 'patch does not apply' error.

kaste:
🤞 fixed the bugs

Another enhancement: Show `file_path` in title if given. 💃

kaste:
Other bugs:

- the diff view now shows line numbers etc
- when a diff view is already open, invoking the 'diff' command switches to the view, but then `refresh` runs twice, and the cursor always moves to the first hunk

kaste:
Hm, probably time to rewrite/refactor.

kaste:
Update: AWAIT_WORKER is that fast because on Travis/Linux `on_activated_async` does not get called. With print debugging I get e.g. on Travis/OSX

```
test_create_new_diff_view (test_diff_view.TestDiffView) ... --> on_activated_async 26
--> not a diff_view 26
--> new diff_view 27
--> end gs_diff 27
--> on_activated_async 27
ok
test_extract_clickable_lines (test_diff_view.TestDiffView) ... --> new diff_view 29
--> end gs_diff 29
--> on_activated_async 28
--> not a diff_view 28
--> on_activated_async 29
ok
test_result_file_regex (test_diff_view.TestDiffView) ... --> new diff_view 31
--> end gs_diff 31
--> on_activated_async 30
--> not a diff_view 30
Package Control: Skipping automatic upgrade, last run at 2019-01-07 11:02:16, next run at 2019-01-07 12:02:16 or after
--> on_activated_async 31
ok
test_reuse_open_diff_view (test_diff_view.TestDiffView) ... --> on_activated_async 32
--> not a diff_view 32
--> on_activated_async 33
--> not a diff_view 33
--> new diff_view 34
--> end gs_diff 34
--> on_activated_async 34
--> on_activated_async 33
--> not a diff_view 33
--> on_activated_async 34
ok
```

But on Linux I get 

```
test_create_new_diff_view (test_diff_view.TestDiffView) ... --> new diff_view 243
--> end gs_diff 243
FAIL
test_extract_clickable_lines (test_diff_view.TestDiffView) ... --> new diff_view 245
--> end gs_diff 245
FAIL
test_result_file_regex (test_diff_view.TestDiffView) ... --> new diff_view 247
--> end gs_diff 247
FAIL
test_reuse_open_diff_view (test_diff_view.TestDiffView) ... --> new diff_view 250
--> end gs_diff 250
FAIL
```

Is this a Travis thing, or a Sublime bug? @randy3k

randy3k:
@kaste I am not sure. Perhaps one needs to click on the sublime window on Linux to activate `on_activated_async`.

kaste:
How should I click on a window? Reduced testcase https://github.com/kaste/on_activated_testcase with https://travis-ci.com/kaste/on_activated_testcase/builds/96493957

randy3k:
I maybe wrong. It was just a guess

asfaltboy:
I'm sorry @kaste but I lost track with what's going on here, 68 commits? 17 files changed? Isn't this PR about adding undo to diff? Adding more changes generally will introduce more uncaught regressions and is harder to review... any chance you could refactor this change to a couple of simpler/specific PRs?

kaste:
Well, yeah it looks like a mess, and that wasn't the plan but there are/were just too many bugs. Now, iirc, https://github.com/divmain/GitSavvy/pull/1075/commits/49a82d51cc730a5737304128c584edfe6def8f06 was a good enough commit, and even though I started on my local branch some refactoring I actually waited before pushing it long enough. But nobody said anything. And if nobody leads you, you have to do it yourself. I was actually unsure if I should push WIP commits here, but _usually_ you push your work at the evening and don't hide it until it's ready and polished. As you can see I at least made a nice bullet list of bugs/observations so it was basically clear from the beginning that I want to check as many items as possible. 

Now (again), this went out basically as a rewrite of the diff view (too many bugs) so the usual review rules do not apply anyway. But the code also isn't ready yet for review. I invite you all to check this branch and just try it and write the bugs down. UX confusion and annoyances count as bugs.

The cursor jumping and/or scrolling is still not okay. I probably have it working locally in a stash.

kaste:
Okay, a lot of good stuff in here in just 20 commits. 

This is mostly implementing the undo and nothing else. Some care has been taken that the cursor/view stays stable or in useful land when opening the diff view, hunking, undoing, or switching. This is still not 100% though. 

As you can see I unchecked most items of the todo list, so there is def stuff that comes after this.

core/commands/diff.py
```diff
@@ -174,51 +178,165 @@ def run(self, edit, cursors=None, navigate_to_next_hunk=False):
...
+        sel.add(cursors)
+    else:
+        for c in it:
+            sel.add(c)
+
+    view.show(sel)
+
+
+@contextmanager
+def no_animations():
```
kaste: In a later version I will get rid of this side-effect by just storing and restoring the `viewport_position`.

core/commands/diff.py
```diff
@@ -174,51 +178,165 @@ def run(self, edit, cursors=None, navigate_to_next_hunk=False):
...
+        setting = 'in_cached_mode'
+
         if (
-            setting == 'in_cached_mode'
-            and self.view.settings().get("git_savvy.diff_view.base_commit")
+            self.view.settings().get("git_savvy.diff_view.base_commit")
             and self.view.settings().get("git_savvy.diff_view.target_commit")
         ):
             # There is no cached mode if you diff between two commits, so
             # we need to abort here
```
kaste: No need to forbid this here, we can exchange `base` with `target` making it essentially a `-R`everse diff.

randy3k:
Awesome. Merged.

kaste:
Woot. 🎉

PR #1078, opened 2019-01-07
Edge case for initial commit created e.g. by GitHub

kaste:
If you create a new repo on GitHub and clone it immediately `master` tracks
`origin/master` without creating it. So the remote is 'gone'.


randy3k:
I didn't try. Does this still work if we create a repo locally without remote?

PR #1079, opened 2019-01-07
Do not imperatively refresh ui after Log actions

kaste:
The three calls herein used `self.view` which is not defined since
`GsLogActionCommand` is a `WindowCommand`. So basically all commands raised
and did nothing.

I think doing nothing is the correct approach here since after closing the
panel, `on_activated` callbacks run for all important views anyway.

To actually refresh _other_ visible views beside the active one would be a
useful feature though.


randy3k:
Thanks. Good catch

PR #1081, opened 2019-01-13
Add symlink support on Windows

deathaxe:
The Issue
=========

Git for Windows does not enable support for symlinks by default as it
requires administrative privileges and causes a couple of edge cases.

Hence calling `git rev-parse --show-toplevel` is not able to resolve
junctions or symlinks and does not return the real repository path, if
the working directory is accessed via junction/symlink.

Another issue is caused by python itself, as its `os.path.realpath()`
function does not resolve junctions/symlinks on Windows as well.

The Motivation
==============

When working on `sublimehq/Packages` the only way to test changes of
single syntaxes is to create a junction of a Package (e.g. Python) in
ST's `Data/Packages` path. Otherwise syntax tests fail and changes of
the syntax-definition or any other module are not tracked by ST.

There may be other use cases for working with symlinks.

The Fix
=======

This commit enables GitSavvy to support this kind of workflow by
borrowing the `realpath patch` and the algorithm to find the repo's
toplevel path from GitGutter.

The fix consists of two parts:

1. A patched `realpath()` function, which is applied if Windows
   supports the underlying `_getfinalpathname()` function.

   On MacOS/Linux the default `os.path.realpath()` is called as it
   already supports symlinks.

2. A local alternative for the `git rev-parse --show-toplevel` command
   which resolves the repository path without calling any git command.

As a result any git command is now called with the real repository
path even if the working directory is accessed via a junction/symlink on all OS's.

### DIFF
--- a/common/util/__init__.py
+++ b/common/util/__init__.py
@@ -8,6 +8,7 @@ from . import log
 from . import actions
 from . import debug
 from . import diff_string
+from . import path
 from . import reload
 
 super_key = "SUPER" if sys.platform == "darwin" else "CTRL"
--- a/common/util/path.py
+++ b/common/util/path.py
@@ -18,25 +18,25 @@ try:
 
             Also note that _getfinalpathname in Python 3.3 throws
             `NotImplementedError` on Windows versions prior to Windows Vista,
-            hence we fallback to `os.path.abspath()` on these platforms.
+            hence we fallback to `os.path.realpath()` on these platforms.
 
         Arguments:
             path (string): The path to resolve.
 
         Returns:
             string: The resolved absolute path if exists or path as provided
-                otherwise.
+                otherwise. If `path` is '' or None the current directory is
+                returned.
         """
-        try:
-            if path:
+        if path:
+            try:
                 real_path = _getfinalpathname(path)
                 if real_path[5] == ':':
                     # Remove \\?\ from beginning of resolved path
                     return real_path[4:]
-                return os.path.abspath(path)
-        except FileNotFoundError:
-            pass
-        return path
+            except FileNotFoundError:
+                return path
+        return os.path.realpath(path)
 
 except (AttributeError, ImportError, AssertionError):
     def realpath(path):
@@ -48,7 +48,7 @@ except (AttributeError, ImportError, AssertionError):
         Returns:
             string: The resolved absolute path.
         """
-        return os.path.realpath(path) if path else None
+        return os.path.realpath(path)
 
 
 def is_work_tree(path):
--- a/common/util/path.py
+++ b/common/util/path.py
@@ -1,0 +1,93 @@
+import os.path
+
+try:
+    from nt import _getfinalpathname
+    from sys import getwindowsversion
+    assert getwindowsversion().major >= 6
+
+    def realpath(path):
+        """Resolve symlinks and return real path to file.
+
+        Note:
+            This is a fix for the issue of `os.path.realpath()` not to resolve
+            symlinks on Windows as it is an alias to `os.path.abspath()` only.
+            see: http://bugs.python.org/issue9949
+
+            This fix applies to local paths only as symlinks are not resolved
+            by _getfinalpathname on network drives anyway.
+
+            Also note that _getfinalpathname in Python 3.3 throws
+            `NotImplementedError` on Windows versions prior to Windows Vista,
+            hence we fallback to `os.path.abspath()` on these platforms.
+
+        Arguments:
+            path (string): The path to resolve.
+
+        Returns:
+            string: The resolved absolute path if exists or path as provided
+                otherwise.
+        """
+        try:
+            if path:
+                real_path = _getfinalpathname(path)
+                if real_path[5] == ':':
+                    # Remove \\?\ from beginning of resolved path
+                    return real_path[4:]
+                return os.path.abspath(path)
+        except FileNotFoundError:
+            pass
+        return path
+
+except (AttributeError, ImportError, AssertionError):
+    def realpath(path):
+        """Resolve symlinks and return real path to file.
+
+        Arguments:
+            path (string): The path to resolve.
+
+        Returns:
+            string: The resolved absolute path.
+        """
+        return os.path.realpath(path) if path else None
+
+
+def is_work_tree(path):
+    """Check if 'path' is a valid git working tree.
+
+    A working tree contains a `.git` directory or file.
+
+    Arguments:
+        path (string): The path to check.
+
+    Returns:
+        bool: True if path contains a '.git'
+    """
+    return path and os.path.exists(os.path.join(path, '.git'))
+
+
+def split_work_tree(file_path):
+    """Split the 'file_path' into working tree and relative path.
+
+    The file_path is converted to a absolute real path and split into
+    the working tree part and relative path part.
+
+    Note:
+        This is a local alternative to calling the git command:
+
+            git rev-parse --show-toplevel
+
+    Arguments:
+        file_path (string): Absolute path to a file.
+
+    Returns:
+        A tuple of two the elements (working tree, file path).
+    """
+    if file_path:
+        path, name = os.path.split(file_path)
+        # files within '.git' path are not part of a work tree
+        while path and name and name != '.git':
+            if is_work_tree(path):
+                return (path, os.path.relpath(
+                    file_path, path).replace('\\', '/'))
+            path, name = os.path.split(path)
+    return (None, None)
--- a/common/util/path.py
+++ b/common/util/path.py
@@ -35,7 +35,7 @@ try:
                     # Remove \\?\ from beginning of resolved path
                     return real_path[4:]
             except FileNotFoundError:
-                return path
+                pass
         return os.path.realpath(path)
 
 except (AttributeError, ImportError, AssertionError):
--- a/core/commands/init.py
+++ b/core/commands/init.py
@@ -105,7 +105,7 @@ class GsClone(WindowCommand, GitCommand):
             if not os.path.exists(os.path.join(folder, project, ".git")):
                 return os.path.join(folder, project)
             else:
-                parent = os.path.realpath(os.path.join(folder, ".."))
+                parent = util.path.realpath(os.path.join(folder, ".."))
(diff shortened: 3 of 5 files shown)


randy3k:
@kaste you are a better one to review it because you own a Windows.

kaste:
I would need better test instructions here. What I don't like about the change is that is presumes it replaces `os.path.realpath` but then doesn't exactly work like that. 

For the edge cases: 

```
>>> os.path.realpath(None)
'C:\\Dev\\Sublime Text 3'
>>> os.path.realpath('')
'C:\\Dev\\Sublime Text 3'
```

This makes it a 👎 when you exchange the `realpath` implementation in the whole repo. The 👍 is of course that we get rid of the `git rev-aprse` call.

Generally note that you should be able to set 

```
	settings:
		git_savyy.repo_path: "my/path/to/.git"
```

in your project settings to force the git repo used for all views.

I also don't know yet if the failing test is a fluke or introduced by the change. On my computer I sure get

```
>>> view.find_all_results()
[('/C/Users/c-flo/AppData/Roaming/Sublime Text 3/Packages/GitSavvy/core/commands/log.py', 0, 0), ('/C/Users/c-flo/AppData/Roaming/Sublime Text 3/Packages/GitSavvy/core/interfaces/branch.py', 0, 0)]
```

which is what I described as pseudo-linux paths.

kaste:
The failing test is introduced by this PR. 

```
--> repo_path /not/here
--> util.path.realpath(repo_path) /not/here
--> os.path.realpath(repo_path) C:\not\here
```

This is relevant bc we do `view.settings().set("result_base_dir", self.repo_path)` hence the failing test. (Note: the feature still works it just exposes the different relaptah implementation.

deathaxe:
> Generally note that you should be able to set

	settings:
		git_savyy.repo_path: "my/path/to/.git"

Generally not about the note: This is useless, if the whole _Package_ path is open as project and contains dozens of repositories!

kaste:
GitSavvy also uses e.g. `os.path.join(repo_path, x)` and `os.path.relpath(x, repo_path)`. How should these work? Why do you want this as a replacement of `realpath` and not as a different version of a `get_repo_path()` function?

deathaxe:
A side note / question:

Just added a print to the `common.util.realpath()` and called `git: status` once. Looks like (without this PR) `git rev-parse` is called 8 times just to show the dashboard? This explains why it is so horrible slow.

kaste:
Okay, let's review this. The PR proposes actually two 'fixes'. 

- It tries to implement a 'better' `os.path.realpath` which follows linux semantics bc on Windows `realpath` is actually an alias for `abspath` and does not resolve symlinks et.al.   
- It tries to implement an own version for `git rev-parse --show-toplevel` again (probably) to make the Windows version behave more like the linux version. 

Both patches are actually independent from each other bc if we pull your `realpath` the current `find_git_repo` would work (bc it would start e.g. not from `dirname(view.file_name())` but from its 'real' directory). 

So I assume the `find_git_repo` patch is for speed and to avoid an 'unnecessary' popen call.  Although *I* think your algo is what git actually does here, I don't find positive evidence that this is the case. All I can find is the recommendation to use `git rev-parse --show-toplevel`, and this is in fact what VSCode et.al. do. I don't have any domain knowledge here about the edge cases to find the right git repo (except for example you do obviously not check `GIT_DIR`) so *I* would *NOT* pull this change. 

*I* also would *NOT* pull the `realpath` patch as it is bc it breaks the UI of GitSavvy and breaks my Windows feeling. Actually all my projects/directories are remapped using junctions or `subst`. So I use this daily. E.g. I `subst` a deep folder of my Dropbox account to `D:\`. In this case I map complete repos as directories to a different pseudo drive (D) and work on this. E.g. I have `D:\mockito-python` here with its `D:\mockito-python\.git`. 

I think it is confusing that the status dashboard now shows `C:\Users\blah\Documents\Dropbox\...\mockito-python`. It is though completely unacceptable that GitSavvy now with your patch opens these 'real' files if I double click on a filename, or [o]pen it from the diff view etc. As a user, I work on the shorter `D:\mockito-python` and that's the whole purpose of the mapping features. (And these features work okay except for Sublime plugins bc actually Sublime does not follow junctions and will thus not reload on save. So only in that case, e.g. for SublimeLinter I actually work with these stupid long `%APPDATA%/...` paths otherwise I would have mapped my plugins to e.g. `D:\Sublime-stuff`.)

Generally, it is okay and probably necessary that _internally_ some paths are resolved via `realpath`. But 'real' paths are implementation details and should not be presented to the end user. For the concrete implementation (`nt._getfinalpathname`) I lack again deeper domain knowledge. The typical test cases here are: Does it work as intended with obscure utf-8 names? How does it handle network mappings etc. (not at all probably)? As your patch always fallbacks on errors this might be good enough though.

deathaxe:
Your initial comments basically express the intention of this PR.

> Both patches are actually independent from each other bc if we pull your realpath the current find_git_repo would work (bc it would start e.g. not from dirname(view.file_name()) but from its 'real' directory).

This is actually wrong, because `dirname(view.file_name())` does not resolve symlinks. Therefore `git rev-parse --show-toplevel` is invoked with a path it can't find the repository-toplevel path on Windows, if it contains symlinks/junctions.

_Example:_

Assume, we have _%APPDATA%\Sublime Text 3\Packages\Erlang\Erlang.sublime-syntax_ open. So `find_git_repo()` calls `find_git_toplevel("%APPDATA%\Sublime Text 3\Packages\Erlang")`. This is the entry point for `git rev-parse --show-toplevel` then. As the git repository is located at _C:\Data\sublimehq\Packages_, git fails finding it and nothoing would change = GitSavvy asks to init a repo.

_Note:_

To fix that issue and turn your statement into a true one, we would need to translate the path via `common.util.realpath()` before calling `git rev-parse --show-toplevel`.

> So I assume the find_git_repo patch is for speed and to avoid an 'unnecessary' popen call. ... I don't have any domain knowledge here about the edge cases to find the right git repo (except for example you do obviously not check GIT_DIR) so I would NOT pull this change.

This is included in this PR as it is bundled in GitGutter as well, which uses both for years now without any complaints. The repo-path was resolved using the `split_work_tree()` function from the beginning on. GitGutter never used `git rev-parse --top-level`. The realpath patch was introduced as a solution for https://github.com/jisaacks/GitGutter/issues/373, as already stated. It works perfectly for nearly 2 years now.

I can just guess `split_work_tree()` was invented for performance reasons. It already existed, when I started maintaining GitGutter.


> I also would NOT pull the realpath patch as it is bc it breaks the UI of GitSavvy and breaks my Windows feeling. Actually all my projects/directories are remapped using junctions or subst. So I use this daily. E.g. I subst a deep folder of my Dropbox account to D:\. In this case I map complete repos as directories to a different pseudo drive (D) and work on this. E.g. I have D:\mockito-python here with its D:\mockito-python\.git.

As you already correclty stated before, the `realpath-patch` makes GitSavvy behave the same way as on Linux. What you describe as "breaks my Windows feeling" is the default behavior on Linux though. The following screenshot is taken from a Ubuntu-VM and the current GitSavvy 2.18. What you see is /home/deathaxe/GitSavvy/core being added as folder to Sublime Text 3, while _GitSavvy_ is actually a symlink to /home/deathaxe/.config/sublime-text-3/Packages/GitSavvy.

grafik

I absolutely follow in the question whether this is desireable or not. But this would require another fix to solve that issue for both Windows and Linux in the same way, because the "inconsistency in behavior" currently causes Linux users to already see what you complain about to be introduced by this PR.

> ... except for Sublime plugins bc actually Sublime does not follow junctions and will thus not reload on save

This statement is actually wrong for ST3176+. It handles junctions well and reloads resources on save. The issue just is, that we can't place the whole Packages repository into Sublime Text to do syntax developement, but need to create junctions for single folders within the whole repo.

> Generally, it is okay and probably necessary that internally some paths are resolved via realpath. But 'real' paths are implementation details and should not be presented to the end user.

This is generelly true, but like I already said, Linux/Mac users see this by default. If we don't like it, we need to find a dedicated solution.


> The typical test cases here are: Does it work as intended with obscure utf-8 names? 

The `nt._getfinalpathname()` was introduced with Windows Vista. I'd expect it to handle unicode and long (>256chars) filenames as well as any other (modern) Windows API function does.

> How does it handle network mappings etc. (not at all probably)? 

Windows filesystems (samba) don't support symlinks/junctions. Hence there is nothing to resolve. This fix applies for local files only by design and intent.

kaste:
I don't know if this break the 'linux feeling' as much as it breaks the 'windows feeling' bc *nix people use symlinks more frequently. I also don't have a clear code idea how to improve the linux behavior here. 

For the `realpath` patch, pathlib's `resolve` method uses basically the same approach (but supports UNC paths as well). So `_getfinalpathname` must be considered stable.

randy3k:
I have to confess that I didn't read the discussion above. I guess we just need more opinions.

kaste:
Generally, we cannot pull this because it regresses GitSavvy on Windows. It solves one problem but breaks the UI. As noted, GitSavvy on *nix has a degraded experience compared to Windows but there are also other parts of GitSavvy (unrelated to this PR's code) where the *nix and Windows UX differ.

To solve the issue here at hand, we need multiple, non-trivial preliminary PRs.

kaste:
@deathaxe I thought of this back and forth. Maybe we can do a 80% solution as follows.

On *nix systems, we call '--show-toplevel' which gives us resolved path. We also traverse up in python which might give us a path containing '.git' in some parent folder. Then we compare both and maybe take the unresolved folder bc it it is a better match for the UI.

```
# *nix
start_dir = ...
resolved_repopath = git('--show-toplevel', working_dir=start_dir)
if not resolved_repopath:
    return

unresolved_repopath = find_nearest('.git', start_dir)
if unresolved_repopath and os.path.realpath(unresolved_repopath) == resolved_repopath:
    return unresolved_repopath

return resolved_repopath

```

On Windows, '--show-toplevel' gives us an unresolved path which is usually what we want except for the edge case described in this issue. 


```
# Windows
return (
    git('--show-toplevel', working_dir=start_dir) 
    or find_nearest('.git', resolved(start_dir)) 
    or None
)
```

PR #1082, opened 2019-01-13
Add support for "env" settings and GIT_OPTIONAL_LOCKS

deathaxe:
Git 2.17+ supports the custom environment variable "GIT_OPTIONAL_LOCKS"
which can be set by editors to ask git to reduce the number of file
locks when querying for status information such as "git status". This
helps to speed up some commands and helps improving concurrent status
queries.

This commit therefore introduces the "env" setting in the 
GitSavvy.sublime-setting, which is injected into the "environ" object
before calling a git command.

The "env" setting contains the "GIT_OPTIONAL LOCKS: 0" by default.
A user can override this setting or add other custom environment
variables.

Note:

Newer versions of git also support a command line argument to disable
optional locks. But using it would cause backward compatibility issues
if a user still works with a git version, which does not support it.

time.

### DIFF
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -23,6 +23,13 @@
     "git_path": "",
 
     /*
+        Set custom environment variables for git command execution.
+    */
+    "env": {
+        "GIT_OPTIONAL_LOCKS": "0"
+    },
+
+    /*
         Change this to `true` when doing dev work on GitSavvy.
      */
     "dev_mode": false,
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -26,7 +26,6 @@
         Set custom environment variables for git command execution.
     */
     "env": {
-        "GIT_OPTIONAL_LOCKS": "0"
     },
 
     /*
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -187,6 +187,9 @@ class GitCommand(StatusMixin,
                 startupinfo.dwFlags |= subprocess.STARTF_USESHOWWINDOW
 
             environ = os.environ.copy()
+            savvy_env = self.savvy_settings.get("env")
+            if savvy_env:
+                environ.update(savvy_env)
             environ.update(custom_environ or {})
             start = time.time()
             p = subprocess.Popen(command,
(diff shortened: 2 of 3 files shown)


kaste:
There is not much information about `GIT_OPTIONAL_LOCKS` I can find. Why or which commands should make this faster?

This makes `git status` calls slower, no? Bc it doesn't write anything it computes back to the cache. (Bc it doesn't hold the lock.) 

Should we only apply this to the `git status` calls? Currently we're are not multi-threaded in GitSavvy land so clashes on the index.lock should be rare.

deathaxe:
GitSavvy is not the only plugin which uses git as a background task. So it is easily possible several plugins run git concurrently depending on ST events. If one of them locks the database for a longer time, this may cause hangs. VSCode propagated using this feature months ago to make things faster.

Why should it make `git status` calls slower? According to the docs it just prevents git from performing optional background operations. Guess something like `auto gc` operations and stuff.

```
       GIT_OPTIONAL_LOCKS
           If set to 0, Git will complete any requested operation without
           performing any optional sub-operations that require taking a
           lock. For example, this will prevent git status from refreshing
           the index as a side effect. This is useful for processes running
           in the background which do not want to cause lock contention with
           other operations on the repository. Defaults to 1.
```

This PR just adds the opportunity to define an additional environment in the settings, so everybody can decide, whether it is useful or not. GutGutter has used this settings for months without drawbacks.

It might be most useful with all git commands, which want to read information from the git database only. Maybe `merge`, `rebase`, `commit` should be excluded. Not sure how to do so without passing the `custom_environment` argument for all the others.

kaste:
Hm. From https://public-inbox.org/git/20170921043214.pyhdsrpy4omy54rm@sigill.intra.peff.net/

> Some of these commands may
refresh the index and write out the result in an
opportunistic way: if they can get the index lock, then they
update the on-disk index with any updates they find. And if
not, then their in-core refresh is lost and just has to be
recomputed by the next caller.

Later 


> 3. Ask "status" not to lock or write the 
This is easy to implement. The big downside is that any
     work done in refreshing the index for such a call is
     lost when the process exits. So a background process
     may end up re-hashing a changed file multiple times
     until the user runs a command that does an index
     refresh themselves.

FWIW VsCode only applies this for `git status` calls.

deathaxe:
Makes sense.

randy3k:
Speaking of the ordering of merging, `environ` should have a higher precedence than `savvy_env`.

PR #1083, opened 2019-01-16
Do not update status bar for 'widgets'

kaste:
Sublime widgets are e.g. the command palette, and the input box at the bottom.
For these views we generally do not need to update the status bar.

Views without a window are also now called transient. These are detached views
after close.

### DIFF
--- a/core/commands/status_bar.py
+++ b/core/commands/status_bar.py
@@ -27,6 +27,31 @@ last_execution = 0
 update_status_bar_soon = False
 
 
+def view_is_transient(view):
+    """Return whether a view can be considered 'transient'.
+
+    Transient views are 'detached' views, widgets (aka quick or input panels), and preview
+    views.
+    """
+
+    # 'Detached' (already closed) views don't have a window.
+    window = view.window()
+    if not window:
+        return True
+
+    # Widgets are normal views but the typical getter don't list them.
+    group, index = window.get_view_index(view)
+    if index == -1:
+        return True
+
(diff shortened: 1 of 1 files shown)


kaste:
If you enable logging, you can clearly see that we call `rev-parse toplevel` 5(!) times per `ctrl+p`.

randy3k:
It makes sense to me

asfaltboy:
Just gotta say, @kaste I ❤️ that you keep improving performance; bravo!

core/commands/status_bar.py
```diff
@@ -27,19 +27,29 @@ def on_post_save(self, view):
 update_status_bar_soon = False
 
 
+def view_is_transient(view):
```
asfaltboy: The description of this PR can go into this function's docstring

kaste: 🙄 	I skipped reading the mail after seeing the ❤️ 	above.

asfaltboy: Only noticed that this could be improved after sending the approval comment

kaste:
Here we also have a too many bugs case: The throttle function is not good enough. We overcall for new views and loading views. Also: for activating unnamed views we call `git toplevel` 5 times bc the repo path isn't cached.

kaste:
Another bug: For secondary views (clones) the status bar will not update after save.

randy3k:
Let's merge this first and address the bugs later.

randy3k:
another bug found: #1099

PR #1087, opened 2019-02-02
Fix diff view syntax settings

kaste:
Bc of the new prelude in the diff view, we at some point had to introduce a new syntax for it. But we forgot to add the typical syntax settings, and the edit settings shortcut.

This PR adds them.


kaste:
Extracted from #1036

PR #1098, opened 2019-02-03
Tags refactor

randy3k:
as discussed in #998, we are going to revert the previous commit which changes `tags` to `tag` but only keeping the user command.

### DIFF
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -338,7 +338,7 @@
         "command": "gs_view_git_log"
     },
     {
-        "caption": "git: tag",
+        "caption": "git: tags",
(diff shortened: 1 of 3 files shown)


randy3k:
@stoivo 
how does it look? I am cc'ing you because you merged the original commit.

randy3k:
@kaste Is this what you anticipated?

I think @stoivo and @asfaltboy may not be free.

kaste:
LGTM 👍

asfaltboy:
Looks fine to me btw, I am all for consistency

stoivo:
Yes I have been away.

randy3k:
Welcome back.

PR #1100, opened 2019-02-11
show status badge when there are no views

randy3k:
close #1099

PR #1101, opened 2019-02-11
Actually update the status bar for transient views

### LINKED ISSUE #1099: [dev] branch status badge no long is shown when there are no views
`git bisect` reveals that the first bad commit is 41387cc069b0eb33d888564267cb73fdcf36bb61

cc @kaste

kaste:
Fixes #1099

To what I know, your fix in #1100 is too complicated. Officially, transient views are exactly these views when you have an otherwise empty window. 

So I think we check the window anyway, then the group for the widgets and otherwise just return. 

Basically saying, it was a *feature* of my code that it was not calling `git status` for these temporary views. If we don't want that, I think we can just do this here.

### DIFF
--- a/core/commands/status_bar.py
+++ b/core/commands/status_bar.py
@@ -30,8 +30,8 @@ update_status_bar_soon = False
 def view_is_transient(view):
     """Return whether a view can be considered 'transient'.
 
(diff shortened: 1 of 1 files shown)


randy3k:
Perfect, it works.

PR #1107, opened 2019-02-14
Make hunking block for consistent subsequent actions

### LINKED ISSUE #1104: [diff view] Cannot `[hhh]unk` in a meaningful speed
If you hit 'hhh' to hunk 3 hunks in a row fast, you likely get a 'cannot apply patch' error from git.  

Basically, view and state are out of sync here. Typically 'hunking' must be blocking on the main thread. Or: 'hunking' must update the view optimistically. Since the latter is almost impossible to do right, we should use the 'blocking' strategy.

kaste:
We must ensure that hitting 'h' very fast does run completely ordered, and
always waits for 'gs_diff_refresh' to complete. ('gs_diff_refresh' updates our
model.)

Please NOTE that running a command from the worker will still run that command
on the main thread. T.i. `set_timeout_async(lambda: view.run_command("refresh"))`
is misleading and a footgun because the "refresh" command here will still start
on the main thread. We basically just await the worker and then put a
task/command on the main worker queue.

Fixes #1104

### DIFF
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -360,6 +360,10 @@ class GsDiffStageOrResetHunkCommand(TextCommand, GitCommand):
     hunk under the user's cursor(s).
     """
 
+    # NOTE: The whole command (including the view refresh) must be blocking otherwise
+    # the view and the repo state get out of sync and e.g. hitting 'h' very fast will
+    # result in errors.
+
     def run(self, edit, reset=False):
         ignore_whitespace = self.view.settings().get("git_savvy.diff_view.ignore_whitespace")
         show_word_diff = self.view.settings().get("git_savvy.diff_view.show_word_diff")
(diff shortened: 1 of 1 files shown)


randy3k:
I don't know if it is better to use sync operations or a lock.

kaste:
Sorry, don't know what you mean here. What are your concerns? 

'Make hunking block' is maybe a bit short. What I actually do here is make the hunk action *one* task. Do not split it up into multiple tasks. That is so that you can put 'hhh' three hunking tasks on the queue, and each of them runs like a function from start to end.

It would be a misunderstanding that it is *necessary* to run these tasks on the main thread then. It is totally possible to run the hunking - each as one block/task - on the worker or any other thread. The reason to run it on the main thread is purely for UI/UX. It just looks and feels nicer if you do it on the main thread because we avoid intermediate redraws and yank.

randy3k:
Ok, thank you for the explanations. I thought you wanted to wait for an action to be completed before triggering another h.  And I was thinking that a lock may do a better job in this aspect.

PR #1108, opened 2019-02-15
Optimize diff view startup

### LINKED ISSUE #1092: [diff view] Activating the diff view blocks main thread
The refresh command runs on the main thread and thus it blocks the UI. This is often desirable e.g. if you hunk or undo, but not so if you just activate the view e.g. via `ctrl+tab` action or mouse.

kaste:
GsDiffCommand either creates a new diff view, or it finds a suitable one
already open and then just focuses it. In both cases we want to make sure
that the view *eventually* shows the most recent diff.

The previous implementation had the following problems:

- `git diff` did run on the main thread (#1092)
- Reusing a diff view called `git diff` twice

We now generally rely on the `on_activated_async` event to update the view
lazy. The init command either creates or focuses a view but it never calls
for a 'refresh' imperatively.

Note that most of the time we actually want a blocking, sync update (to keep
the view and the repo state in sync) so we add a switch to the refresh command.

Fixes #1092

### DIFF
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -52,6 +52,8 @@ class GsDiffCommand(WindowCommand, GitCommand):
 
         if view_key in diff_views and diff_views[view_key] in sublime.active_window().views():
             diff_view = diff_views[view_key]
+            self.window.focus_view(diff_view)
+
         else:
             diff_view = util.view.get_scratch_view(self, "diff", read_only=True)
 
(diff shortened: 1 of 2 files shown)


kaste:
Current dev


Creating a view  (M=Main thread; D=Worker)

```
 (M) [125ms] $ git diff --stat --patch --no-color -- .\common\util\debug.py
 (D) [ 78ms] $ git status --porcelain -z -b
 (D) [ 94ms] $ git status --porcelain -z -b
```

Reusing a view

```
 (M) [109ms] $ git diff --stat --patch --no-color -- .\common\util\debug.py
 (M) [ 94ms] $ git diff --stat --patch --no-color -- .\common\util\debug.py
 (D) [ 94ms] $ git status --porcelain -z -b
```


PR

Creating 

```
 (D) [109ms] $ git status --porcelain -z -b
 (D) [109ms] $ git diff --stat --patch --no-color -- .\core\commands\diff.py
 (D) [109ms] $ git status --porcelain -z -b
```

Reusing

```
 (D) [ 84ms] $ git diff --stat --patch --no-color -- .\common\util\debug.py
 (D) [ 94ms] $ git status --porcelain -z -b
```

Aside: When creating, you can nicely see the status bar overcalls on new which means the implemented throttler is not good enough.

PR #1114, opened 2019-02-20
Do not provide both `file_path` and `current_file`

### LINKED ISSUE #1096: Diff multiple files via [e] from status dashboard doesn't work
- Make random edits in multiple (e.g. 3) files
- Goto status dashboard, select all files and hit [e]

Expected it opens e.g. 3 views. Actually, it opens two(!) views, both showing the diff of the first selected file.

kaste:
`current_file` basically takes precedence and fills `file_path` with
the current filename of the active view (using `self.file_path`). This
is not what we want here.

Fixes #1096

### DIFF
--- a/core/commands/log.py
+++ b/core/commands/log.py
@@ -198,7 +198,6 @@ class GsLogActionCommand(PanelActionMixin, WindowCommand, GitCommand):
         self.window.run_command("gs_diff", {
             "in_cached_mode": cache,
             "file_path": self._file_path,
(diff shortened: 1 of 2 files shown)


randy3k:
afaik, the argument is used here: https://github.com/divmain/GitSavvy/blob/4bb3b1e30d600d0db77e15f3364da901b6fc5d38/Default.sublime-commands#L256

kaste:
The issue here is that you MUST not use *both* arguments at the same time.

PR #1115, opened 2019-02-20
Do not close panel views when cycling dashboards

### LINKED ISSUE #1097: Cycling through dashboards hides any panel
Regardless which panel is open (e.g. just Sublime's console) if I hit `<tab>` it gets closed. I don't know why Savvy cares here.

kaste:
Fixes #1097

### DIFF
--- a/core/interfaces/__init__.py
+++ b/core/interfaces/__init__.py
@@ -35,7 +35,6 @@ class GsTabCycleCommand(TextCommand, GitCommand):
 
         window = self.view.window()
         if window:
-            window.run_command("hide_panel", {"cancel": True})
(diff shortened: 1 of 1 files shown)


randy3k:
> It believes it was used to hide the additional information panel of `git: graph`.
We may need a more specific solution rather than not doing `hide_panel`

randy3k:
#1116 replaces this

kaste:
🤔 no don't think so, currently \<tab>ing closes any panel which is resolved here. 

#1116 adds a fancy auto-close panel feature for the log graph only.

kaste:
Rebased. Since you pulled #1116 you should also pull this one.

PR #1118, opened 2019-02-25
Efficient hunking of multiple hunks

### LINKED ISSUE #1105: [diff view] Hunking multiple hunks is in-efficient
You can 'hunk' multiple 'hunks' using multiple cursors. The code here is still the first implementation. 

- If you have multiple cursors in the same hunk, it throws.
- Otherwise, it issues a separate `git apply` per hunk. This is also interferes with the undo feature bc you hit [h] once but then have to undo x times. 

Hunks can be merged to form a patch, and we can just do that.

kaste:
Fixes #1105 

Multiple hunks can be combined into a patch which then can applied at once.

### DIFF
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -5,7 +5,7 @@ current diff.
 
 from contextlib import contextmanager
 from functools import partial
-from itertools import dropwhile, takewhile
+from itertools import chain, dropwhile, takewhile
 import os
 import re
 import bisect
@@ -20,7 +20,7 @@ from ...common import util
 
 
 if False:
-    from typing import Callable, Iterator, List, Optional, Tuple, TypeVar
+    from typing import Callable, Iterable, Iterator, List, Optional, Tuple, TypeVar
     from mypy_extensions import TypedDict
 
     T = TypeVar('T')
@@ -357,6 +357,16 @@ def unpickle_sel(pickled_sel):
     return [sublime.Region(a, b) for a, b in pickled_sel]
 
 
+def unique(items):
+    # type: (Iterable[T]) -> List[T]
+    """Remove duplicate entries but remain sorted/ordered."""
+    rv = []  # type: List[T]
+    for item in items:
+        if item not in rv:
+            rv.append(item)
+    return rv
+
+
 def set_and_show_cursor(view, cursors):
     sel = view.sel()
     sel.clear()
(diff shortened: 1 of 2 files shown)


core/commands/diff.py
```diff
@@ -388,52 +398,57 @@ def run(self, edit, reset=False):
             set((self.view.size() + 1, ))
         ))
 
-        self.apply_diffs_for_pts(cursor_pts, reset)
+        filter_ = partial(filter, None)
+        flatten = chain.from_iterable
 
-    def apply_diffs_for_pts(self, cursor_pts, reset):
-        in_cached_mode = self.view.settings().get("git_savvy.diff_view.in_cached_mode")
+        patches = unique(flatten(filter_(self.head_and_hunk_for_pt(pt) for pt in cursor_pts)))
```
asfaltboy: Correct me if I'm wrong but I believe we can simply use a set (or set comprehension), as `head_and_hunk_for_pt` returns either None or substr (which is [guaranteed to be a string][1]). What do you say?

[1]: https://www.sublimetext.com/docs/3/api_reference.html#sublime.View

kaste: Hm, which parts of this line sould be replaceable? btw: `head_and_hunk_for_pt -> Optional[Tuple[str, str]]`

So we go:

```
			List[Optional[Tuple[str, str]]] 
- filter -> List[Tuple[str, str]]
- flat   -> List[str]
```

asfaltboy: Yea tuples of strings are usable in set just fine. I'm suggesting:
```
patches = ''.join(set(flatten(filter_(self.head_and_hunk_for_pt(pt) for pt in cursor_pts))))
```

But now that I think of it, it might not be good to jumble up the order of the strings; at the very least the head and hunk should come together ? I think we can scrape my suggestion in that case.

kaste: Yes for the outer unique, we must keep everything ordered.

kaste: Basically

```
HEAD_A, HUNK_B, HEAD_A, HUNK_C, ...
->
HEAD_A, HUNK_B, HUNK_C
```

kaste:
If you will I have the ©️ on the stubs. You start with `stubgen` (CLI tool that ships with mypy, just try it!), but then basically every Type everywhere is just `Any` 😒 	but you don't have to write all the method names at least unless the tool fails for some class...

kaste:
I *should* put them on GitHub and link them to here and SublimeLinter etc of course. Not sure how to do that easily.

asfaltboy:
@kaste this only applies to standard diff, not inline-diff right? I'm not certain, but it feels to me like we should be allowing multi-line / multi-hunk staging in inline diff as well, maybe in a separate PR

kaste:
Hooray

PR #1119, opened 2019-02-26
Mostly accurate jumping from diff to file

kaste:
Jumping from diff to file really was super sloppy implemented. It almost never jumped to the correct position. 

Implements mostly accurate jumping from diff to file by actually
parsing the hunk and recounting its lines.

For lines which are present on the b side (the view we jump to) this
is accurate of course. For deleted lines ('-') we, mildly arbitrary,
have to choose a good enough line.

If the user is not within a hunk, jump to the first visible line with content.

### DIFF
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -567,10 +567,6 @@ class GsDiffStageOrResetHunkCommand(TextCommand, GitCommand):
         self.view.run_command("gs_diff_refresh")
 
 
-HUNKS_LINES_RE = re.compile(r'@@ -(\d+)(?:,\d+)? \+(\d+)(?:,\d+)? ')
-HEADER_TO_FILE_RE = re.compile(r'\+\+\+ b/(.+)$')
-
-
 class GsDiffOpenFileAtHunkCommand(TextCommand, GitCommand):
 
     """
@@ -703,6 +699,9 @@ def real_rowcol_in_hunk(hunk, relative_rowcol):
         return line.b, 1
 
 
+HUNKS_LINES_RE = re.compile(r'@@ -(\d+)(?:,\d+)? \+(\d+)(?:,\d+)? ')
+
+
 def split_hunk(hunk):
     # type: (str) -> Optional[List[HunkLine]]
     """Split a hunk into (first char, line content, row) tuples
@@ -735,6 +734,9 @@ def _recount_lines(lines, b):
 def line_indentation(line):
     # type: (str) -> int
     return len(line) - len(line.lstrip())
+
+
+HEADER_TO_FILE_RE = re.compile(r'\+\+\+ b/(.+)$')
 
 
 def extract_filename_from_header(header):
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -668,6 +668,7 @@ def real_rowcol_in_hunk(hunk, relative_rowcol):
             for index, (first_char, _, _) in enumerate(hunk_lines, 1)
             if first_char in ('+', '-')
         )
+        col = 1
 
     first_char, line, b = hunk_lines[row_in_hunk - 1]
 
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -8,7 +8,6 @@ from functools import partial
 from itertools import chain, dropwhile, takewhile
 import os
 import re
-import bisect
 
 import sublime
 from sublime_plugin import WindowCommand, TextCommand, EventListener
@@ -20,7 +19,7 @@ from ...common import util
 
 
 if False:
-    from typing import Callable, Iterable, Iterator, List, Optional, Tuple, TypeVar
+    from typing import Callable, Iterable, Iterator, List, Optional, Set, Tuple, TypeVar
     from mypy_extensions import TypedDict
 
     T = TypeVar('T')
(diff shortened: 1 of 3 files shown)


core/commands/diff.py
```diff
@@ -18,6 +18,17 @@
 from ...common import util
 
 
+if False:
```
asfaltboy: I might be wrong, but shouldn't this be `MYPY = False; if MYPY:` ? Otherwise, would it still be used correctly by mypy?

Ref: https://github.com/python/mypy/issues/3216#issuecomment-296318286

core/commands/diff.py
```diff
@@ -468,54 +484,198 @@ class GsDiffOpenFileAtHunkCommand(TextCommand, GitCommand):
...
+        if not rowcol:
+            return None
+
+        row, col = rowcol
+
+        filename = extract_filename_from_header(header)
+        if not filename:
+            return None
+
+        return filename, row, max(col, 1)
```
asfaltboy: another suggestion, how about we return a named tuple instead of regular tuple here? This would make it easier to refer to each jump position item's arguments without indexing / unpacking, and can still be passed as "star args" if want?

core/commands/diff.py
kaste: I've seen both and both do work with mypy. If Guido wants your version I can change it but both do look idiosyncratic.

asfaltboy: Good to know the `if False` works too, agree it's super weird

kaste:
Overall, the code isn't very clear esp. the `col` thing. I think the problem here is when exactly col (and row) are zero based or 1-based. Usually iirc `rowcol` is zero based but `window.open_file` expects 1-based rowcols. git's diff is 1 based for the line (and doesn't provide a col). No I wonder why I often return `b, 0` etc. when I do a `max(col, 1)` later anyway. 🤔 

Now, I hope the tests are good. 'Mostly accurate' is iirc only when there are only deleted lines ('-') in a hunk it is a bit arbitrary where to jump to. Should we jump to the next or the previous line? Currently I always jump to col 0 in that case as well. Maybe that should change.

kaste:
Also, 'mostly accurate' are the symmetric '+-' blocks like

```
- foo
- foox
+ bar
+ barz
```

If you're on the 'foox' line it still jumps to the first '+' line, here 'bar'. Ideally maybe it would jump to the 'barz' line. But then it is not enough to just recount the lines, I also must group them to find such '+-' blocks.

kaste:
I actually rewrote this already without mention it here. 

Two notes: 

1. I extracted `parse_diff_in_view` and `head_and_hunk_for_pt` almost verbatim from `GsDiffStageOrResetHunkCommand`. These are also extracted and used in #1120. In a follow up PR I would refactor `GsDiffStageOrResetHunkCommand` to use these two functions instead. 

2. I don't want to introduce the namedtuple here. The data structure is private and only used in the once-per-filename filter. I actually can imagine to support selections here as well, t.i. if you select a word in the diff it will select the same word in the original file, which would make the namedtuple moot immediately.

kaste:
I learned in the meantime that 'show_commit' uses this functionality here as well, so there is a small fixup for that.

It was (maybe) a bit confusing to jump to the first changed line if the cursor is on the hunk header, so we jump to the first visible line instead.

kaste:
This is basically a fix for a currently broken feature. PTAL!

core/commands/diff.py
```diff
@@ -573,54 +576,167 @@ class GsDiffOpenFileAtHunkCommand(TextCommand, GitCommand):
...
+            return None
+
+        return filename, row, col
+
+
+def relative_rowcol_in_hunk(view, hunk_start, pt):
+    # type: (sublime.View, int, int) -> Tuple[int, int]
+    """Return rowcol of given pt relative to hunk start"""
+    head_row, _ = view.rowcol(hunk_start)
+    row_on_pt, col = view.rowcol(pt)
```
stoivo: rename to `pt_row`

core/commands/diff.py
```diff
@@ -573,54 +577,178 @@ class GsDiffOpenFileAtHunkCommand(TextCommand, GitCommand):
...
+
+    Note that rows point to available rows on the b-side.
+    """
+
+    head, *tail = hunk.rstrip().split('\n')
+    match = HUNKS_LINES_RE.search(head)
+    if not match:
+        return None
+
+    b = int(match.group(2))
```
stoivo: b, we can use a better name.

stoivo: Not neccessary, just a litt confusing when you start

stoivo:
You can do the fixup now. When through once tonight and will do once more tomorrow

kaste:
Ok, done another rebasing.

core/commands/diff.py
```diff
@@ -571,54 +575,178 @@ class GsDiffOpenFileAtHunkCommand(TextCommand, GitCommand):
...
+    # first visible line with some content instead.
+    if row_in_hunk == 0:
+        row_in_hunk = next(
+            (
+                index
+                for index, line in enumerate(hunk_lines, 1)
+                if line.mode in ('+', ' ') and line.text.strip()
+            ),
+            1
+        )
```
stoivo: As I understand these lines of code. Get me the index of the first line with  + or - and the like text in not blank. Why the not blank check?

stoivo:
a type in this commit message https://github.com/divmain/GitSavvy/pull/1119/commits/273ca5ff44670b0c6577834829c1e8fe5b547257

`Select fist line with content if on a hunk header line ` => `Select first line with content if on a hunk header line`

core/commands/diff.py
kaste: This is basically an arbitrary choice. As I tested this it just felt better if the cursor jumped to a position with code on it. Iirc I added this as a separate commit so it was literally fine tuning the algorithm.

These edge case decisions are why I say mostly accurate jumping. Most of the time it is accurate jumping t.i. fixing the previous buggy behavior. And then I added two arbitrary decisions which I think feel a tiny bit better.

kaste:
Fixed the typo, that was actually the commit where I added the 'not on a blank line' variant.

kaste:
🙏

PR #1120, opened 2019-02-27
Implement dynamic context lines (zooming)

### LINKED ISSUE #1065: Diff: Allow adding split hunks
Right now, the `git: diff` action allows adding whole hunks in the vain of `git add --patch`.

`git add --patch` allows splitting large hunks by pressing `s` when at the `Stage this hunk` prompt:

```
Stage this hunk [y,n,q,a,d,K,g,/,s,e,?]? ?
y - stage this hunk
n - do not stage this hunk
q - quit; do not stage this hunk or any of the remaining ones
a - stage this hunk and all later hunks in the file
d - do not stage this hunk or any of the later hunks in the file
g - select a hunk to go to
/ - search for a hunk matching the given regex
K - leave this hunk undecided, see previous hunk
s - split the current hunk into smaller hunks
e - manually edit the current hunk
? - print help
```

`s - split the current hunk into smaller hunks`.

This is possible when there are some unchanged lines between the `+`/-`-prefixed lines in a hunk. Then you can stage the two things independently.

It would be great if GitSavvy could support adding small hunks.

One may even argue that adding small hunks should be the default for an interactive tool like `git: diff`. It is simply not the default in the `git add --patch` terminal UI to make it possible to jump more quickly over things you don't want to stage. But in GitSavvy's `git: diff` view, this tradeoff isn't necessary, because we can just scroll down.

kaste:
Fixes #1065

Add `[+]`/`[-]` shortcuts to increase or decrease the contextual lines
`git diff` will output.

Uses the already implemented fuzzy hunk search to place the cursor correctly
after refreshing the view.

Note: We set `"--unidiff-zero"` when hunking if we do not show any contextual
lines which is otherwise discouraged bc it is mildly unsafe.

Added type annotations and some comments to old functions.

### DIFF
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -4,6 +4,7 @@ current diff.
 """
 
 from contextlib import contextmanager
+from functools import partial
 from itertools import dropwhile, takewhile
 import os
 import re
@@ -16,6 +17,17 @@ from .navigate import GsNavigate
 from ..git_command import GitCommand
 from ..exceptions import GitSavvyError
 from ...common import util
+
+
+if False:
+    from typing import Callable, Iterator, List, Optional, Tuple, TypeVar
+    from mypy_extensions import TypedDict
+
+    T = TypeVar('T')
+    ParsedDiff = TypedDict('ParsedDiff', {
+        'headers': List[Tuple[int, int]],
+        'hunks': List[Tuple[int, int]]
+    })
 
 
 DIFF_TITLE = "DIFF: {}"
@@ -63,6 +75,7 @@ class GsDiffCommand(WindowCommand, GitCommand):
             settings.set("git_savvy.diff_view.in_cached_mode", in_cached_mode)
             settings.set("git_savvy.diff_view.ignore_whitespace", False)
             settings.set("git_savvy.diff_view.show_word_diff", False)
+            settings.set("git_savvy.diff_view.context_lines", 3)
             settings.set("git_savvy.diff_view.base_commit", base_commit)
             settings.set("git_savvy.diff_view.target_commit", target_commit)
             settings.set("git_savvy.diff_view.show_diffstat", self.savvy_settings.get("show_diffstat", True))
@@ -132,6 +145,7 @@ class GsDiffRefreshCommand(TextCommand, GitCommand):
         target_commit = self.view.settings().get("git_savvy.diff_view.target_commit")
         show_diffstat = self.view.settings().get("git_savvy.diff_view.show_diffstat")
         disable_stage = self.view.settings().get("git_savvy.diff_view.disable_stage")
+        context_lines = self.view.settings().get('git_savvy.diff_view.context_lines')
 
         prelude = "\n"
         if self.file_path:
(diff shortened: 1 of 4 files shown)


core/commands/diff.py
```diff
@@ -340,6 +371,76 @@ def no_animations():
...
+        # type: (sublime.Edit, int) -> None
+        settings = self.view.settings()
+        current = settings.get('git_savvy.diff_view.context_lines')
+        next = max(current + amount, 0)
+        settings.set('git_savvy.diff_view.context_lines', next)
+
+        diff = parse_diff_in_view(self.view)
+        cur_hunks = [
+            self.view.substr(sublime.Region(*head_hunk[1]))
+            for head_hunk in filter_(head_and_hunk_for_pt(diff, s.a) for s in self.view.sel())
```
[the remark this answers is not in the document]
kaste: That's totally correct. To get easier merge conflicts I just copy-pasted this function from the other PR's.

common/commands/view_manipulation.py
```diff
@@ -54,6 +55,20 @@ def run(self, edit, text, nuke_cursors=False, restore_cursors=False):
...
 
+        # We cannot measure the `viewport_position` until right after this
+        # command actually finished. So we defer to the next tick/micro-task
+        # using `set_timeout`.
+        sublime.set_timeout(
+            lambda: self.view.run_command('gs_pin_x_axis', {'x': x})
+        )
+
+
+class gs_pin_x_axis(TextCommand):
```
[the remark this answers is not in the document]
kaste: Removed from this PR

core/commands/diff.py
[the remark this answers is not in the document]
kaste: Change in the other way around so that I now extract header and hunk for the fuzzy search.

kaste:
>  .. it seems to keep the hunks in place while simply changing the hunk line numbers/counts...

Maybe that's what the feature does. It just increases or decreases the contextual lines. E.g. if you press '-' you get more but shorter hunks.

kaste:
(GitHub has way more aggressive expanding, they expand by 20 lines if you click in their diff views.)

core/commands/diff.py
```diff
@@ -18,6 +19,17 @@
 from ...common import util
 
 
+if False:
```
stoivo: What's this for?

kaste: That's for 'mypy' typing. mypy will read and import the stuff in the 'False' branch but Sublime's python will not execute this code. 

Another way to write this is 

```
MYPY = False
if MYPY:
    ...from typing ...
```

which is probably just as confusing. @asfaltboy already asked the same question. :grin: but we first have to decide if these type annotations are welcome here at all.

stoivo:
Can you rebase on dev and clean the chage default commit into original.

kaste:
Done.

tests/test_diff_view.py
```diff
@@ -432,6 +432,46 @@ def setUp(self):
     def tearDown(self):
         unstub()
 
+    @p.expand([
```
stoivo: I'm not to fimiliar with writing these teste what does this do?

kaste: `@p.expand` = paramterize the tests

`expand` takes a list of tuples/args, and will create a separate test for each args/invocation signature.

PR #1137, opened 2019-07-05
Add `match_current_position` flag to `GsInlineDiffCommand`

kaste:
If set to true it will jump to the correct position you were editing in the
inline diff.

E.g. a keybinding

```
  {
      "keys": ["ctrl+shift+["],
      "command": "gs_inline_diff",
      "args": { "match_current_position": true }
  },
```

Workflow: edit buffer/file, save, issue command, inline diff will show up and jump to the correct position you were on while editing, ctrl+tab or ctrl+w to continue editing on the file, and so on.

### DIFF
--- a/core/commands/inline_diff.py
+++ b/core/commands/inline_diff.py
@@ -26,6 +26,41 @@ inline_diff_views = {}
 diff_view_hunks = {}
 
 
+def capture_cur_position(view):
+    try:
+        sel = view.sel()[0]
+    except Exception:
+        return None
+
+    return view.rowcol(sel.begin())
+
+
+def place_cursor_and_show(view, row, col):
+    view.sel().clear()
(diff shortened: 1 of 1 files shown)


core/commands/inline_diff.py
```diff
@@ -26,6 +26,42 @@
...
+
+    return view.rowcol(sel.begin())
+
+
+def place_cursor_and_show(view, row, col):
+    view.sel().clear()
+    pt = view.text_point(row, col)
+    view.sel().add(sublime.Region(pt, pt))
+    view.show_at_center(pt)
+    view.show_at_center(pt)
```
stoivo: twice?

kaste: Good catch, removed that.

PR #1138, opened 2019-07-05
Stoivo/inline diff big files

kaste:
@stoivo I cannot add commits to PR's. So this one works for me.

### DIFF
--- a/core/commands/inline_diff.py
+++ b/core/commands/inline_diff.py
@@ -178,6 +178,9 @@ class GsInlineDiffRefreshCommand(TextCommand, GitCommand):
     """
 
     def run(self, edit, match_position=None):
+        sublime.set_timeout_async(lambda: self.run_async(match_position=match_position), 0)
+
+    def run_async(self, match_position=None):
         file_path = self.file_path
         in_cached_mode = self.view.settings().get("git_savvy.inline_diff_view.in_cached_mode")
         ignore_eol_arg = (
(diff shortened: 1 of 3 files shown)


core/commands/inline_diff.py
```diff
@@ -343,8 +340,7 @@ class GsInlineDiffFocusEventListener(EventListener):
     latest file status when the view regains focus.
     """
 
-    def on_activated(self, view):
-
+    def on_activated_async(self, view):
         if view.settings().get("git_savvy.inline_diff_view") is True:
             view.run_command("gs_inline_diff_refresh")
```
kaste: That's a mistake. If you run a command like `view.run_command("gs_inline_diff_refresh")` it always starts at the main thread.

Look at how it is done for the diff view 

https://github.com/divmain/GitSavvy/blob/29642a3f5731be4cbd63f9acd56b4b714fbc0001/core/commands/diff.py#L129-L138

`sync` being the default you call it with `sync=False` in on_activated and in the main gs_inline_diff command.

stoivo: Do you know if there is a reliable way to add sleep to be blocking so I can check it works like I expect

kaste: Don't know what you're trying to do. `time.sleep` is always blocking. It is also important to check on which thread you run. (E.g. `import threading; print threading.current_thread()`)

stoivo: Take a look now

kaste:
LGTM! But I had to fix the commit message bc you got it backwards.

stoivo:
Good, it was late last night. thanks

PR #1139, opened 2019-07-05
Sometimes navigation scrolls to the right

stoivo:


core/commands/navigate.py
```diff
@@ -32,6 +32,11 @@ def run(self, edit, forward=True):
         sel.add(sublime.Region(new_position, new_position))
         self.view.run_command("show_at_center")
 
+        # The following shouldn't strictly be necessary, but Sublime sometimes
+        # jumps to the right when show_at_center for a column-zero-point occurs.
+        _, vp_y = view.viewport_position()
+        view.set_viewport_position((0, vp_y), False)
```
kaste: `view` is undefined. (Maybe we should start using flake8?)

core/commands/navigate.py
```diff
@@ -32,6 +32,11 @@ def run(self, edit, forward=True):
         sel.add(sublime.Region(new_position, new_position))
         self.view.run_command("show_at_center")
```
kaste: The modern version of this is `self.view.show_at_center(new_position)`

kaste:
That's probably a good idea! 

The commit message is wrong, it should be more like "Pin viewport's x-axis to 0 after `navigate`".

core/commands/navigate.py
kaste: Following a rule for legacy code: Whenever you open a file to edit, leave it in a better state.

stoivo: Thats true, thanks

kaste:
👍 :shipit:

kaste:
The same 'shift-to-right' can happen if we replace the buffers content, so maybe we should add that to our replace-content command as well.

stoivo:
I have never experienced that, but if you say so. Would you make that PR.

I am for adding flake8. We have used pycodestyle for a while. As I quickly rescearch I fould that we can run both and that should confilict.
https://github.com/divmain/GitSavvy/pull/1140

stoivo:
A small correction. We did have flake8 setup in setup.cfg but we should run it on CI and  I should setup the linter in sublime.

kaste:
The whole project has a dozens flake8 errors right now that's why I never turned them on on CI 😁

PR #1140, opened 2019-07-06
Run flake8 on CI

stoivo:


asfaltboy:
❤️it, great idea! One thing to note is that flake8 basically runs pycodestyle  (+pyflakes + mccabe) for us, so there should be no need to install or run pycodestyle separately.,

stoivo:
What's you souce on that?

.travis.yml
```diff
@@ -30,12 +30,12 @@ before_install:
...
     - sh travis.sh bootstrap
     - sh travis.sh install_package_control
-    - pip3 install python-coveralls pycodestyle;
+    - pip3 install python-coveralls flake8;
 
 script:
     - sh travis.sh run_tests --coverage
     - sh travis.sh run_syntax_tests
-    - pycodestyle --statistics --count
+    - flake8 cmdbox
```
kaste: What should `cmdbox` be? If I checkout this PR and run `flake8 .` I still get errors.

stoivo: Not much considerration, just copy pasted the first I have over https://simpleisbetterthancomplex.com/packages/2016/08/05/flake8.html

falake8 . gives me error too. I will ammend the commit to run flake8 . . Then fix those errors too

stoivo:
@kaste can you tel me why `SublimeLinter-flake8` can't read the setup.py?

kaste:
You mean setup.cfg? SublimeLinter works with that, actually SublimeLinter itself uses this configuration file format.

stoivo:
Do I need to add it to some setting because it is'nt working over here.  I get errors line too long which we set the max length to 120

kaste:
Should work with the defaults, maybe you modified your global settings?

asfaltboy:
I use SublimeLinter-flake8 and use a simple `.flake8` file, maybe we can add that to be explicit?

```ini
[flake8]
count = False
ignore = E401, W503, W504
max-line-length = 120
```

kaste:
That wouldn't necessarily help @stoivo because either it reads the config files or not. (Or the config is overwritten by other settings.) 🤷‍♀

stoivo:
SublimeLinter.sublime-settings
```
{
	"linters": {
	    "eslint": {
	      "selector": "text.html.vue, source.ts, source.js - meta.attribute-with-value"
	      // "selector": ", source.js"
	    }
	  }
}
```

stoivo:
Is `yield from` just a helper to do 
```py
```

stoivo:
Almost done
```
./core/interfaces/tags.py:6:1: F403 'from ..commands import *' used; unable to detect undefined names
./core/interfaces/tags.py:355:32: F405 'GsNavigate' may be undefined, or defined from star imports: ..commands
./core/interfaces/__init__.py:10:38: F405 'GitCommand' may be undefined, or defined from star imports: .branch, .rebase, .status, .tags
./core/interfaces/status.py:8:1: F403 'from ..commands import *' used; unable to detect undefined names
./core/interfaces/status.py:901:35: F405 'GsNavigate' may be undefined, or defined from star imports: ..commands
```

We have two options:
1. Ignore, F403 and F405 for core/interfaces
2. inclide all  command we need explicitly  

I prefer the 2.

asfaltboy:
> We have two options:
> 
> 1. Ignore, F403 and F405 for core/interfaces
> 2. inclide all  command we need explicitly
> 
> I prefer the 2.

If there are many imports needed (e.g more than 5), a 3rd option could be to specify the explicit `__all__ = ["function", "object", ...]` var in the imported module. This should also turn off the error if I recall correctly.

Reference: https://docs.python.org/3.3/tutorial/modules.html#importing-from-a-package

asfaltboy:
> Is `yield from` just a helper to do
> 
> ```python
> def g(x):
>     yield from range(x, 10)
>     yield from range(9, 10)
> 
> 
> def f(x):
>     for i in range(x, 10):
>         yield i
>     for i in range(9, 10):
>         yield i
> 
> 
> def test(x):
>     print(list(g(x)))
>     print(list(f(x)))
> ```
> 
> **EDIT:** is it used as a wait?
> 

It's not only that but also allows for a "bi-directional channel" to communicate data and exception to/from a generator. In our case, `setup_graph_view_async` will yield all values in multiple functions deep, until it is done, finally returning the view. It is a somewhat confusing but elegant interface to composed function calls.

I think it's definitely less straightforward than just calling function, but I somehow hope that AsyncIO will make its way into sublime (python 3.6 is next?) and we'll be able to just "await" all IO calls, using `yield from` will then make more sense. @randy3k should be able to confirm or deny our assumptions.

Ref: https://stackoverflow.com/a/26109157/484127

Usage in AsyncIO is sort of similar to Go's channels, [also seen in other languages][1]...

[1]: https://www.reddit.com/r/golang/comments/97v3r1/other_programming_languages_with_channels/

kaste:
@stoivo Go for 2, explicit imports in those cases. Otherwise we must use `from x import *` quiet often in this code base, e.g. if the module exports commands or event listeners. Exporting them manually would be error prone and tedious.

stoivo:
@kaste, since you work with Python on a profesonall basis, you'll probably be better suited to understand and fix this. 
Un travis we are running python 3.3 on linux since it is the version sublime runns. The Mac build runns python 3.6.

# on travis mac 
flake8 . --version
3.7.8 (mccabe: 0.6.1, pycodestyle: 2.5.0, pyflakes: 2.1.1) CPython 3.6.5 on version

# on travis linux
flake8 . --version
3.7.8 (mccabe: 0.6.1, pycodestyle: 2.5.0, pyflakes: 2.1.1) CPython 3.3.6 on Linux


Flake8 droped support for python3.3 in 3.6.0 https://flake8.pycqa.org/en/latest/release-notes/3.6.0.html . So we should run flake 3.5.0 but that one seems to ignoe our validation rules.

.travis.yml
```diff
@@ -13,11 +13,12 @@ matrix:
         - name: Linux
           os: linux
           language: python
-          python: 3.3
+          python: 3.6
```
kaste: 👍

.travis.yml
```diff
@@ -30,12 +31,13 @@ before_install:
 install:
     - sh travis.sh bootstrap
     - sh travis.sh install_package_control
-    - pip3 install python-coveralls pycodestyle;
+    - pip3 install python-coveralls flake8==3.5.0;
```
kaste: Just take the latest flake8

.travis.yml
```diff
@@ -30,12 +30,11 @@ before_install:
 install:
     - sh travis.sh bootstrap
     - sh travis.sh install_package_control
-    - pip3 install python-coveralls pycodestyle;
+    - pip3 install python-coveralls flake8
 
 script:
     - sh travis.sh run_tests --coverage
     - sh travis.sh run_syntax_tests
-    - pycodestyle --statistics --count
```
kaste: You don't run flake8. 😕

stoivo: Hmm, odd. Dont know why it is removed. Doesnt matter too much, no need to lint on both appveyor and travis. But the better fix it github action https://github.com/divmain/GitSavvy/pull/1170
I do some more work tring to reach him

asfaltboy:
> What's you souce on that?

sorry missed your query, its here: http://flake8.pycqa.org/en/latest/faq.html#why-does-flake8-use-ranges-for-its-dependencies

.travis.yml
asfaltboy: @stoivo you mean to email @divmain to try to enable github actions?

stoivo: I have emailed him and emailed some other of the groups his in to see if they can ask him to concact us. 

I wrote question to you on https://gitter.im/divmain/GitSavvy

stoivo:
@asfaltboy, thanks

PR #1142, opened 2019-07-09
Fix reloader code

kaste:
The old reloader did actually not reload the window and text commands which is
very annoying. We could fix it ourself, but there is no real minimal fix for it,
or just use the code from @randy3k's `AutomaticPackageReloader`.

The following is a copy of
https://github.com/randy3k/AutomaticPackageReloader/blob/b568fcdf4e208d403f0fe5ee967016d589efa765/reloader/reloader.py
under MIT but @randy3k is also a member of GitSavvy.

Added the entry point function `reload_plugin`.

Discussion:

I tried a minimal fix but essentially this is the right code. We could remove some superfluous code like reloading dependencies as we don't have them but then we loose the ability to diff easily against a future version of this file upstream.

Since we do not use the reloader on upgrade (actually why?) we could also just drop our reloader and recommend and use the `AutomaticPackageReloader` package. 🤷‍♀

### DIFF
--- a/common/util/reload.py
+++ b/common/util/reload.py
@@ -1,122 +1,233 @@
+import sublime
+import sublime_plugin
+import os
+import posixpath
+import threading
 import builtins
 import functools
 import importlib
 import sys
-import types
+from inspect import ismodule
 from contextlib import contextmanager
-
-import sublime_plugin
-
-from .debug import StackMeter, trace
+from .debug import StackMeter
 
 
-dprint = trace.for_tag("reload")
+try:
+    from package_control.package_manager import PackageManager
+
+    def is_dependency(pkg_name):
+        return PackageManager()._is_dependency(pkg_name)
+
+except ImportError:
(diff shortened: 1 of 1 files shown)


asfaltboy:
I'm ok with merging this change, as it seems to works great for me 🥇 .

And then, we can discuss further improvements with Randy for a separate PR(s).

For instance, we could contribute [a `package_control.events.post_upgrade` hook][3] for triggering auto-reload on package upgrade to @randy3k 's [AutomaticPackageReloader][1], and then add it as a [dependency][2] for GitSavvy? 🙂 


[1]: https://github.com/randy3k/AutomaticPackageReloader
[2]: https://packagecontrol.io/docs/dependencies
[3]: https://packagecontrol.io/docs/events#API

randy3k:
Thank all guys. I am in the middle of a cross country moving and extremely busy these days. Great job!

PR #1145, opened 2019-07-12
Fix: Add `file_path` for graph current file

kaste:
The command graph current file is currently broken bc the actual file_path is not added to the command. So this is a simple fix, but an embarrassing bug.


stoivo:
Thanks. I will merge this, a regressons since 6a6c612a0ceb92d8e7ef90ae990f1aff2da725d6. 

Just so you know. I appriciate all the PRs but I dont have that much time around. I will try to merge one by one the comming weeks.

PR #1146, opened 2019-07-12
Remember last log and graph pre-select action

kaste:
Remember the last main(?) action, e.g. 'log all branches' etc.

We do this independently for log and graph_log maybe bc I like graph for all branches, but log for the current branch. It would be probably even better if we remembered the default actions for the 'current_file' variants as well bc graph current file is usually best with current branch. I could add that.


kaste:
This one works already in dev. I obviously broke that while doing the other things. 🤷‍♀

PR #1149, opened 2019-07-12
Show refs in the log panel

kaste:
We all know the log panel. This PR adds ref names (branch names, tags) to the list. There is a bit of fancy decoration/formatting around these refs. This makes it easier to search for these names.

image

E.g. search a tag 

image

### DIFF
--- a/core/git_mixins/history.py
+++ b/core/git_mixins/history.py
@@ -6,6 +6,7 @@ import sublime
 LogEntry = namedtuple("LogEntry", (
     "short_hash",
     "long_hash",
+    "ref",
     "summary",
     "raw_body",
     "author",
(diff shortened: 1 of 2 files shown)


core/git_mixins/history.py
```diff
@@ -62,8 +63,8 @@ def log(self, author=None, branch=None, file_path=None, start_end=None, cherry=N
                 continue
             entry, raw_body = entry.split("\x00")
 
-            short_hash, long_hash, summary, author, email, datetime = entry.split("\n")
-            entries.append(LogEntry(short_hash, long_hash, summary, raw_body, author, email, datetime))
+            short_hash, long_hash, ref, summary, author, email, datetime = entry.split("\n")
+            entries.append(LogEntry(short_hash, long_hash, ref.strip(), summary, raw_body, author, email, datetime))
```
stoivo: I dont think you need strip

kaste:
Thanks for merging. A note here: the default commands set the limit to 6000 which is usually everything up to the first commit. With that *searching* doesn't really work reliable because there are just too much entries to match the search term against. (Decorating is still nice imo because e.g. if you log the current branch you maybe see when you hit the master.)

But you can of course have your own bindings and usages of the log command. The fewer items you display the better your search experience for natural search terms.

On the other side, if you search for a specific commit the 6000 limit works okay bc sha's are unique.

stoivo:
I can't recall why the limit is set to 6000. I have 3 repos with more then 6000 commits. When I open the log panel I either go from the top and hit the arrow down or search with the sha.

PR #1150, opened 2019-07-12
Remember last branch used by log filter by branch

kaste:
Log as well as graph log can "filter by branch". With this PR the last used branch will be remembered.

### DIFF
--- a/core/commands/log.py
+++ b/core/commands/log.py
@@ -118,12 +118,17 @@ class GsLogByAuthorCommand(LogMixin, WindowCommand, GitCommand):
 
 
 class GsLogByBranchCommand(LogMixin, WindowCommand, GitCommand):
+    _selected_branch = None
 
     def run_async(self, **kwargs):
-        show_branch_panel(lambda branch: self.on_branch_selection(branch, **kwargs))
+        show_branch_panel(
+            lambda branch: self.on_branch_selection(branch, **kwargs),
+            selected_branch=self._selected_branch
+        )
 
     def on_branch_selection(self, branch, **kwargs):
         if branch:
+            self._selected_branch = branch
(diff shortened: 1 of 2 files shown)


stoivo:
When you you find youself in the situasion where you want to do this? Git log by branch and selecte the same branch twice. I think selection the current branch is better.

kaste:
Well, if you do just `git: log` a panel opens with `current branch/all/filter by branch`. So if I want the current branch I actually pick it from here. If I hit `filter by branch` I already know I want to look at a different branch. 

A user could also have shortcuts to its fav commands though. If she usually uses `GsLogByBranchCommand` it will initially *still* select the current branch (because on init the selected branch is `None` which translates to selecting the current branch). Just after that it will always preselect the last used branch. I think that's an useful improvement.

stoivo:
I don't see it as a good argument for adding the feature. It seams like featue which only maybe possible could be in use. I dont think I will ever use this, an if I do I think this would probably annoy me more than help me. If someone are using `GsLogByBranchCommand` directly they probly want current branch most of the time. 

Anyway, rejection of PR feel bad and If you feel like there are someone who would benefit from this we can merge it.

kaste:
Oh, you changed your mind. 😁

stoivo:
I will open a revert PR

randy3k:
Given that a number of panels memoize the last selected options, I guess it is consistent to prechoose the last branch here. Just my 2 cents

randy3k:
By the way, I think we should expose these commands.

```
    {
        "caption": "git: log current branch",
        "command": "gs_log_current_branch",
    },
    {
        "caption": "git: graph current branch",
        "command": "gs_log_graph_current_branch",
    },
```

PR #1152, opened 2019-07-14
Fix: yaml.load deprecation warning

deathaxe:
Hotfixes should be submitted to `master` branch. As part of the PR process, you will be asked to provide the information necessary to make that happen.

Pull requests on new features should be submitted to `dev` branch and will be regularly merged into `master` after evaluations over an extended period of time.

### DIFF
--- a/common/util/file.py
+++ b/common/util/file.py
@@ -26,7 +26,7 @@ def _determine_syntax_files():
         try:
             # Use `sublime.load_resource`, in case Package is `*.sublime-package`.
             resource = sublime.load_resource(syntax_file)
-            for extension in yaml.load(resource)["file_extensions"]:
+            for extension in yaml.safe_load(resource)["file_extensions"]:
                 if extension not in syntax_file_map:
                     syntax_file_map[extension] = []
                 extension_list = syntax_file_map[extension]


asfaltboy:
The change looks legit, going to merge

asfaltboy:
Just to recap what we're using the `yaml` package for. On GitSavvy startup, we pre-load all syntax definitions for various file types (not necessarily GitSavvy's), so that we can use those definitions when, for example, we load a historical version of a file or inline-diff.

Fortunately, we have a try/except around the loader, so invalid YAML issues are ignored (which will result in plain text rendering instead). But, this means that our current parser fails to parse syntax definitions for many formats (that are coming out of the box with sublime).

Note, it happens even when using the old "unsafe load" method, thus belongs in a separate issue. Perhaps even worth filing in <https://github.com/sublimehq/Packages>, just to get the list of YAML extensions (addons) used.

Here is a list of (92) syntax files that fail to parse for me, for future reference (obtained by adding a trace call in [the related try/except clause][1]):

<details>
<summary>Click to expand</summary>

```python
GS [debug] FAILED to load Packages/Batch File/Batch File.sublime-syntax could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<unicode string>", line 169, column 14:
        - match: =
                 ^
GS [debug] FAILED to load Packages/C#/C#.sublime-syntax could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<unicode string>", line 766, column 22:
                - match: =
                         ^
GS [debug] FAILED to load Packages/D/DMD Output.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/Erlang/Erlang.sublime-syntax could not determine a constructor for the tag 'tag:yaml.org,2002:merge'
  in "<unicode string>", line 50, column 14:
        - match: <<
                 ^
GS [debug] FAILED to load Packages/Git Formats/Git Attributes.sublime-syntax could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<unicode string>", line 150, column 14:
        - match: =
                 ^
GS [debug] FAILED to load Packages/Git Formats/Git Common.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/Go/Go.sublime-syntax while parsing a flow node
expected the node content, but found ':'
  in "<unicode string>", line 288, column 13:
        {match: :=   , scope: keyword.operator.a ... 
                ^
GS [debug] FAILED to load Packages/HTML/HTML.sublime-syntax could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<unicode string>", line 340, column 18:
            - match: =
                     ^
GS [debug] FAILED to load Packages/JavaScript/Regular Expressions (JavaScript).sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/LaTeX/LaTeX Log.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/Lua/Lua.sublime-syntax could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<unicode string>", line 291, column 14:
        - match: =
                 ^
GS [debug] FAILED to load Packages/Makefile/Make Output.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/Makefile/Makefile.sublime-syntax could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<unicode string>", line 622, column 14:
        - match: =
                 ^
GS [debug] FAILED to load Packages/Markdown/MultiMarkdown.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/OCaml/camlp4.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/PHP/PHP Source.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/PHP/Regular Expressions (PHP).sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/Rust/Cargo.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/ShellScript/Bash.sublime-syntax could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<unicode string>", line 494, column 26:
                    - match: =
                             ^
GS [debug] FAILED to load Packages/ShellScript/commands-builtin-shell-bash.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/ShellScript/Shell-Unix-Generic.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/EditorConfig/EditorConfig.sublime-syntax while scanning a directive
  in "<unicode string>", line 1, column 1:
    %YAML1.2
    ^
expected alphabetic or numeric character, but found '.'
  in "<unicode string>", line 1, column 7:
    %YAML1.2
          ^
GS [debug] FAILED to load Packages/LSP/Syntaxes/Diagnostics.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/LSP/Syntaxes/References.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/NeoVintageous/res/Command-line output.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/NeoVintageous/res/Help.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/PackageDev/Package/Regex Replacement (Escaped).sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/PackageDev/Package/Regex Replacement.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/PackageDev/Package/Regular Expression (Escaped).sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/PackageDev/Package/Scope Selector.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/PackageDev/Package/Sublime JSON/Sublime JSON.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/PackageDev/Package/Sublime Text Snippet/Sublime Text Snippet (Raw).sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/Requester/syntax/requester-benchmarks.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/Requester/syntax/requester-history.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/Requester/syntax/requester-response.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/Requester/syntax/requester-test.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/SublimeLinter/panel/message_view.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/SublimeLinter/panel/panel.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/Babel/JavaScript (Babel).sublime-syntax could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<unicode string>", line 1168, column 14:
        - match: =
                 ^
GS [debug] FAILED to load Packages/GitSavvy/syntax/blame.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/GitSavvy/syntax/branch.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/GitSavvy/syntax/dashboard.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/GitSavvy/syntax/diff.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/GitSavvy/syntax/diff_view.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/GitSavvy/syntax/make_commit.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/GitSavvy/syntax/rebase.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/GitSavvy/syntax/rebase_interactive.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/GitSavvy/syntax/show_commit.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/GitSavvy/syntax/show_commit_diffstat.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/GitSavvy/syntax/status.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/GitSavvy/syntax/tags.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/UnitTesting/res/unit-testing-test-result.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/UnitTesting/tests/_Syntax_Compat_Failure/TestSyntax.sublime-syntax 'file_extensions'
GS [debug] FAILED to load Packages/UnitTesting/tests/_Syntax_Compat_Success/TestSyntax.sublime-syntax 'file_extensions'
```
</details>

Note 2: many of the above fail because of the syntax file missing the `file_extensions` key.

[1]: https://github.com/divmain/GitSavvy/blob/31345e911757f4bb0196aae31dcc754cf95fcf43/common/util/file.py#L35

deathaxe:
Looksl like pyyaml not to follow the official specifications under all circumstances. All listed parse errors are valid yaml statements. Should rather be filed as issue at the pyyaml repo.

asfaltboy:
I disagree. First I doubt that pyyaml got the spec wrong, it's been around too long. If you try any of these oddities in any parser you'll see the same: https://yaml-online-parser.appspot.com/

The reason is that the "mapping" values are unquoted, so they cannot start with certain "indicator characters" such as colon or equals chars. The easiest solution would be to just quote them 😅

deathaxe:
> The easiest solution would be to just quote them

Agree, and not apposed to that idea.

But on the other hand I know the official YAML.sublime-syntax and [Package Dev's Sublime Text Syntax Definitions](https://github.com/SublimeText/PackageDev/tree/master/Package/Sublime%20Text%20Syntax%20Definition) following the [YAML 1.2 specs](http://yaml.org/spec/1.2/spec.html) very closely. 

I can remember some discussions about the topic of quoted vs. unquoted mapping values with exactly the result, stated in my previous post. 

Not sure if it is the correct "rule", but the specs at See: https://yaml.org/spec/1.2/spec.html#:%20mapping%20value// say:

>  Normally, YAML insists the “:” mapping value indicator be separated from the value by white space. A benefit of this restriction is that the “:” character can be used inside plain scalars, as long as it is not followed by white space. This allows for unquoted URLs and timestamps. It is also a potential source for confusion as “a:1” is a plain scalar and not a key: value pair.

> Note that the value may be completely empty since its existence is indicated by the “:”. 

This means "indicator characters" are meant to be used in plain unquoted scalars as the double colon followed by whitespace indecates the key-value-separation.

deathaxe:
>  I doubt that pyyaml got the spec wrong, it's been around too long.

That's very correct. pyyaml is a YAML1.1 parser while sublime-syntax is YAML1.2

While YAML1.2 is fully backward compatible there may be tokens, which are not compatible with 1.1 though. All parse errors you see are such situations.

asfaltboy:
Almost, except that it's not valid YAML 1.2 either; here's a patch using [ruamel][1] to do the same:

<details>
<summary>Path to GitSavvy</summary>

```diff
 common/util/file.py | 19 ++++++++++++++-----
 dependencies.json   |  2 +-
 2 files changed, 15 insertions(+), 6 deletions(-)

diff --git a/common/util/file.py b/common/util/file.py
index 0f0b441..7b58050 100644
--- a/common/util/file.py
+++ b/common/util/file.py
@@ -1,9 +1,13 @@
-import sublime
-import threading
-import yaml
 import os
+import threading
 from contextlib import contextmanager
 
+from ruamel.yaml import YAML
+
+import sublime
+
+from ..util import debug
+
 
 if 'syntax_file_map' not in globals():
     syntax_file_map = {}
@@ -22,16 +26,21 @@ def determine_syntax_files():
 
 def _determine_syntax_files():
     syntax_files = sublime.find_resources("*.sublime-syntax")
+    yaml = YAML(typ='safe')
     for syntax_file in syntax_files:
         try:
             # Use `sublime.load_resource`, in case Package is `*.sublime-package`.
             resource = sublime.load_resource(syntax_file)
-            for extension in yaml.safe_load(resource)["file_extensions"]:
+            parsed_resource = yaml.load(resource)
+            if "file_extensions" not in parsed_resource:
+                continue
+            for extension in parsed_resource["file_extensions"]:
                 if extension not in syntax_file_map:
                     syntax_file_map[extension] = []
                 extension_list = syntax_file_map[extension]
                 extension_list.append(syntax_file)
-        except Exception:
+        except Exception as e:
+            debug.trace("error parsing syntax file: ", e)
             continue
 
 
diff --git a/dependencies.json b/dependencies.json
index 45bd89b..da09b33 100644
--- a/dependencies.json
+++ b/dependencies.json
@@ -1,7 +1,7 @@
 {
     "*": {
         ">=3000": [
-            "pyyaml"
+            "ruamel-yaml"
         ]
     }
 }
```
</details>

And the results:
<details>
<summary>Result of trace</summary>

```
GS [debug] error parsing 'Packages/Batch File/Batch File.sublime-syntax' syntax file: could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<unicode string>", line 169, column 14:
        - match: =
                 ^ (line: 169)
GS [debug] error parsing 'Packages/C#/C#.sublime-syntax' syntax file: could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<unicode string>", line 766, column 22:
                - match: =
                         ^ (line: 766)
GS [debug] no 'file_extensions' key in 'Packages/D/DMD Output.sublime-syntax'
GS [debug] error parsing 'Packages/Erlang/Erlang.sublime-syntax' syntax file: could not determine a constructor for the tag 'tag:yaml.org,2002:merge'
  in "<unicode string>", line 50, column 14:
        - match: <<
                 ^ (line: 50)
GS [debug] error parsing 'Packages/Git Formats/Git Attributes.sublime-syntax' syntax file: could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<unicode string>", line 150, column 14:
        - match: =
                 ^ (line: 150)
GS [debug] no 'file_extensions' key in 'Packages/Git Formats/Git Common.sublime-syntax'
GS [debug] error parsing 'Packages/Go/Go.sublime-syntax' syntax file: while parsing a flow node
expected the node content, but found ':'
  in "<unicode string>", line 288, column 13:
        {match: :=   , scope: keyword.operator.a ... 
                ^ (line: 288)
GS [debug] error parsing 'Packages/HTML/HTML.sublime-syntax' syntax file: could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<unicode string>", line 340, column 18:
            - match: =
                     ^ (line: 340)
GS [debug] no 'file_extensions' key in 'Packages/JavaScript/Regular Expressions (JavaScript).sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/LaTeX/LaTeX Log.sublime-syntax'
GS [debug] error parsing 'Packages/Lua/Lua.sublime-syntax' syntax file: could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<unicode string>", line 291, column 14:
        - match: =
                 ^ (line: 291)
GS [debug] no 'file_extensions' key in 'Packages/Makefile/Make Output.sublime-syntax'
GS [debug] error parsing 'Packages/Makefile/Makefile.sublime-syntax' syntax file: could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<unicode string>", line 622, column 14:
        - match: =
                 ^ (line: 622)
GS [debug] no 'file_extensions' key in 'Packages/Markdown/MultiMarkdown.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/OCaml/camlp4.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/PHP/PHP Source.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/PHP/Regular Expressions (PHP).sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/Rust/Cargo.sublime-syntax'
GS [debug] error parsing 'Packages/ShellScript/Bash.sublime-syntax' syntax file: could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<unicode string>", line 494, column 26:
                    - match: =
                             ^ (line: 494)
GS [debug] no 'file_extensions' key in 'Packages/ShellScript/commands-builtin-shell-bash.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/ShellScript/Shell-Unix-Generic.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/ANSIescape/ANSI.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/LSP/Syntaxes/Diagnostics.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/LSP/Syntaxes/References.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/NeoVintageous/res/Command-line output.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/NeoVintageous/res/Help.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/PackageDev/Package/Regex Replacement (Escaped).sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/PackageDev/Package/Regex Replacement.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/PackageDev/Package/Regular Expression (Escaped).sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/PackageDev/Package/Scope Selector.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/PackageDev/Package/Sublime JSON/Sublime JSON.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/PackageDev/Package/Sublime Text Snippet/Sublime Text Snippet (Raw).sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/Requester/syntax/requester-benchmarks.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/Requester/syntax/requester-history.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/Requester/syntax/requester-response.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/Requester/syntax/requester-test.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/SublimeLinter/panel/message_view.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/SublimeLinter/panel/panel.sublime-syntax'
GS [debug] error parsing 'Packages/Babel/JavaScript (Babel).sublime-syntax' syntax file: could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<unicode string>", line 1168, column 14:
        - match: =
                 ^ (line: 1168)
GS [debug] error parsing 'Packages/GitSavvy/syntax/blame.sublime-syntax' syntax file: could not determine a constructor for the tag 'tag:yaml.org,2002:value'
  in "<unicode string>", line 147, column 18:
            - match: =
                     ^ (line: 147)
GS [debug] no 'file_extensions' key in 'Packages/GitSavvy/syntax/branch.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/GitSavvy/syntax/dashboard.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/GitSavvy/syntax/diff.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/GitSavvy/syntax/diff_view.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/GitSavvy/syntax/make_commit.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/GitSavvy/syntax/rebase.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/GitSavvy/syntax/rebase_interactive.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/GitSavvy/syntax/show_commit.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/GitSavvy/syntax/show_commit_diffstat.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/GitSavvy/syntax/status.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/GitSavvy/syntax/tags.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/UnitTesting/res/unit-testing-test-result.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/UnitTesting/tests/_Syntax_Compat_Failure/TestSyntax.sublime-syntax'
GS [debug] no 'file_extensions' key in 'Packages/UnitTesting/tests/_Syntax_Compat_Success/TestSyntax.sublime-syntax'
```
</details>

I don't get it. If sublime syntax defines literal character tokens (e.g "=", ":=", "<<", etc), why won't we just quote these?

[1]: https://yaml.readthedocs.io/en/latest/overview.html

Thom1729:
I figured out the problem.

In the first place, plain scalars like `=` and `<<` are perfectly valid. They're absolutely legal according to the spec, and ruamel.yaml is parsing them correctly.

The problem occurs after parsing, during the composition phase. Because those values don't have an explicit tag, they are assigned a non-specific tag. Plain scalars are assigned the non-specific tag `?`. (Non-plain scalars get `!`.) Then, ruamel.yaml resolves the non-specific tag to a specific tag based on the value of the scalar. `=` gets `tag:yaml.org,2002:value` and `<<` gets `tag:yaml.org,2002:merge`. This is (IMO) the bug. These are the [Language-Independent Types for YAML™ Version 1.1](https://yaml.org/type/). Instead, for YAML 1.2 documents ruamel.yaml should default to the [YAML 1.2 Core Schema](https://yaml.org/spec/1.2/2009-07-21/spec.html#id2604037) and tag both scalars as `tag:yaml.org,2002:str`. (Non-plain scalars are always resolved to `tag:yaml.org,2002:str`, which is why those work as expected.)

The error is raised during the construction phase because the constructor doesn't know what to do with nodes tagged `tag:yaml.org,2002:value and `tag:yaml.org,2002:merge`. It's not a parse error, but a construction error, and it's caused by the faulty tag resolution during composition.

I've submitted [an issue](https://bitbucket.org/ruamel/yaml/issues/308/resolver-uses-non-core-tags-in-yaml-12). The author of ruamel.yaml is very responsive, so hopefully we'll hear something before long.

There's an easy workaround. RoundTripLoader gracefully handles unknown tags, so if you replace `yaml.safe_load(resource)` with `yaml.round_trip_load(resource)`, everything will work fine. RoundTripLoader is also “safe”, so you don't have to worry about tags like `!!python/object:module.name`. The downside is that it doesn't use the optimized C parser, but I actually cut that from the `ruamel-yaml` dependency anyway because I'm lazy and don't want to deal with binaries.

A cleaner but less easy workaround would be to fix the non-specific tag resolution. This shouldn't be too much work; the resolver source code is fairly self-explanatory.

AvdN:
If there are any implicit tags you don't want to use you can easily remove them in ruamel.yaml:
```
import ruamel.yaml
ir = ruamel.yaml.resolver.implicit_resolvers
for idx in range(len(ir), 0, -1):
    typ = ir[idx-1][1].rsplit(':', 1)[1]
    if typ in ['value', 'merge']:
        del ir[idx-1]

yaml = ruamel.yaml.YAML(typ='safe')  # or leave out for 'rt'
data = yaml.load("""\
- <<: {a: 1, b: 3}
  b: 2
- =
""")

print(data)
```
the above prints ``[{'<<': {'a': 1, 'b': 3}, 'b': 2}, '=']`` (and fails without the for loop)

asfaltboy:
Bravo, and thank you both for your expertise! If nothing else, this conversation has enlightened us about the flexibility (complexity?) of YAML format and it's schemas. Very, very interesting, thanks a lot for taking the time to investigate and explain!

I still have one small issue I have with the assumption that the character is a valid scalar (but I shall comment this in the referenced ticket sublimehq/Packages#2053 instead of here).

asfaltboy:
With regard to the syntax parsing, even if we fix this are we sure this even works well? It seems (for me at least) that the syntax highlighting is always turned off in inline diff (or is there a setting I'm missing?)

kaste:
Even for python files? Always worked for me, but I'm not sure if this is actually an optional feature 🤷‍♀ 

image

PR #1157, opened 2019-08-09
Colorize the path to the next commit

kaste:
Nobody can follow these graphs anyway 😦 

image

Well, let's fix that :grin:

FfHhNWiTTY

🌶

### DIFF
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -92,6 +92,12 @@
             "remove_background": "#A83732",
             "remove_foreground_bold": "#F9F9F4",
             "remove_background_bold": "#702820"
+        },
+        "log_graph": {
+            "commit_dot_foreground": "#9911",
+            "commit_dot_background": "#991",
+            "path_foreground": "#991",
+            "path_background": "#99991109"
         }
     },
 
--- a/core/commands/log_graph.py
+++ b/core/commands/log_graph.py
@@ -17,7 +17,7 @@ from ...common.theme_generator import XMLThemeGenerator, JSONThemeGenerator
 
 MYPY = False
 if MYPY:
-    from typing import Iterator, Tuple
+    from typing import Iterator, Set, Tuple
 
 
 COMMIT_NODE_CHAR = "●"
@@ -313,8 +313,13 @@ class GsLogGraphCursorListener(EventListener, GitCommand):
 
 def colorize_dots(view):
     # type: (sublime.View) -> None
-    dots = tuple(_find_dots(view))
+    dots = tuple(find_dots(view))
     _colorize_dots(view.id(), dots)
+
+
+def find_dots(view):
+    # type: (sublime.View) -> Set[colorizer.Char]
+    return set(_find_dots(view))
 
 
 def _find_dots(view):
--- a/core/commands/log_graph.py
+++ b/core/commands/log_graph.py
@@ -314,7 +314,7 @@ class GsLogGraphCursorListener(EventListener, GitCommand):
 def colorize_dots(view):
     # type: (sublime.View) -> None
     dots = tuple(_find_dots(view))
-    _colorize_dots(view.id(), dots, view.change_count())
+    _colorize_dots(view.id(), dots)
 
 
 def _find_dots(view):
@@ -329,8 +329,8 @@ def _find_dots(view):
 
 @lru_cache(maxsize=1)
 # ^- throttle side-effects
-def _colorize_dots(vid, dots, _change_count):
-    # type: (sublime.ViewId, Tuple[colorizer.Char], int) -> None
+def _colorize_dots(vid, dots):
+    # type: (sublime.ViewId, Tuple[colorizer.Char]) -> None
     view = sublime.View(vid)
     view.add_regions('gs_log_graph_dot', [d.region() for d in dots], scope=DOT_SCOPE)
     paths = [c.region() for d in dots for c in colorizer.follow_path(d)]
--- a/core/commands/log_graph.py
+++ b/core/commands/log_graph.py
@@ -1,14 +1,23 @@
 from functools import lru_cache, partial
 import re
+import threading
 
 import sublime
 from sublime_plugin import WindowCommand, TextCommand, EventListener
 
-from ..git_command import GitCommand
+from . import log_graph_colorizer as colorizer
 from .log import GsLogActionCommand, GsLogCommand
 from .navigate import GsNavigate
-from ...common import util
+from ..git_command import GitCommand
+from ..settings import GitSavvySettings
 from ..ui_mixins.quick_panel import show_branch_panel
+from ...common import util
+from ...common.theme_generator import XMLThemeGenerator, JSONThemeGenerator
+
+
+MYPY = False
+if MYPY:
+    from typing import Iterator
 
 
 COMMIT_NODE_CHAR = "●"
@@ -17,6 +26,8 @@ GRAPH_CHAR_OPTIONS = r" /_\|\-\\."
 COMMIT_LINE = re.compile(
     "^[{graph_chars}]*[{node_chars}][{graph_chars}]* (?P<commit_hash>[a-f0-9]{{5,40}})".format(
         graph_chars=GRAPH_CHAR_OPTIONS, node_chars=COMMIT_NODE_CHAR_OPTIONS))
+DOT_SCOPE = 'git_savvy.graph.dot'
+PATH_SCOPE = 'git_savvy.graph.path_char'
 
 
 class LogGraphMixin(object):
@@ -39,6 +50,7 @@ class LogGraphMixin(object):
         view.set_syntax_file("Packages/GitSavvy/syntax/graph.sublime-syntax")
         view.run_command("gs_handle_vintageous")
         view.run_command("gs_handle_arrow_keys")
+        threading.Thread(target=partial(augment_color_scheme, view)).run()
 
         settings = view.settings()
         settings.set("git_savvy.repo_path", repo_path)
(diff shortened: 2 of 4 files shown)


kaste:
I used the Theme patcher here but I really dislike it: I always have to restart Sublime, it often pop ups an error message, it is really cumbersome to style bc it is not instant. (This is the same experience as with the inline diff.)  Maybe we should just document which scopes we use, or make them customizable.

kaste:
Don't know why one test fails on Windows. The tested functionality def works on my Windows computer. 😢

kaste:
Ok, via 88f1778 proved that the ThemeGenerator makes the tests fail. How could we handle that?

GitSavvy.sublime-settings
```diff
@@ -92,6 +92,12 @@
             "remove_background": "#A83732",
             "remove_foreground_bold": "#F9F9F4",
             "remove_background_bold": "#702820"
+        },
+        "log_graph": {
```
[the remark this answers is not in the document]
kaste: Currently, there is no optionality here. We probably need a new setting/switch for that. 

What do you mean with 'it breaks something we haven't seen yet'. The final implementation should be permissive and not deadlock/freeze. If we see patterns/graphs the algo can't handle, we right now just stop following that char. Hopefully, users will report that so we can add such cases to the algo.

core/commands/log_graph_colorizer.py
```diff
@@ -0,0 +1,225 @@
...
+    def decorator(fn):
+        if ch in registered_handlers:
+            raise RuntimeError('{} already has a handler registered'.format(ch))
+        registered_handlers[ch] = fn
+        return fn
+
+    return decorator
+
+
+class GraphChar:
```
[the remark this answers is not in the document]
kaste: Do you already spot something to optimize? Currently I use the go east, west, south-east properties. I will look at the code again to see if it's easier to read if we just move up, down, left, right. So what's now `c.sw.w` would become `c.down().left().left()`. Probably not, if you just can look at 8 directions in ascii art.

core/commands/log_graph.py
```diff
@@ -188,6 +227,7 @@ class GsLogGraphNavigateCommand(GsNavigate):
     offset = 0
 
     def get_available_regions(self):
+        # return self.view.find_by_selector('keyword.graph.commit.git-savvy')
```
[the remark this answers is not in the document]
kaste: Yeah, done.

core/commands/log_graph_colorizer.py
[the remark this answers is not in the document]
asfaltboy: I honestly don't see that many issues with the code besides the lack of docstrings :) do you want me to write what I am able to understand and then you can correct me?

kaste:
The opportunities we have here: maybe a key to toggle these drawings on/off, maybe [p] to navigate to the parents, [c] to navigate to the children from a commit. 

(Navigating to parents would have been possible before bc we can read the parents from the raw commit data. But this here is way faster. Navigating to the children is always mildly hard just with git commands.)

Both parents and children can be ambiguous, in that case we would need to show a panel. (Or select the nearest?) Still this could work out better than just to navigate up and down visually which is what we have right now. 

Not that I do this right now. First, what do we do with the ThemeGenerator? One pain point is that if we apply our custom theme to all graph views, the user basically looses the ability to style this view and get immediate feedback. The user will first edit the original theme which doesn't work anymore, then she could try and find our patched theme but all edits on it will be lost. It is also right now a hidden theme and edits will not update the view immediately. What are the cons of actually patching the original theme in place? (Of course, we would do a backup.)

asfaltboy:
1. for 88f1778 I tried moving the augment to the bottom of the function without using thread:

```diff
 core/commands/log_graph.py | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)

diff --git a/core/commands/log_graph.py b/core/commands/log_graph.py
index 4148e64..dcb13d4 100644
--- a/core/commands/log_graph.py
+++ b/core/commands/log_graph.py
@@ -50,7 +50,7 @@ class LogGraphMixin(object):
         view.set_syntax_file("Packages/GitSavvy/syntax/graph.sublime-syntax")
         view.run_command("gs_handle_vintageous")
         view.run_command("gs_handle_arrow_keys")
-        threading.Thread(target=partial(augment_color_scheme, view)).run()
+        # threading.Thread(target=partial(augment_color_scheme, view)).run()
 
         settings = view.settings()
         settings.set("git_savvy.repo_path", repo_path)
@@ -59,6 +59,7 @@ class LogGraphMixin(object):
         view.set_name(self.title)
 
         view.run_command("gs_log_graph_refresh", {"navigate_after_draw": True})
+        augment_color_scheme(view)
 
     def prepare_target_view(self, view):
         pass
```

Obviously it's a little bit slower to show up, but it doesn't throw an error for me.

2. on support for color schemes: If I'm not mistaken, to be consistent across color schemes, we would need to leave those schemes alone and define the syntax definition scopes as if we were using the language syntax highlighting to color our elements.

This means that we need to experiment with a theme until we find the colors that highlight the graph dots/lines the nicest. E.g for Monokai it looks like "selected strings" looks appropriate.

Ref:
https://www.sublimetext.com/docs/3/color_schemes.html#global_settings-selection
https://www.sublimetext.com/docs/3/scope_naming.html#string

That said, I don't think it's in the scope of this PR to fix it though...

kaste:
Just moving `augment_color_scheme` did not work. But it is actually easy: If we write a new file, we cannot immeditaley use it, Sublime needs at least one tick to 'register' this new theme and make it loadable/applyable. 🤦‍♂

kaste:
Also, try again: it's now super fast 🏎

kaste:
There was no more time left last week, and next it's Urlaub. I'll iterate (docs, docs, docs) here after that.

stoivo:
Looks and work. Awsome feature. This is a lot more usefull than having each branch in a separate color. I will try to get time to review it in hopefully next week, defenetly september.

core/commands/log_graph_colorizer.py
```diff
@@ -0,0 +1,267 @@
...
+    if str(next_char) in test:
+        yield next_char
+
+
+@handles(COMMIT_NODE_CHAR)
+def follow_dot(char):
+    # type: (Char) -> Iterator[Char]
+    yield from contains(char.e, '-')
+    yield from contains(char.s, '|' + COMMIT_NODE_CHAR)
+    if char.e != '/':
```
stoivo: Would this match
```
 | 
 ●
/|
```
and not 
```
 | /
 ●/    
/|
```
match this one ?

I have never seen that seen that in a graph

core/commands/log_graph.py
```diff
@@ -314,7 +314,7 @@ def on_post_window_command(self, window, command_name, args):
 def colorize_dots(view):
     # type: (sublime.View) -> None
     dots = tuple(_find_dots(view))
-    _colorize_dots(view.id(), dots, view.change_count())
```
stoivo: If the buffer changed I think we should colorize over againg

core/commands/log_graph.py
```diff
@@ -313,10 +313,15 @@ def on_post_window_command(self, window, command_name, args):
...
 def colorize_dots(view):
     # type: (sublime.View) -> None
-    dots = tuple(_find_dots(view))
+    dots = tuple(find_dots(view))
     _colorize_dots(view.id(), dots)
 
 
+def find_dots(view):
+    # type: (sublime.View) -> Set[colorizer.Char]
+    return set(_find_dots(view))
```
stoivo: When does this happen? multiple cursons on one line?

stoivo:
Do you have any objecsion on doing it like 
```
diff --git a/core/commands/log_graph.py b/core/commands/log_graph.py
index 0eb406b..ceab3df 100644
--- a/core/commands/log_graph.py
+++ b/core/commands/log_graph.py
@@ -281,10 +281,15 @@ class GsLogGraphCursorListener(EventListener, GitCommand):
             return
 
         draw_info_panel(view, self.savvy_settings.get("graph_show_more_commit_info"))
+
+    def on_selection_modified(self, view):
+        if not self.is_applicable(view):
+            return
+
         # `colorize_dots` queries the view heavily. We want that to
         # happen on the main thread (t.i. blocking) bc it is way, way
         # faster.
-        sublime.set_timeout(lambda: colorize_dots(view))
+        colorize_dots(view)
 
     def on_post_window_command(self, window, command_name, args):
         # type: (sublime.Window, str, dict) -> None
```

stoivo:
I feel like the 4 listed commit below could be in the first commit `Colorize the path to the next commit`

- Fix type comments
- Rename `GraphChar` -> `Char`
- Reorder functions
- Improve type information

stoivo:
And a last note, I am amazed how much simplier this was than what I expcted. This is supercool. Thank you

core/commands/log_graph.py
kaste: Yes.

core/commands/log_graph.py
kaste: Look at hash function of `Char` which does this check for us. 👍

core/commands/log_graph_colorizer.py
kaste: Yes, basically on the east is never any char. Iirc correctly I handled '*' and '|' in the same function in the first iteration so this is kind of a left over from that probably. 👍

kaste:
If we do it `on_selection_modified` we run one tick earlier. What I want is: all really important sync event listeners run before us, so if other plugins have *really* a reason to run here as soon as possible they can. Since we "only" do visual dances here, we run after them which is fast enough for us.

kaste:
Only could do light fixups bc I get merge conflicts too quickly.

core/commands/log_graph.py
stoivo: > Since a dictionary is used to cache results, the positional and keyword arguments to the function must be hashable.
https://docs.python.org/3/library/functools.html

Does this mean that is calles __hash__?

kaste: Yes!

stoivo:
Code look good. Not so happy about the commits (several does't make any sence for history). Anyway I dont want to bother you about it now

PR #990, opened 2018-08-17
Fix: typo in remote address

randy3k:
as title

PR #1158, opened 2019-08-28
Fix: Git clone URL in README.md

franciscolourenco:
Previous URL was missing colon:
```
fatal: repository 'git@github.com/divmain/GitSavvy.git' does not exist
```

Opted for the https URL because it is the default given by Github.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ matrix:
     include:
         - name: Linux
           os: linux
+          dist: trusty
           language: python
           python: 3.3
         - name: OSX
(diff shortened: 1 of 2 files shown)


stoivo:
Thank you very much. 
https://github.com/divmain/GitSavvy/pull/990 @randy3k  I do clone most of my repos with ssh. GitHub recommens https https://help.github.com/en/articles/which-remote-url-should-i-use. Github does't say anything about why so I assume it is either scaling thing or that have to use a github accout.

kaste:
Should be https bc it's more reliable. Basically it works everywhere.

stoivo:
Since github is recommening it i would go with it.

stoivo:
I will fix CI for you. Thanks for the PR

PR #1161, opened 2019-08-30
Fix: Git clone URL in README.md

stoivo:
Previous URL was missing colon:
```
fatal: repository 'git@github.com/divmain/GitSavvy.git' does not exist
```

Opted for the https URL because it is the default given by Github.


stoivo:
Thanks, I removed  one empty line and pushed the commti to the other branch so I could merge this. Since this PR is missing a review.

PR #1162, opened 2019-09-04
Change the length of the commit messages

luispuerto:
Hey! 

I just think that the default commit message length should be 72, as it's stated in the git guidelines here: 

>Capitalized, short (50 chars or less) summary

>More detailed explanatory text, if necessary.  Wrap it to about 72
characters or so.  In some contexts, the first line is treated as the
subject of an email and the rest of the text as the body.  The blank
line separating the summary from the body is critical (unless you omit
the body entirely); tools like rebase can get confused if you run the
two together.

https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#Commit-Guidelines

Therefore messages for commit command should be also 72 so they don't exceed the character limit and are marked in red. 

I guess this is a hotfix, change it if you think otherwise.

### DIFF
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -149,7 +149,7 @@
     "pedantic_commit": true,
     "pedantic_commit_ruler": true,
     "pedantic_commit_first_line_length": 50,
-    "pedantic_commit_message_line_length": 80,
+    "pedantic_commit_message_line_length": 72,
     "pedantic_commit_warning_length": 20,
 
     /*
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -11,25 +11,27 @@ from ..exceptions import GitSavvyError
 
 
 COMMIT_HELP_TEXT_EXTRA = """##
-## You may also reference or close a GitHub issue with this commit.  To do so,
-## type `#` followed by the `tab` key.  You will be shown a list of issues
-## related to the current repo.  You may also type `owner/repo#` plus the `tab`
-## key to reference an issue in a different GitHub repo.
+## You may also reference or close a GitHub issue with this commit.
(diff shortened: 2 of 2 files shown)


stoivo:
```
pycodestyle --statistics --count
.\core\commands\commit.py:14:68: W291 trailing whitespace
.\core\commands\commit.py:15:70: W291 trailing whitespace
.\core\commands\commit.py:16:66: W291 trailing whitespace
.\core\commands\commit.py:17:63: W291 trailing whitespace
.\core\commands\commit.py:24:68: W291 trailing whitespace
```
fix please

If thats what git is recomending I can agree that we should follow the same convension. And it changeable if anybody want to change

GitSavvy.sublime-settings
```diff
@@ -149,7 +149,7 @@
     "pedantic_commit": true,
     "pedantic_commit_ruler": true,
     "pedantic_commit_first_line_length": 50,
-    "pedantic_commit_message_line_length": 80,
```
stoivo: Update comment in docstring above

asfaltboy:
Some background behind the 50 character choice here:
https://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting

As the last answer points out:
```
$ git help commit | grep -C 1 50
      Though not required, it’s a good idea to begin the commit message with a
      single short (less than 50 characters) line summarizing the change,
      followed by a blank line and then a more thorough description. The text
```

These are all obviously related...

GitSavvy.sublime-settings
luispuerto: What do you mean exactly with this? Sorry I'm a little bit of a noob.

luispuerto:
I think I've fixed the build in the last commit.

PR #1163, opened 2019-09-11
Add log and graph current branch

stoivo:
https://github.com/divmain/GitSavvy/pull/1150#issuecomment-530094785

time.


Default.sublime-commands
```diff
@@ -77,6 +77,10 @@
         "caption": "git: graph",
         "command": "gs_log_graph"
     },
+    {
+        "caption": "git: graph current branch",
+        "command": "gs_log_graph_current_branch",
```
randy3k: the build failed. May be caused by the extra commas?

Default.sublime-commands
```diff
@@ -288,6 +292,10 @@
         "caption": "git: log",
         "command": "gs_log"
     },
+    {
+        "caption": "git: log current branch",
+        "command": "gs_log_current_branch",
```
randy3k: same here

Default.sublime-commands
stoivo: Maybe, I'll check

kaste:
A note: Adding more commands is also a bit problematic, bc you can't easily overwrite those with custom commands. Afaik when a user now adds a command with the same caption they appear twice. E.g. I want to have something like 'log current branch' to actually show only the commits from the tip to the merge-base. I already have a few custom commands, and it's an unsolved question how we can make GitSavvy hackable and customizable for its users.

PR #1164, opened 2019-09-20
Allow flipping `base` and `target` in diff view

kaste:
We already use `<TAB>` to flip between staged and unstaged changes. We can
use the same command to flip `base` with `target`.

I.e. open the branch dashboard, press 'f' on a branch. Now use '<TAB>'.


kaste:
The command still is called `GsDiffToggleCachedMode`. Feel free to suggest a better name.

stoivo:
Looks legit

PR #1165, opened 2019-09-24
Fix: `disable_logging` should reset to previous state

kaste:
The `disable_logging` context manager sets `enabled=True` on exit which is obviously not what we want. We need to restore the previous state.

### DIFF
--- a/common/util/debug.py
+++ b/common/util/debug.py
@@ -32,11 +32,12 @@ def stop_logging():
 @contextmanager
 def disable_logging():
     global enabled
+    previous_state = enabled
     enabled = False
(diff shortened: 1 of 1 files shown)


kaste:
Hm, this code was here since 2938321db00b5d15ab9281b4a99e528d78e90eb7 and acted like a memory leak for most users. If you run Sublime for a month, `_log` must grow to thousands of entries. 🤔 Strange...

kaste:
Note that on my status branch this was fixed because I removed this context manager altogether. (And most of the time I am on my own branch.)

PR #1167, opened 2019-09-24
Simplify `nearest_branch` algorithm

### LINKED ISSUE #1166: Let's talk about `nearest_branch` selection for `rebase`
The `rebase` dashboard preselects a good base candidate automatically. The implementation is imo flawed. Let's look at it.

For the simple case:

image

A: The feature branch I just pushed here is still on top of `dev`, and `nearest_branch` suggest `dev` 👍 

```
nearest_branch: found 1 relatives: ['dev']
nearest_branch: Found best candidate dev
```

Let's look at this. `colorize-paths` is a PR which is already behind `dev`. `follow-paths-upwards` is an upcoming PR derived from the 'colorize' feat.

image

B: If I check out 'colorize' `nearest_branch` fails 😮 and falls back to the 'upstream' `fork/colorize-paths-in-graph-view`, which is a no-op because 'local' and 'remote' are synced. (Without an 'upstream' it would fallback to 'master'!)

```
nearest_branch: found 1 relatives: ['dev']
nearest_branch: No valid nearest found. Possibly on a root / detached branch!
``` 

C: If I checkout `follow-paths-upwards`, it selects `colorize-paths`. 👍 

```
nearest_branch: found 2 relatives: ['colorize-paths-in-graph-view', 'dev']
nearest_branch: Found best candidate colorize-paths-in-graph-view
```

What are good candidates for rebasing?

First, 'dev' is always a good candidate here ('dev' would be ' master' in other repos). 

For B: You have options:

1. 'dev', the branch you originally started your feat which moved along. E.g. you need to rebase on 'master' because of merge conflicts.

2. the actual merge-base, t.i. the commit you forked originally. E.g. you re-arrange the PR but don't want to resolve possible merge conflicts, you want to stay in the feat branch.

3. the upstream if available, t.i. the commit you already pushed. If you rebase from upstream you don't have to force-push afterwards, it's safe.

Since we currently don't offer candidates to the user but just select one good candidate, should we just go cheap and take the *first* relative found.  Are there known cases we do want to handle differently?

The main bug was imo introduced here 2dcc3aacb5057a552f78259f47fa6810820b18e1 with switching from `--no-merged` to `--merged`. `--merged` wasn't correct either bc it breaks C. (For C `--merged` would select 'dev'.) Imo we don't want neither, just omit this arg. But then the check `git branch --contains <common>` cannot fail so we could omit altogether. (We know that the `sha` is in `relative` bc we have it from its log.)

WDYT 🤷‍♀

kaste:
Closes #1166

This is the 'cheap' solution mentioned in #1166. We just take the first relative. By that we go from minimal 4 git commands down to 1.

### DIFF
--- a/core/git_mixins/rebase.py
+++ b/core/git_mixins/rebase.py
@@ -26,43 +26,6 @@ class NearestBranchMixin(object):
                 relatives.append(branch_name)
         return relatives
 
-    def _nearest_from_relatives(self, relatives, branch):
-        """
-        Find the nearest branch from the "branch-out nodes" of all relatives.
-        """
-        util.debug.add_to_log('nearest_branch: filtering branches that share branch-out nodes')
-        diff = difflib.Differ()
-        max_revisions = 100
-        branch_commits = self.git(
-            "rev-list", "-{}".format(max_revisions), "--first-parent", branch).splitlines()
-        for relative in relatives:
-            util.debug.add_to_log('nearest_branch: Getting common commits with {}'.format(relative))
-            relative_commits = self.git("rev-list", "-{}".format(max_revisions),
-                                        "--first-parent", relative).splitlines()
-
-            # Enumerate over branch vs relative commit hashes and look for a common one
-            common = None
-            for line in diff.compare(branch_commits, relative_commits):
-                if not line.startswith(' '):
-                    util.debug.add_to_log('nearest_branch: commit differs {}'.format(line))
-                    continue
-                common = line.strip()
-                util.debug.add_to_log('nearest_branch: found common commit {}'.format(common))
-                break
-
-            if not common:
-                util.debug.add_to_log('nearest_branch: No common commit found with {}'.format(relative))
-                continue
-
-            # Found common "branch-out node", get reachable branches for commit
-            branches = self.git("branch", "--contains", common, "--merged").splitlines()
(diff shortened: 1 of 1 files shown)


randy3k:
It kind of makes sense to me. 
@asfaltboy what do you think? You were the one who wrote it.

stoivo:
I did't touch rebase much, so up to you guys

kaste:
git show-branch also has sorting options. Maybe it is enough to fiddle with them if we run into problems. After all we only want a good candidate, as explained in the issue there are valid other candidates but then we must open a panel and ask the user.

kaste:
I actually don't use the rebase dashboard at all but wanted to use the 'nearest_branch' algo for something else.

asfaltboy:
I think @kaste is right, we had originally considered picking nearest neighbour and be done with it, but then we saw some issues with many similar forked branches (see discussion starting around [this comment][1]).

Apparently, we complicated matters trying to find an earlier common node. In hindsight, we probably should have introduced these two methods separately to have time to test. I have no objection to merging this fix, as it fixes some issues I've been seeing and from my (limited) testing seems to work in all scenarios mentioned.

[1]: https://github.com/divmain/GitSavvy/pull/532#issuecomment-257115375

kaste:
So, let's try this on `dev`.

PR #1168, opened 2019-09-24
Pick #1162 for `dev` branch

kaste:
Change the length of the commit messages

Now the commit messages and the default length of the rulers is 72.

(Surely this wasn't a hotfix 😬, I want this in dev for testing.)

### DIFF
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -149,7 +149,7 @@
     "pedantic_commit": true,
     "pedantic_commit_ruler": true,
     "pedantic_commit_first_line_length": 50,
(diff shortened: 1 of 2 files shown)


kaste:
(Actually how is the release story here anyway. Just never do it?)

randy3k:
> (Actually how is the release story here anyway. Just never do it?)

Sorry for not being around because I was busy moving during the summer. Will try to pick up. @stoivo may know more about current development, are we ready for a release? 

PS: @divmain are you still around? I highly recommend joining the team.

stoivo:
I was thinking to go throught the open PR week, freeze dev, then release.  I feel bad for not having time to look at the PR before.

PR #1171, opened 2019-09-26
Optimize inline diff refresh

kaste:
We generally get down from 4 to 2 git calls per refresh. We esp.
get rid of the `git hash-object -w --stdin` which persists the
object in the database, but these are temporary objects which
probably often needs to get garbage collected later.

Do I miss something special here? 🤷‍♂

### DIFF
--- a/core/commands/inline_diff.py
+++ b/core/commands/inline_diff.py
@@ -188,6 +188,7 @@ class GsInlineDiffRefreshCommand(TextCommand, GitCommand):
     def _run(self, match_position=None):
 
         file_path = self.file_path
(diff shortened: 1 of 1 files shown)


stoivo:
It feels weird, the logic looks over compicated. I ask myself why did Dale write it like that.  So I inversidated where it came from and I found that it wriginates from the beginning  b602cf858ef8

The only think I can think of is that we might have wanted to to be able to undo in a safer way then what we do now. Since the refferences would re persisted longer than the time sublime is open. Since we ar'nt using it I open fo update this.

I will let this one rest until friday night.

stoivo:
Change target to bust and update commit count

PR #1174, opened 2019-10-04
Do not mark obvious parent in graph view

kaste:
If the selected commit has only a parent right below, do not colorize
it for less visual noise.

T.i. in such cases of consecutive commits there is no 'path' to the parent so I think we should omit colorizing like so: 

image

Compare to current dev:

image

WDYT?


stoivo:
👎  I prefer to keep the same  for consistensy.

kaste:
Fair enough. We'll see how it looks if I manage to colorize the path upwards as well.

PR #1175, opened 2019-10-11
Release/2.19.0

stoivo:
Lets make the release. 

https://github.com/divmain/GitSavvy/pull/1151 has some features, but its a bigger PR and we can merge it later and release more frequent.


kaste:
#1137 should be documented too (or did I miss it?).

kaste:
Too many typos in the release message. You slipped one fix into it (fix rebase interactive...). For visibility it is better imo to open a PR and immediately merge it. That way all contributors just get a fyi. 

Otherwise good and happy to see it.

stoivo:
> #1137 should be documented too (or did I miss it?).

No, I skiped it, since I see it as a feature only people really interested would use and they would probably look at the changes too. I have no reason not to add it so I will, does this look ok?
```
   Opening inline diff takes an agument to open inline diff on the same location? More info looat at #1137
```
It make me think that maybe we should use that as a default for `git: diff current file inline` command too?



> Too many typos in the release message.

Thanks for letting me know. I know I not good typing but I defenetly can do better.  I fixed some 

> You slipped one fix into it (fix rebase interactive...). For visibility it is better imo to open a PR and immediately merge it.

Yes, thats right. Except that I can't merge with out someones approvals. I know that this feature  is not documented anywhere, probably would have been good If we did. Still that would be better for everyone I did so. 



> Otherwise good and happy to see it.

Good, Thanks to you for all the grate work!

kaste:
I implemented an old issue with #1137 and so I think it should be visible. I don't think it should be the default for the Command Palette at least that's not how I use it. I have custom key bindings for all diff views, and in general my workflow. 

Not merging without approvals is an odd workflow for such a small project. It's driven by fear. For SublimeLinter we approve via positive comments, and generally we have lots of comments and writing everywhere.

kaste:
Ah, update message is 👍 now. Just add #1137.

kaste:
Maybe you want to give me access to issue triage. Closing, tagging...

asfaltboy:
Unfortunately, we cannot, none of us maintainers have any perms but the commit bit, except the owner @divmain, who we have trouble reaching out to.

kaste:
Oh shit, so basically we or you own the repo bc nobody else is there but then we don't have the permissions to further move the repo. And basically you, the current maintainers, are also moving away, like none of the new issues are tagged, very few PRs over the last year. But two years ago you were all very active.

divmain is gone since 2017, isn't he?

stoivo:
`git: edit and stage diff` is broken

kaste:
Ah well, that's unrelated to this release. It's broken since 2.17.0, and since nobody noticed, that's maybe a wontfix, a feature we could remove after discussing. I would have to try this feature, I mean you can't edit whatever you want in such a view bc git will look at the context lines and reject further edits. You could probably fix typos. But then it should probably an integrated feature of the diff view. 

graph current file is broken since 2.18.0 without anyone filing an issue. 🤷‍♀

PR #1178, opened 2019-10-21
Try to fail github action CI

stoivo:
Just a test to make sure Flake8 still works


stoivo:
Action worked but we should keep this open untill we fixed tests too.

kaste:
Sure it works

PR #1180, opened 2019-10-22
Apply `color-scheme` using relative path to theme

### LINKED ISSUE #1179: Colors are no longer working for inline diff
Please include the following information when filing a bug:

- Sublime Text version number: 3211
- Git version number: 2.23.0
- OS type and version: Linux 5.0.0-32-generic #34-Ubuntu SMP Wed Oct 2 02:06:48 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
   The inline diff no longer displays color: 
image

Error:
```
GitSavvy: The theme GitSavvy.active-diff-view.rb.hidden-color-scheme is not ready to load. Maybe restart to get colored highlights.
```

I am using Gruvbox color scheme. It still works in some other schemes, but not all.

kaste:
Fixes #1179

Partially reverts 023a309

### DIFF
--- a/common/theme_generator.py
+++ b/common/theme_generator.py
@@ -87,7 +87,13 @@ class ThemeGenerator():
         """
         Apply the transformed theme to the specified target view.
         """
-        pass
+        self.write_new_theme(name)
+
+        path_in_packages = self.get_theme_path(name)
+
+        # Sublime expects `/`-delimited paths, even in Windows.
+        theme_path = os.path.join("Packages", path_in_packages).replace("\\", "/")
+        try_apply_theme(target_view, theme_path)
 
 
 class XMLThemeGenerator(ThemeGenerator):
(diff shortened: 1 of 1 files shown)


kaste:
I had to reword the commit message, and now I need a new approval? That doesn't sound too practical for a small team. Anyone has admin rights and can or want to change these rules?

kaste:
appveyor doesn't work. Could you also remove the requirement for that, so we can actually merge and release?

kaste:
(I'm actually no fan of requiring travis either. It may break at any time, and we don't have DevOps we can ping to help us. We're only 4 people. Not 4 people in a company with another x people. We can't also break anything here as there is no build pipeline and no business behind us.)

randy3k:
@kaste I have just marked as you an admin. (I think no one will oppose this idea).

randy3k:
I have removed the appveyor webhook for now. The appveyor service is only possible with @divmain's appveyor.

randy3k:
disabled the status check

kaste:
Then let's 🚢

asfaltboy:
Can't we move appveyor/travis tasks to github actions? I think they support both linux/darwin and windows images!

stoivo:
We can @randy3k said somewhere that we have't run UnitTest in github action yet.

divmain:
@randy3k @asfaltboy @stoivo - is there something you need to get AppVeyor working?  I've just changed the AppVeyor permissions settings that I could find to give you all the permissions you should need.  But please let me know if there's something else.

randy3k:
It looks like Appveyor now supports github team. 
https://www.appveyor.com/docs/team-setup/#setting-up-appveyor-account-for-github-organization
So the best practice for us is to set up an organization account for timbrel (I can do it, but I believe @divmain will be the best person to do so.) and enroll the whole team 
as administrators.


Screen Shot 2019-10-25 at 2 04 38 PM
(I didn't click the blue button because I don't want the whole team accesses to my appveyor account)

randy3k:
But speaking of the functionality, it is now working https://travis-ci.org/timbrel/GitSavvy

randy3k:
But speaking of functionality, it is now working (after I updated the webhook url) https://ci.appveyor.com/project/divmain/GitSavvy

PR #1181, opened 2019-10-22
Colorize word-diff

kaste:
For the diff view, we have a 'word-diff' setting, but the output was annoying.

Let's fix that 😝 

mFpzzXJ9XW

Note that there is actually no one-size regex for word-diffs. Before, we just used `--word-diff` which uses a "default" which doesn't help in our context because it doesn't split enough. (E.g. it does not even split by word, and actually very often still by line.)  So hitting `w` now is a three state toggle, 'off' and then there come two regexes. The first one looks complicated and tries to match words, the second one (simply a `.`) matches single characters. 

Note that all three modes (t.i. the universal diff included) *can* look confusing. It is really to offer choices. 

Note that word diffs often suppress whitespace changes. The previous default did this, and it is probably okay. (That said I didn't find a regex that shows whitespace in a decent or clever way.)

Last: if we had a better ThemeGenerator abstraction this would we a really small PR.

### DIFF
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -175,13 +175,8 @@ def augment_color_scheme(view):
 
 WORD_DIFF_PATTERNS = [
     None,
-    # "Default",
-    # "[^ ]+",
-    # r"[a-zA-Z_\-\x80-\xff]+|[\xc0-\xff][\x80-\xbf]+",
     r"[a-zA-Z_\-\x80-\xff]+|[^[:space:]]|[\xc0-\xff][\x80-\xbf]+",
-    # r"[a-zA-Z_\-]+|.",
     ".",
-    # "[a-zA-Z]+|[^ ]",
 ]
 WORD_DIFF_MARKERS_RE = re.compile(r"{\+(.*?)\+}|\[-(.*?)-\]")
 
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -260,16 +260,24 @@ class GsDiffRefreshCommand(TextCommand, GitCommand):
         self.view.settings().set("git_savvy.diff_view.raw_diff", diff)
         text = prelude + '\n--\n' + diff
 
-        added_regions = []
-        removed_regions = []
+        added_regions = []  # type: List[sublime.Region]
+        removed_regions = []  # type: List[sublime.Region]
 
         if word_diff_regex:
             def extractor(match):
-                a, b = match.span()
+                # We generally transform `{+text+}` (and likewise `[-text-]`) into just
+                # `text`.
+                text = match.group()[2:-2]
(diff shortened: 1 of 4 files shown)


stoivo:
Look good, updated the popup like 

Screenshot 2019-10-22 at 22 30 07

core/commands/diff.py
```diff
@@ -205,9 +265,26 @@ def _run(self):
         self.view.settings().set("git_savvy.diff_view.raw_diff", diff)
         text = prelude + '\n--\n' + diff
 
+        added_regions = []
+        removed_regions = []
+
+        if word_diff_regex:
+            def extractor(match):
+                a, b = match.span()
```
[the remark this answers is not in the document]
asfaltboy: Sounds more like C or Golang people... pythonista, myself included, prefer longer descriptive names... perhaps start and end would work better?

Ref: https://docs.python.org/3/library/re.html#re.Match.span

core/commands/diff.py
```diff
@@ -223,7 +300,10 @@ def run(self, edit, setting):
 
         setting_str = "git_savvy.diff_view.{}".format(setting)
         current_mode = settings.get(setting_str)
-        next_mode = not current_mode
+        if setting == 'show_word_diff':
+            next_mode = (current_mode + 1) % len(WORD_DIFF_PATTERNS)
+        else:
+            next_mode = not current_mode
```
asfaltboy: I like this, but maybe it should be a separate command, e.g `GsDiffCycleSetting` ?

core/commands/diff.py
```diff
@@ -205,9 +265,26 @@ def _run(self):
...
 
+        added_regions = []
+        removed_regions = []
+
+        if word_diff_regex:
+            def extractor(match):
+                a, b = match.span()
+                regions = added_regions if match.group()[1] == '+' else removed_regions
+                offset = len(added_regions) + len(removed_regions)
+                regions.append(sublime.Region(a - offset * 4, b - (offset + 1) * 4))
```
asfaltboy: Either I'm slow or it's late, or both.. but I can't figure out why we multiply by 4 ... any chance we could add a docstring here?

core/commands/diff.py
```diff
@@ -205,9 +265,26 @@ def _run(self):
...
+                return match.group()[2:-2]
```
kaste: The `4` is here in `[2:-2]` in that we shorten the original text by `4` characters for each match. The original text has e.g. `{+foo+}` and we transform it to just `foo`.

kaste:
Let's have a look at it again, folks!

kaste:
LGTM. 

TODO: Refactor the ThemeGenerator thing, this basically adds 50+ lines noise to the file.

kaste:
I just applied the fixup commit.

Issue #1038, opened 2018-11-08
Add a default action in the status dashboard


kaste:
Probably relatively simple to explain: You move the cursor around, and press `<space>` and it will stage _or_ unstage the file depending of the status.

randy3k:
Good idea.

stoivo:
Ehh, what?
Do you want a keybinding on `<space>`?

kaste:
`<space>` is a good candidate for such a keybinding. The key should invoke a 'DefaultAction'. E.g. if the cursor is under a staged file, it will unstage the file. If the cursor is on an unstaged file, it will stage that file. For stashes, there is probably not a good default so it will do nothing here. (You could show the stash.) 

The default action should not be destructive. It is the equivalent of clicking on a file in the git GUI.

asfaltboy:
Let's prioritize this, it supports very common workflows!

kaste:
It's not trivial on top of the current code ... def not a low hanging fruit.

PR #1182, opened 2019-10-22
Iterate on stash view

kaste:
- Make `apply`, `pop`, and `drop` commands more DRY
- Save `stash_id` on the view

  BREAKING: If a user now issues a command e.g. `gs_stash_apply` via
  the Palette, it will select the current open/visible stash (if she
  is looking at such a stash view.)

- Add typical shortcuts for the stash view.

  `<enter>`: will open an action panel to 'apply', 'pop', or 'drop' the
           current stash
  `<a>`:     apply current stash
  `<p>`:     pop current stash
  `<D>`:     drop current stash

TODO:

- [x] If you're looking on a stash, 'pop' and 'drop' should close the window.
- [x] All commands should flash the status bar with some success info.
- [x] Add help.html thingy? Don't remember how this works right now.
~~- [ ] Really difficult to tell if or why we need three commands here, instead of one 
  parametrized?~~
- [x] Helpful message for undo after 'drop'

MOTIVATION:

Mostly psycho-safety. Stashes are part of my daily workflow, and often have lots of them. 'dropping' and 'popping' is destructive (well the reflog but..). Before I 'drop' a stash I usually `ts` and look at it long enough if I really want to discard this work. Then it feels safer to just call 'drop this' from the view, instead of going e.g. to the status dashboard and then to check trice if the cursor is *really* at the correct line.

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -1949,5 +1949,14 @@
             { "key": "setting.git_savvy.vintageous_friendly", "operator": "equal", "operand": true },
             { "key": "setting.git_savvy.stash_view", "operator": "equal", "operand": true }
         ]
+    },
+    {
+        "keys": ["?"],
+        "command": "gs_interface_toggle_popup_help",
+        "args": { "view_name": "stash_view" },
+        "context": [
+            { "key": "setting.command_mode", "operator": "equal", "operand": false },
+            { "key": "setting.git_savvy.stash_view", "operator": "equal", "operand": true }
+        ]
     }
 ]
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -1912,4 +1912,42 @@
             { "key": "setting.git_savvy.stash_view", "operator": "equal", "operand": true }
         ]
     },
+    {
+        "keys": ["."],
+        "command": "gs_diff_navigate",
+        "args": { "forward": true },
+        "context": [
+            { "key": "setting.command_mode", "operator": "equal", "operand": false },
+            { "key": "setting.git_savvy.stash_view", "operator": "equal", "operand": true }
+        ]
+    },
+    {
+        "keys": [","],
+        "command": "gs_diff_navigate",
+        "args": { "forward": false },
+        "context": [
+            { "key": "setting.command_mode", "operator": "equal", "operand": false },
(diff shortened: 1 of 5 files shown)


stoivo:
I don't think we need to make drop an uppercase `D`. `d` should be fine since we have the destrutive warning. Uppervase `D` on branch is to destroy branch before it's merged.

kaste:
I think the `destructive` warning is good for new users, but after a while you sure turn it off, and then it's all or nothing, it's tuned off for *all* such actions. Generally, I have it off, and still are scared. 

I used `D` exactly because of `delete unmerged branch`. Since we have a tie here: @asfaltboy, @randy3k, what do you think `d` or `D` for dropping a stash. Note: dropping is destructive and cannot be undone without trickery. 

Aside: I now print a nice note when dropping a stash so that the commit sha is not lost at least.

asfaltboy:
I'm happy with `D` for this action, for added safety, like with force delete.

 I would also love to be able:
- to enter this `git: stash show` view if we hit "o" or "enter" when in the status dashboard view while focused on a stash line
- to enable vim insert mode on showing the view (so that I can hit "?" to see help, mostly because I'm new to the shortcuts and need the help); the current workaround is hitting "i" and then "?". see https://github.com/timbrel/GitSavvy/blob/77343194ec24dca0d7ee46e5e37de0f912498da5/common/commands/view_manipulation.py#L72-L85

kaste:
> to enter this git: stash show view if we hit "o" or "enter" when in the status dashboard view while focused on a stash line

Yeah, wish it would be trivial to implement. There is #1038.

> to enable vim insert mode on showing the view

Is it as simple as calling `view.run_command("gs_handle_vintageous")`?  Shouldn't this be DRY? All our scratch, read only views should be in `input` mode.  (Then we would call this in `create_scratch_view` (if `read_only == True`).

All views with navigation should maybe call `gs_handle_arrow_keys` as well?

kaste:
I think it is better for testing and motivation for these additive features to get merged rather quickly. It's called 'dev', and this LGTM.

randy3k:
Awesome

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

kaste:
💐

PR #1186, opened 2019-10-25
assume the default value of confirm_force_push is True

randy3k:
I don't know how it happened but it actually did happen that the value of `confirm_force_push` becomes `None` after I disable and re-enable GitSavvy.
Anyway, it is a bit safer to assume the default value of True here.

### DIFF
--- a/core/commands/push.py
+++ b/core/commands/push.py
@@ -23,7 +23,7 @@ class PushBase(GitCommand):
         """
         Perform `git push remote branch`.
         """
-        if self.savvy_settings.get("confirm_force_push"):
+        if self.savvy_settings.get("confirm_force_push", True):
             if force:
                 if not sublime.ok_cancel_dialog(CONFIRM_FORCE_PUSH.format("--force")):
                     return


kaste:
Do you use project or view settings in that project, or does it read the value from the global settings?

randy3k:
There are no project-wise settings. But it only happened when I disabled, re-enabled, used AutoPackageReloader....I am not sure if I could replicate it.

kaste:
I cannot reproduce. What can happen is when you disable the plugin then the modules are still in memory, and you could theoretically call "dead" code, and e.g. the settings would return `None` then. 

The con of this PR is that we basically pull fear into the code base. Can we trust the system, must we duplicate the defaults everywhere?  How do we keep these defaults in sync then. 🤔

randy3k:
There are quite a number of locations where we also put the default value in `settings.get()`. We didn't have consistency anyway.

kaste:
"consistency" is not a good argument when I'm not able to reproduce.  It is likely that this gets reverted over time when we remove these defaults from other call sites.

PR #1193, opened 2019-10-30
Accurate jumping in word diff mode

### LINKED ISSUE #1176: Open file from diff in word-diff mode throws
This is a regression of the accurate jumping. Honestly I hardly knew about the word diff mode, and it's not part of my workflow but the accurate jumping expects a unified diff otherwise it will raise bc it depends on the line mode char +- on each line.

There is no trivial fix I guess bc for accurate jumping we need to differentiate between available and removed chars.

Probably three options.

1. As a hotfix just disable the open functionality.
2. Do inaccurate jumping as before.
3. Just parse it 🤪 as well.

kaste:
Fixes #1176

~Here are two commits. Only the first one is the actual implementation. The second commit splits this functionality into two commands, and uses "proper inheritance" for the late binding.  However, I'm not super confident this is a strong step.  After all, I've learned that for the key-bindings a `setting.x` only matches true bool values, so we have to introduce a settings pair `show_word_diff, word_diff_mode` just for that.  `word_diff_mode` still is of type int, and actually selects which diff we show. `show_word_diff` mirrors this setting, and exposes it as `bool` just for the key bindings.~

~It's probably not worth it.  Keep in mind that we also have duplication and confusion with the "ShowCommit" view/commands. They don't support (inherit) the new word diff color modes, and thus should still be broken after this PR.  ("ShowCommit" afaik is a simplified diff view, but it shows hilarious code reuse with the actual diff view.  E.g. the similar Toggle command over there, and the halfway reuse of this "Open File" command.)~

This is straight forward and will guide us to a probably nice factoring in the future.

### DIFF
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -685,8 +685,14 @@ class GsDiffOpenFileAtHunkCommand(TextCommand, GitCommand):
                     seen.add(filename)
                     yield item
 
+        word_diff_mode = bool(self.view.settings().get('git_savvy.diff_view.show_word_diff'))
         diff = parse_diff_in_view(self.view)
-        jump_positions = filter_(self.jump_position_to_file(diff, pt) for pt in cursor_pts)
+        algo = (
+            self.jump_position_to_file_for_word_diff_mode
+            if word_diff_mode
+            else self.jump_position_to_file
+        )
+        jump_positions = filter_(algo(diff, pt) for pt in cursor_pts)
         for jp in first_per_file(jump_positions):
             self.load_file_at_line(*jp)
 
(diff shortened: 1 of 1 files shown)


kaste:
I reverted the second refactoring commit

core/commands/diff.py
```diff
@@ -738,6 +744,67 @@ def jump_position_to_file(self, diff, pt):
...
+        removed_chars_before_pt = sum(
+            region.size()
+            for region in removed_regions_before_pt
+            if line_start <= region.begin() < pt
+        )
+
+        # Compute the *relative* row in that hunk
+        head_row, _ = view.rowcol(hunk_start)
+        pt_row, col = view.rowcol(pt)
+        rel_row = pt_row - (head_row + 1)
```
kaste: Here is small bug, namely when you're on the `@@ 456,23 ...` line it will not jump to the first content line.  So it basically is accurate if the cursor is somewhere on the content. 

Probably

```suggestion
        head_row, _ = view.rowcol(content_start)
        pt_row, col = view.rowcol(pt)
        rel_row = max(0, pt_row - head_row)
```

kaste: Done

kaste:
This works okay, it is obviously better than throwing :grin:

How we factor this code need to be done in a different PR.

PR #1199, opened 2019-11-08
Refactor/renovate status dashboard

### LINKED ISSUE #1042: Cannot unstage renamed files
- Rename a file
- Open status dashboard
- Stage both the new untracked and the old deleted file
- Staged area show something like this: `    popups/blame_view.html -> popups/blame_view_bogus.html`
- Try to unstage, status bar indicates success, but nothing happens

Actually invokes

```
  {
    "type": "git", 
    "seconds": 0.13777518272399902, 
    "stderr": "", 
    "stdout": "Unstaged changes after reset:\nM\tcore/interfaces/status.py\n", 
    "command": [
      "reset", 
      "HEAD", 
      "popups/blame_view.html -> popups/blame_view_bogus.html"   # <- Ouch
    ], 
    "stdin": null
  }, 
```

### LINKED ISSUE #1210: [status] `<h>` open file on remote doesn't do anything
`GsStatusOpenFileOnRemoteCommand` defers to /calls `gs_open_file_on_remote` which is not implemented.

### LINKED ISSUE #1211: [status] `<o>` can try to open stash names when you have multiple selections
Although `<o>` keybinding has a constraint which works for the simple case, further checks needs to be done in `GsStatusOpenFileCommand` (just like for every other command actually). Esp. for multi-selections

(Since the expressed constraints in -key-binding are never sufficient for our purpose, it is probably better to remove these constraints altogether.)

### LINKED ISSUE #1213: [status] Staging/unstaging multiple files is inefficient
When staging/unstaging multiple files we do this file per file instead of in-one-go. (For discarding, we already do this. 🤷‍♂)

Typical for-in code: 

https://github.com/timbrel/GitSavvy/blob/34b96163b8723315b84dcb85864df6115c098217/core/interfaces/status.py#L535-L540

I also generally wonder if `refresh_repo_status_and_render`shouldn't be in a `finally` clause in such cases. We should probably revalidate the repo status if the action partly fails.

kaste:
Fixes #1042 
Fixes #1211 
Closes #1210 
Fixes #1213 

Basically this is a tedious refactoring. I generally follow the initial 
idea to select/extract information from the drawn view. divmain used
extra regions, and these weren't super accurate, so that we still had to 
strip/split the returned strings. (That's basically why unstaging a rename
failed #1042.) But the syntax stuff in Sublime is fine and fast. So we extract
the desired information from the view just using ordinary selectors.

The constraints we can express in the key-bindings aren't good enough for
our use-case. (divmain mentions this in an commit.) So, instead of having
it in two places, first some constraints in the key-binding definition, 
and then the accurate check in the python code, we of course move to the
python land. (This basically solves some edge cases like #1211.)

### Modify the underlying syntax to help us select the correct things

I modified the syntax to make these changes work. Because
I don't want to introduce visual regressions I use weird scope names. 
These are all `meta` and no theme will have them defined.  

Adapt all commands to the new scopes.

### Try to focus a file when status changes

In the dashboard, it is easy that the cursor gets into meaningless
land after an action. Likely, if after a commit the workspace gets
clean, or after it was clean, it becomes dirty again.

Before, we only moved the cursor `on_new_dashboard`. Now, we determine
after each render if the cursor is still on a meaningful line and if
not move it to the next file.

Couple of notes:

- For the inital draw, this is faster, because we focus
  immediately after the render which is just way earlier than
  `on_new_dashboard`.

- We only jump to files, excluding stash entries, which would become
  annoying. (We don't want to jump to the stashes section after commit
  for example.)

### Tests

I added some testcases so we get notified when we mess up the syntax
definitions.


kaste:
(Oh, we have syntax tests. Will see how these work.)

core/interfaces/status.py
```diff
@@ -386,11 +464,30 @@ class GsStatusOpenFileCommand(TextCommand, GitCommand):
     """
 
     def run(self, edit):
-        lines = util.view.get_lines_from_regions(self.view, self.view.sel())
-        file_paths = (line.strip() for line in lines if line[:4] == "    ")
```
kaste: Here, we had to `strip` and check for `    ` 4 leading spaces because we had no fine grained selectors. This is all over the place herein.

Default.sublime-keymap
```diff
@@ -113,89 +113,79 @@
         "command": "gs_status_open_file",
         "context": [
             { "key": "setting.command_mode", "operator": "equal", "operand": false },
-            { "key": "setting.git_savvy.status_view", "operator": "equal", "operand": true },
-            { "key": "selector", "operator": "equal", "operand": "git-savvy.status meta.git-savvy.status.section meta.git-savvy.status.file" }
```
kaste: We remove all "selector" constraints from the keymap because they were not sufficient anyway.

PR #1207, opened 2019-11-15
fix a status badge regression in ST 4

### LINKED ISSUE #1206: [ST 4] branch status badge no long is shown when there are no views
In the alpha release of ST4, the [old bug](https://github.com/timbrel/GitSavvy/issues/1099) of not showing status badge is regressed. The fix in #1101 no longer works.

The original bug was a regression of https://github.com/timbrel/GitSavvy/pull/1083. Reverting to a commit just before it, the bug is no longer seen.

randy3k:
Related #1206 

Tested on both ST3 and ST4.

### DIFF
--- a/core/commands/status_bar.py
+++ b/core/commands/status_bar.py
@@ -7,16 +7,13 @@ from ...common.util import debug
 
 
 class GsStatusBarEventListener(EventListener):
-
-    # these methods should be run synchronously to check if the
-    # view is transient.
-    def on_new(self, view):
+    def on_new_async(self, view):
         view.run_command("gs_update_status_bar")
 
-    def on_load(self, view):
+    def on_load_async(self, view):
         view.run_command("gs_update_status_bar")
 
-    def on_activated(self, view):
+    def on_activated_async(self, view):
         view.run_command("gs_update_status_bar")
 
     def on_post_save(self, view):
--- a/core/commands/status_bar.py
+++ b/core/commands/status_bar.py
@@ -30,18 +30,16 @@ update_status_bar_soon = False
 def view_is_transient(view):
     """Return whether a view can be considered 'transient'.
 
-    For our purpose, transient views are 'detached' views or widgets
-    (aka quick or input panels).
+    We especially want to exclude widgets and preview views.
     """
 
-    # 'Detached' (already closed) views don't have a window.
+    # 'Detached' (already closed) views and previews don't have
+    # a window.
     window = view.window()
     if not window:
         return True
 
-    # Widgets are normal views but the typical getters don't list them.
-    group, index = window.get_view_index(view)
-    if group == -1:
+    if view.settings().get('is_widget'):
         return True
 
     return False
--- a/core/commands/status_bar.py
+++ b/core/commands/status_bar.py
@@ -10,9 +10,6 @@ class GsStatusBarEventListener(EventListener):
     def on_new_async(self, view):
         view.run_command("gs_update_status_bar")
 
-    def on_load_async(self, view):
-        view.run_command("gs_update_status_bar")
-
     def on_activated_async(self, view):
         view.run_command("gs_update_status_bar")
 


core/commands/status_bar.py
```diff
@@ -39,9 +39,8 @@ def view_is_transient(view):
     if not window:
         return True
 
-    # Widgets are normal views but the typical getters don't list them.
-    group, index = window.get_view_index(view)
-    if group == -1:
+    # disable if the view is not active (widgets is never active)
```
asfaltboy: `widgets is never active` should be `widgets are never active`

kaste:
Please test again. Since we actually only want to filter "widgets" we can do so using the traditional `settings().get('is_widget')` which would be way more readable. 

I also did proper commit messages here. We can be sloppy here and there, but that was a too sloppy commit message imo.

randy3k:
Actually, the reason that I didn’t use the `is_widget` is because a regular view could also set this settings to `True` to get rid of a lot of things, for example some TextCommand's or EventListener's

It happens quite commonly in packages which need to use a regular view to render special information. Particularly, I was using this hack in Terminus.

kaste:
Ah, interesting. I suggest you leave comments in the commits :grin:

More seriously, I usually understand/visualize the status bar as bound/related to the view. So, following this, since Terminus is not under source control I never "realized" this as a missing feature. But I do see now that e.g. the Find results view has the branch information. It is obviously not marked with "is_widget". 

So, although the views indicate via "is_widget", hey we're very special, we should still run on them? Which other packages do this? Since Terminus is very special, we could also explicitly do a whitelisting here.

randy3k:
I actually got the idea of setting `is_widget` from [here](https://github.com/facelessuser/sublime-markdown-popups/blob/master/st3/mdpopups/st_code_highlight.py#L227).

Whitelisting is good enough for me. We could check `view.setting().get("terminus_view")`.

kaste:
Fun fact, I mostly tried above code with Terminus and deduced that output panels have `is_widget`. 🙄  Let's see if we can go with something like `view in window.views()` which is readable and semantically what we want: a view which is "gettable" from a window, t.i. a pretty normal view an end-user uses.  It is not an output panel, nor widget.

randy3k:
I have explicitly whitelisted Terminus. Actually, I guess it also makes sense to run `gs_update_status_bar` in Terminus in case that users perform git operations in Terminus.

core/commands/status_bar.py
```diff
@@ -24,16 +24,20 @@ def on_post_save(self, view):
...
-    We especially want to exclude widgets and preview views.
+    We especially want to exclude widgets, preview views.
     """
 
     # 'Detached' (already closed) views and previews don't have
     # a window.
     window = view.window()
     if not window:
-        return True
+        return False
```
kaste: I don't follow. Am I drunken or is this a mistake?

core/commands/status_bar.py
```diff
@@ -24,16 +24,20 @@ def on_post_save(self, view):
 def view_is_transient(view):
     """Return whether a view can be considered 'transient'.
 
-    We especially want to exclude widgets and preview views.
+    We especially want to exclude widgets, preview views.
```
kaste: ? Why no "and"? You can do ", and" of course.

randy3k:
Oops, I was testing something..

randy3k:
Jon:
> FWIW `window.get_view_index()` will be changed in 4053 to be compatible with ST3

I am not sure if this is still necessary.

kaste:
@randy3k I do not quite understand the UX with Terminus here.  If I start the usual bash, I initially get the correct information in the status bar, okay, but of course Terminus being a terminal it will not notify back if you do some git actions in it which results in information being shown not in sync. bash itself shows the status e.g. in the prompt, and Sublime probably the same in the built in widgets we now have, and GitSavvy shows the outdated information on the left in the status bar. 

What am I missing here?  Why do you want this explicitly? "Blacklisting" makes more sense, no?

(FWIW: After some commands I have I call `refresh_interface` after finish. But for long living shells in Terminus there isn't that option.)

randy3k:
> Why do you want this explicitly? 
Basically you have answered it..
> Terminus being a terminal it will not notify back if you do some git actions in it which results in information being shown not in sync

My OCD cannot stand the badge information being outdated.

randy3k:
Lol, I thought we won’t need to merge this because the bug is fixed my upstream.

kaste:
It had enough useful stuff.

randy3k:
We need to explicitly whitelist `Terminus` though if we use `view.settings().get('is_widget')`.

kaste:
> My OCD cannot stand the badge information being outdated.

What the hell does that mean, LOL.  I said Terminus being a long running process does not notify us when we execute git commands in it. So *our* information in the status bar usually gets our of sync which I dislike because then *my* OCD comes in and does a `ctrl+tab, ctrl+tab` move to retrigger it. (The Sublime badges are never outdated because they have change notifiers installed on the file system.) 

So, will you somehow trigger GitSavvy from Terminus as well?

randy3k:
> So, will you somehow trigger GitSavvy from Terminus as well?

This is what I am thinking about. Perhaps we could run `gs_update_status_bar` every once in a while when there is something new to render. To prevent too frequent updates, we may limit, say, only update the status bar every few seconds.

kaste:
You cannot poll `git status`. It itself takes multiple seconds on normal projects. (This works only on GitHub OSS projects, not on 10000 files projects.)

kaste:
Maybe check mtime of `.git/HEAD` and `.git/index`?

randy3k:
> Maybe check mtime of `.git/HEAD` and `.git/index`?

I am not sure if we have access to the current working directory of a tty. 
Originally, I was just thinking of firing blindly `gs_update_status_bar`.

kaste:
This is generally tricky. I don't think Terminus should do this though. GitSavvy could detect if a user enters/leaves a Terminus view and then starts a smart polling. But i.e. `.git/HEAD` moves too fast when rebasing. So this can get more complex rather quickly. We also don't want to do this when you start a REPL.

PR #1208, opened 2019-11-15
[ST 4] set a default value for syntax

randy3k:
Also in ST 4. For a brief moment before the syntax is applied, `view.settings().get('syntax')` could be `None` and an exception is thrown (Nonetype is not iterable).

### DIFF
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -194,7 +194,7 @@ class GsPedanticEnforceEventListener(EventListener, SettingsMixin):
     """
 
     def on_selection_modified(self, view):
-        if 'make_commit' not in view.settings().get('syntax'):
+        if 'make_commit' not in view.settings().get('syntax', ''):
             return
 
         if not self.savvy_settings.get('pedantic_commit'):


kaste:
This is unlikely a ST4 issue but an edge case we introduce. We apply the syntax in the worker thread but the the selection handler runs on the main thread. Thus we can always get an event before the syntax is even applied. Namely: between creating the view and assigning the syntax we're in an undefined state.

Did you made an explcit reproduction like

```
view.assign_syntax(syntax)
assert view.settings().get('syntax') == syntax
```

randy3k:
This is the complete trackback in case you are interested. 

The error is thrown even when opening a regular file, not only the commit view. I think it is due to a subtle change of ST 4 in the way to handle `on_selection_modified`. I am guessing that `on_selection_modified` run asynchronously and sometimes it does so before the syntax is applied.


```py
  File "/Applications/Sublime Text.app/Contents/MacOS/Lib/python33/sublime_plugin.py", line 1305, in run_
    return self.run(edit, **args)
  File "/Users/Randy/Dropbox/Sublime Text 3/Packages/GitSavvy/common/commands/log.py", line 37, in run
    {'characters': msg, 'force': True, 'scroll_to_end': True})
  File "/Applications/Sublime Text.app/Contents/MacOS/Lib/python33/sublime.py", line 939, in run_command
    sublime_api.view_run_command(self.view_id, cmd, args)
  File "/Applications/Sublime Text.app/Contents/MacOS/Lib/python33/sublime_plugin.py", line 772, in on_selection_modified
    run_view_callbacks('on_selection_modified', view_id)
  File "/Applications/Sublime Text.app/Contents/MacOS/Lib/python33/sublime_plugin.py", line 616, in run_view_callbacks
    callback(v, *args)
  File "/Applications/Sublime Text.app/Contents/MacOS/Lib/python33/sublime_plugin.py", line 128, in profiler
    return event_handler(*args)
  File "/Users/Randy/Dropbox/Sublime Text 3/Packages/GitSavvy/core/commands/commit.py", line 197, in on_selection_modified
    if 'make_commit' not in view.settings().get('syntax'):
TypeError: argument of type 'NoneType' is not iterable
```

kaste:
The surprising moment could be that 'syntax' is initially `None` before it gets some syntax applied. T.i. if you apply a syntax this operation is probably sync. 

You should for sure report something upstream because that's the stuff you must report for alpha software.

Issue #1201, opened 2019-11-10
'NoneType' object has no attribute 'settings'


kaste:
```
Traceback (most recent call last):
  File "C:\Users\c-flo\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\git_command.py", line 220, in git
    stdout, stderr = self.decode_stdout(stdout), self.decode_stdout(stderr)
  File "C:\Users\c-flo\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\git_command.py", line 277, in decode_stdout
    fallback_encoding = self.savvy_settings.get("fallback_encoding")
  File "C:\Users\c-flo\AppData\Roaming\Sublime Text 3\Packages\GitSavvy\core\settings.py", line 12, in get
    project_savvy_settings = view.settings().get("GitSavvy", {}) or {}
AttributeError: 'NoneType' object has no attribute 'settings'
```

https://github.com/timbrel/GitSavvy/blob/b832d08a551c400a43a0ce0d6634e9c9a860e0c3/core/settings.py#L4-L13

The code quality is mediocre.  `self.parent` is unused, what could be its purpose?  Getting the dependencies `window` and `view` here freshly, on each hit is slow and can lead to unwanted results in case the user switches to another window or view in between. (After all, we mostly do slow `git` calls, some in sequence, sure the world can change in say 1s.)

PR #1209, opened 2019-11-16
[ST 4] check if view is None in settings

randy3k:
It seems that `sublime.active_window()` could be None in ST 4.
If that is the cause, we read settings directly from window.project_data()

### DIFF
--- a/core/settings.py
+++ b/core/settings.py
@@ -9,17 +9,22 @@ class GitSavvySettings:
     def get(self, key, default=None):
         window = sublime.active_window()
         view = window.active_view()
-        project_savvy_settings = view.settings().get("GitSavvy", {}) or {}
 
-        if key in project_savvy_settings:
-            return project_savvy_settings[key]
-
-        # fall back to old style project setting
-        project_data = window.project_data()
-        if project_data and "GitSavvy" in project_data:
-            project_savvy_settings = project_data["GitSavvy"]
+        if view:
+            project_savvy_settings = view.settings().get("GitSavvy", {}) or {}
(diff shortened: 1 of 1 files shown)


kaste:
This make it relatively complex. We should go the other way. The main dependency ("arg") here is a `view`.  So we generally should pass down a view to get the settings. `GitSettings(view).get('...')`. We only use a `window` for what we already deprecated.  So in this round we can easily also remove this code. 

Since we only ever get a single key from the `view.settings()` namely `GitSavvy` which itself is then a mapping with the same shape as our global settings. We should **eager** read this setting. Basically we hereby freeze the settings as soon as we construct the object. 

This has great performance impact because querying a view from a thread is expensive. Frozen settings are of course predictable because they cannot change. 

The burden would be on the caller to provide a `view`. The caller can decide to not do the action for example for views which are already closed, gone.  Or the caller decides to just use the active view from the window. The point is, the caller has to make this decision because a sole `GitSettings()` as what we have now (t.i. without args) doesn't know and cannot know which view is applicable as the user can change views and windows whenever she wants. 

See #1201

randy3k:
One consideration is that this mixin is used mixedly in WindowCommand's, ViewCommand's and EventLisenter's. The code would be simplified here, but it will get more complex in the places actually calling it.

kaste:
The `SettingsMixin` is only mixed into `GitCommand` and otherwise used *once* in `commit.py`.

randy3k:
@kaste Thank you for the checking.

randy3k:
I don't think it is needed anymore.

PR #1216, opened 2019-12-09
Fix/Enhancement: Improve newline handling in IgnoreMixin

joreiff:
Python automatically converts newlines ('\n') for files opened in text
mode (e.g. 'at') to the OS's preferred line endings ('\n' on Linux and
'\r\n' on Windows). The manual line-ending handling in the IgnoreMixin
class therefore leads to extra '\r' bytes on Windows. This can be fixed
using the 'newline' argument of open().

Further, the IgnoreMixin currently always appends a newline before and after
the new entry. This leads to every second line being blank. The second
commit fixes the issue by appending the first newline only if the file does not
already end with a newline.

### DIFF
--- a/core/git_mixins/ignore.py
+++ b/core/git_mixins/ignore.py
@@ -23,5 +23,15 @@ class IgnoreMixin():
                 linesep = os.linesep
 
         gitignore = os.path.join(self.repo_path, ".gitignore")
-        with util.file.safe_open(gitignore, "at", encoding="utf-8", newline=linesep) as ignore_file:
-            ignore_file.write("\n" + path_or_pattern + "\n")
+        with util.file.safe_open(gitignore, "a+t", encoding="utf-8", newline=linesep) as ignore_file:
+            self._ensure_on_new_line(ignore_file)
+            ignore_file.write(path_or_pattern + "\n")
+
+    @staticmethod
+    def _ensure_on_new_line(ignore_file):
+        ignore_file_size = ignore_file.tell()
+        if ignore_file_size:
+            ignore_file.seek(ignore_file_size - 1)
+            ignore_file_end = ignore_file.read(1)
+            if ignore_file_end != "\n":
+                ignore_file.write("\n")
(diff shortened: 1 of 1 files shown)


kaste:
That's cool, I stumbled upon this code as well. But shouldn't we just read the file, splitlines, append, join again and add a newline. Would that be more readable? The file has only a couple of lines so performance doesn't count. Wdyt?

joreiff:
Sure, I will update the PR tomorrow. I only did it this way because the existing code used the file object directly. Working with the whole content as a string, on the other hand, will simplify the implementation.

joreiff:
On second thought, I do not think this would simplify the code much. We would still have to check for an empty string (when using `.read().split()`) or a string without `\n` (when using `.readlines()`) and special-case empty files (no lines to check). Additionally, appending seems like a safer operation than overriding. Admittedly, the chance of data loss is very slim, but why risk it if there is no need to?

Instead, how about simplifying `_ensure_on_new_line` to something like the following?
```
@staticmethod
def _ensure_on_new_line(ignore_file):
    ignore_file.seek(max(ignore_file.tell() - 1, 0))
    if ignore_file.read(1) not in (None, "\n"):
            ignore_file.write("\n")
```

kaste:
Hm. I thought read + splitlines would handle all that 🤔

joreiff:
Ah, you are right, I was thinking about `split()`, not `splitlines()`. Then how about:
```
if os.path.exists(gitignore):
    with util.file.safe_open(gitignore, "r", encoding="utf-8") as fp:
        ignore_lines = fp.read().splitlines()
else:
    ignore_lines = []

ignore_lines += [path_or_pattern, ""]
with util.file.safe_open(gitignore, "w", encoding="utf-8", newline=linesep) as fp:
    fp.write("\n".join(ignore_lines))
```

kaste:
That would be code I could read without blinking at least 🤫😁

joreiff:
PR updated

kaste:
Works and is readable. Thanks 😎

PR #1220, opened 2019-12-19
Implement intra-line diffing in the diff view

### LINKED ISSUE #527: Feature request: add diff highlight to all diff views
Currently it's only used in inline diff, and I would personally benefit having it in diff / diff cached views.

Current implementation discussed in  #141

### LINKED ISSUE #800: Highlight changed words in commit-diffs
Would be nice if the commit-diffs (enabled by settings `"show_commit_diff": true`) would highlight the word differences in the displayed lines, like it's done in the "diff file inline"-mode:
screen shot 2017-10-30 at 11 59 12

### LINKED ISSUE #1203: `gs_show_file_at_commit` with given line has cursor/visual glitch
wdUXYoW1ED

Here I'm in the show commit interface and press `o` to open the file at the hunk position. Notice how the cursor appears on pos `(0, 0)`, the highlighted/active line is correct though, and when I move the cursor `right` it appears at or jumps to the correct position.

Basically a visual glitch/bug.

### LINKED ISSUE #1228: [show-file-at-commit] Title showing "COMMIT: ..."
The title is wrong because we're showing a "FILE" not a commit. (Probably copy pasted from the "show-commit" command.)

kaste:
Fixes #527 
Fixes #800 

Implements intra line diff colorization for the diff view and the show commit panel. Like so:

image

image

Before actually implementing the intra diff, I had to refactor the mess I created before.  (Or, I switched one mess for the other. 🤷‍♀) The actual feature takes roughly ~ 250 lines after that.

We expose a single entry function `annotate_intra_line_differences(view)` which operates autonomous. Usually you call this function directly after a draw on the main thread.  It will then efficiently annotate the viewport and then switch automatically to the main thread to annotate the other chunks.

For the actual intra diff we compare line by line, either on the actual string (as a sequence of chars) or if that yields to many changes we tokenize the string and compare again.  Comparing line-by-line usually yields better looking results and is way faster than comparing whole chunks.  Some fine-tuning might still be necessary, t.i. in some situations it might be actually better to compare whole chunks, or to skip comparing altogether, e.g. if the numbers of "+" and "-" lines differs vastly or is just very high. Unfortunately, I couldn't think of any algorithm to detect these yet. 

The introduced base type for everything is a `TextRange` which captures a `text` portion with its region/slice `a b`. This works out nicely for the previous algorithms we had, e.g. jumping to a hunk in the original file etc. There must be some overlap with the already present inline-diff view (which actually does something similar) but I couldn't figure that out as well. 

For the annotations, I switched to the builtin scopes "diff.inserted.char" and "diff.deleted.char" just post-fixed with `.git-savvy`. (These are used for the inline diffs Sublime provides in its newer versions hence every user with every theme has them.) 

Further work: 

Also, enabled for the prepare commit message view ✔️, and the for the show commit view. (Maybe there are other views as well?)  

This works for combined diffs without much ado, although the dynamic highlighting now is smarter than our "syntax". E.g.

image

Here we now dynamically highlight and detect the first changeset which otherwise is not recognized as such by our syntax (likely because of the leading space).

Drive-by fixes:

- Fix parsing of diffs with e.g. renames, t.i. diffs which include file(-headers) without any hunks. 

- Proper extract the commit headers of diffs

- Fix #1203, fix #1228 

- Proper parse combined diffs (diffing 3 or more sources), and fix the jump to file code accordingly

- Performance: only ask git for the short hash if we actually have a long hash to begin with in "show_commit" and "show_commit_info"

- Do not reset cursor when toggling whitespace or word-diff in "show_commit" view


kaste:
The general highlighting algorithm now runs on the show commit *panel* so you can easy-test this on all commits in combination with our graph view.

kaste:
Since it is basically Christmas now, I can't go to much details here. The performance is already okayish now, the results also look promising. Generally I compare line by line now (before: hunk by hunk), and also often tokenize the input line which can give way better results. 🎅

kaste:
That's not in a bad state now.  It is ready to get some reviews, comments.  

Nothing is squashed and maybe polished, but there are also interesting to read parts here, code wise.

kaste:
I'm green and have updated the original comment.

nh2:
Fantastic, it looks great!

Could this be made work for `git: diff current file inline` as well?

kaste:
Thanks! 😸 Sure this can be used for the inline diff view but I couldn't quite "get" the old code (and the old code is shared with some other components, so I could not just replace it...) The old code is somehow specialized for `-U0` diffs iirc whereby the new code should (🤞) handle all the diffs.  Then, I was thinking that the next "inline" diff should maybe just use real inline annotations straight in the editable view. 🤷‍♀

PR #1237, opened 2020-01-23
Permissive reloader

### LINKED ISSUE #1205: `Reload` cmd should be available even if some modules fail to load
If you introduce i.e. a syntax error in one file, then reload. The reloader will complain, but after that the command is not available anymore. So you have to restart Sublime.

(In SublimeLinter the reloader itself is still available after hard errors which is more convenient.)

kaste:
Fixes #1205

Before we reload we `unload_module` every plugin. In case of a hard
error while reloading (e.g. often a syntax error), we actually try
to bring the old modules back via `reload_missing` but did not register
the actual plugins. We thus have all the code somehow (inconsistent)
loaded but none of the commands for the plugins which actually raised.
(In GitSavvy we only have *one* plugin, so actually no commands are
registered after that.)

With this patch we call `reload_plugin` again after failures. The goal
is to keep at least the reload command itself working.

This is good for developers but could be bad for users. However, we
really don't expect reload failures for end-users.

### DIFF
--- a/common/util/reload.py
+++ b/common/util/reload.py
@@ -112,7 +112,7 @@ def reload_package(pkg_name, dummy=True, verbose=True, then=None):
 
     # Reload packages
     try:
-        with intercepting_imports(all_modules, verbose), importing_fromlist_aggresively(all_modules):
+        with intercepting_imports(all_modules, verbose), importing_fromlist_aggressively(all_modules):
             for pkg_name in packages:
                 for plugin in package_plugins(pkg_name):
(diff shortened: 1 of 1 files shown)


kaste:
With `reload_plugin` we actually restore old behavior, t.i. in a previous version of the reloader we did exactly that. (But AutomaticPackageReloader does not!)

What I don't understand is why we only `reload_missing` before, instead of putting *all* modules back which feels more like a complete rollback.

PR #1249, opened 2020-01-30
Iterate graph view

### LINKED ISSUE #1198: [commit compare] Must augment color scheme for the colorize path feat to work
image

The commit compare view (That view could really be better though!) shares enough code to run the "follow path" code, however it doesn't share enough code to have the augmented color scheme applied as well.

### LINKED ISSUE #1239: [graph view] Reuse available views
For graph views we currently always create a new view. Instead we should just focus and reuse existing views if their content match. (This is consistent with all the other other special views we have.)

### LINKED ISSUE #1241: [graph view] cursor should follow HEAD or a branch after refresh
For long-lived graph views, the cursor tends to get on the wrong line (also updating the commit panel) if you add commits or checkout etc.  Instead we should try to stay on the tip of the current branch or on the specific commit the user was on.

kaste:
Closes #1198
Fixes #1239
Fixes #1241

Iterate log graph view:

- Add a help/prelude at the top

- The cursor now follows the commit/branch when refreshing

- Remove the "special" commit compare view. 
  This view did a A..B vs B..A log, we now show `A B (merge-base A B)^!` which is 
  generally very nice. 

- Deprecate the old action panel for the command `git: graph`. We now just show the
  graph, and scroll to/follow HEAD.

- WIP shortcuts `[a]` to toggle `--all`, `[h]` to navigate to the HEAD commit

- Very good performance by reusing the views. Try from the branches dashboard: `[g]` on 
  some branches, or `[H]` to compare. 

- Hard deprecate `git_graph_args` setting since we need full control over the output format for the features to work.

Notable:

- Implement a generic text command (in `runtime.py`).
- Do not trigger "refresh" when closing widgets. E.g. every interaction with the Goto anything or Command palette always triggered a complete refresh which is really too much.


kaste:
Current state

image

kaste:
Interesting performance.  For a moderate repo (~9000 commits) the `git log` call is 3x faster then the actual render. The drawing already blocks here for ~1 second (on my cheaptop). That's a surprise. I thought I need to optimize the git call first.

kaste:
Merge for now, work will continue in other PR's, not that this isn't usable. 

We will maybe or probably have some regressions where the UI doesn't update after a destructive git action. That is because we switch here to manual triggering mode and naturally when some code path does not call "refresh" when it should the views don't update until a `ctrl+tab, ctrl+tab` dance.

PR #1254, opened 2020-02-21
Normalize (live) panel output

### LINKED ISSUE #1020: Translate/strip color codes from push log

kaste:
Fixes #1020 

For the panel output, generally normalize newlines to `\n` only. 

For #1020 we already had a patch at ea98d36 which stripped ANSI sequences from `self.git` output. This is probably not what we want (the OP sees ANSI sequences in the push log, this can only mean in the output panel) because the return value of `self.git` is not passed to the panel. So we revert said commit and instead strip for the panel only.

### DIFF
--- a/common/util/log.py
+++ b/common/util/log.py
@@ -1,23 +1,27 @@
 import sublime
 
 
-def panel(*msgs, run_async=True):
-    msg = "\n".join(str(msg) for msg in msgs)
+def universal_newlines(string):
+    return string.replace('\r\n', '\n').replace('\r', '\n')
+
+
+def panel(message, run_async=True):
+    message = universal_newlines(str(message))
     view = sublime.active_window().active_view()
     if run_async:
         sublime.set_timeout_async(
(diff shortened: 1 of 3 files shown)


kaste:
For the `rebase --interactive` git produces `\r` line-endings. Ideally we would handle them just like on the command line, and replace such lines t.i. not transform them into `\n`.  Otherwise we get 

```
Rebasing (7/21)
Rebasing (8/21)
Rebasing (9/21)
Rebasing (10/21)
Rebasing (11/21)
Rebasing (12/21)
Rebasing (13/21)
....
```

and so on t.i. we print what is basically a progress marker.

PR #1263, opened 2020-03-08
Redo the graph actions

### LINKED ISSUE #1260: Allow creation of lightweight tags
`QuickTag` (!) should also allow to create light tags t.i. the message should be optional. Currently it aborts silently.

kaste:
Fixes #1260 

The graph actions were really limited. We add all commands usually found e.g. in `git k` or possibly `Sublime Merge`. 

Esp. 

- Checkout a branch (not only a commit detached)
- Create and delete tags
- Reset HEAD to 
- Allow cherry-picking and reverting multiple commits

image

Add `[f]` (unset `F`) to enter additonal filters via a typical input panel.  That's not the best UI because it requires the user to know all the flags (it is thus not really accessible) but it is otherwise just too useful. 

Examples: 

- set `--reflog` to immediately display "gone" commits e.g. after a rebase or deletion of a tag, branch. 

- set `-Ssearch_term` or `-Gsearch_term` to quickly pick-axe
- quickly limit via `-n` or `--since` etc.

Probably all these use-cases could be supported more directly, e.g. since we're in an editor, ususally just selecting a word and tell git to pick-axe should be probably a command accessible in the editor views. And "reflog" could be a switch.

But I guess we have add and fine-tune this over time. (No excuse to not add something immediately useful.)

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -1296,6 +1296,30 @@
         ]
     },
     {
+        "keys": ["b"],
+        "command": "gs_log_graph_edit_branches",
+        "context": [
+            { "key": "setting.command_mode", "operator": "equal", "operand": false },
+            { "key": "setting.git_savvy.log_graph_view", "operator": "equal", "operand": true }
+        ]
+    },
+    {
+        "keys": ["f"],
+        "command": "gs_log_graph_edit_filters",
+        "context": [
+            { "key": "setting.command_mode", "operator": "equal", "operand": false },
+            { "key": "setting.git_savvy.log_graph_view", "operator": "equal", "operand": true }
+        ]
+    },
+    {
+        "keys": ["F"],
+        "command": "gs_log_graph_reset_filters",
+        "context": [
+            { "key": "setting.command_mode", "operator": "equal", "operand": false },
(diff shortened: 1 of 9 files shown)


kaste:
Still missing: `Move 'branch_name' to <commit|HEAD>` which does a `checkout -B branch_name`. 

Basically we had `reset HEAD to selected commit` vs `move selected branch to HEAD`

kaste:
- Follow ups: The usage of a "TypedDict" lead to a mess here. Probably just use a custom type with some properties.

- The reset command was not ready to receive a `commit_hash` to operate on, and the mixed-in `LoginMixin` could not handle that case either. So just to do that I had to modify two classes, one of them `LoginMixin` used basically everywhere. (Why shouldn't that possibly break anything or everything?)  Longterm: A `LoginMixin` we actually use often would be a short `Command` (`ProvideCommitHashCommand`) we can use when we require a `commit_hash` but the user did not provide any in the args. This "provider" should then open the panel, asking for a commit, and then call the original command with the additional arg `commit_hash`.

- The PR would benefit from a speedy `get_short_hash`, and generally a cached `get_remotes()`.

PR #1265, opened 2020-03-08
Fix: Allow unstaging merge conflicted files

kaste:
Use-case: If you really dislike the changes you can unstage and then
discard the introduced changes.


kaste:
And it's only a oneliner.

PR #1266, opened 2020-03-08
Iterate on `show commit`

### LINKED ISSUE #1204: Combine commit info panel and commit info view which are roughly the same
We have the Show Commit view with key bindings like `o` to open the file etc., but the show commit *panel* has no features at all, not even navigation.  That's not very intuitive imo.

### LINKED ISSUE #1242: [show commit] Open graph view to see the context of the commit
While looking at a specific commit, the user might want to see its context.  We thus should provide a keybinding/command which opens the graph/panel scrolled to that commit.

kaste:
Fixes #1204 
Fixes #1242 

- Enable keyboard shortcuts for the commit info *panel* (#1204)

- If the commit you're looking at is actually the `HEAD`, open the workspace file on `[o]`. (T.i. the normal writable version.)

- Add `[O]` to open the workspace (checked out) file a hunk is showing. 

(T.i. `[o]` shows the historical version of the file (readable) except you're looking at the HEAD commit, and `[O]` shows the current version of the hunk (read- and writeable).)

- Add `[g]` to open a graph focusing the current commit. (Aka: What is the context of this commit?)

Minor syntax improvements: highlight references (for example to issues or users) in the commit message, style commit meta header as a block.

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -709,6 +709,14 @@
         ]
     },
     {
+        "keys": ["g"],
+        "command": "gs_show_commit_open_graph_context",
+        "context": [
+            { "key": "setting.command_mode", "operator": "equal", "operand": false },
+            { "key": "setting.git_savvy.show_commit_view", "operator": "equal", "operand": true }
+        ]
+    },
+    {
         "keys": ["s"],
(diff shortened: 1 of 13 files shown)


kaste:
- Would benefit from a faster `get_short_hash()` fn

PR #1274, opened 2020-03-14
Add help for "g" to show in graph in commit popup

asfaltboy:
as introduced in #1242


kaste:
Yeah, I was hesitant to add all documentation because it is a bit in flux mode where I add some workflows via shortcuts but don't know yet if they always lead to a good linking between each of our views.

But `[g]` is probably safe in this regard.

asfaltboy:
Thanks Kaste, I found this very useful, saves time via for this workflow:

  blame file -> find commit hash in graph to see context

PR #1280, opened 2020-03-27
Add "delete branch" as graph action

kaste:
We add a "delete branch" command to the graph actions. In the graph you can clearly see for example if a branch has been merged, and is not needed anymore. So this is a good fit.

We generally extract the command from the `branch` dashboard, and add the feature to offer a forceful deletion if the branch is not fully merged yet. (This is thus available for both views.) Before we just showed the git error to the user.

### DIFF
--- a/core/commands/__init__.py
+++ b/core/commands/__init__.py
@@ -1,5 +1,6 @@
 from .amend import *
 from .blame import *
+from .branch import *
 from .changelog import *
 from .checkout import *
 from .cherry_pick import *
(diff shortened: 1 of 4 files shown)


kaste:
Implementation problems:

- `show_branch_panel` should call `on_done` callback only if the user selected a branch, t.i. not on abort. 

- Catching a `GitSavvyError` is cumbersome again. Esp. that constructing such an error object already has the side-effect of showing the panel is 😕

kaste:
To consider:

- Deletion is not fully destructive esp. since we show an undo message. We may get rid of the `actions.destructive` dialog.

PR #1281, opened 2020-03-27
Simplify `gs_custom`

### LINKED ISSUE #1212: `GsCustomCommand` should call `refresh_gitsavvy` when done when in threaded mode
`GsCustomCommand` actually runs said command, `refresh_gitsavvy`, in normal mode. No reason why it doesn't in threaded mode.

kaste:
Fixes #1212

There is technically no reason for a custom `Thread` class because
the default implements all features we need.

Rearrange the program so that running or not running in a separate
thread has the same features, t.i. setting a `status_message` or for
example `output_to_buffer` works in both cases now.

Especially, running `refresh_gitsavvy_interfaces` after the command
now works (#1212).

### DIFF
--- a/core/commands/custom.py
+++ b/core/commands/custom.py
@@ -2,6 +2,7 @@ import sublime
 import threading
 from sublime_plugin import WindowCommand
 
+from ..runtime import enqueue_on_worker
 from ..git_command import GitCommand
 from ...common import util
 from ..ui_mixins.input_panel import show_single_line_input_panel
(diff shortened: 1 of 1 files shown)


core/commands/custom.py
```diff
@@ -68,22 +57,26 @@ def run_async(self,
...
+                    "nuke_cursors": True,
+                })
+                if syntax:
+                    view.set_syntax_file(syntax)
+
+            util.view.refresh_gitsavvy_interfaces(self.window)
 
-        util.view.refresh_gitsavvy(self.window.active_view())
+        if run_in_thread:
+            threading.Thread(target=program, daemon=True).start()
```
asfaltboy: Question: why don't we use the `savvy_executor` via [`enqueue_on_savvy` or `run_on_new_thread`](https://github.com/timbrel/GitSavvy/blob/38dd8c5c10bd3616d5f57f1dd4296601dee53c8b/core/runtime.py#L38-L45) instead of `threading.Thread` directly here?

kaste: I wanted but (after reading the PR) `daemon=True` is a requirement here. (The example was: "It is good that Git GUI does *NOT* close automatically when Sublime exits."

asfaltboy:
Tried out a few of the custom commands I have defined (with and without threads), the core feature looks to be intact and I cannot see any problems.

Issue #1289, opened 2020-04-01
[GRAPH view] fails with unrecognized argument error


peteratemarsys:
Hi,

When I'm on the status view and try to switch to the graph view, I get the following console and error message:

```
$ git log --graph --decorate --date=format:%Y-%m-%d %H:%M --pretty=format:%h%d %<|(80,trunc)%s | %ad, %an --decorate-refs-exclude=refs/remotes/origin/HEAD --exclude=refs/stash --all --simplify-by-decoration --sparse

fatal: unrecognized argument: --decorate-refs-exclude=refs/remotes/origin/HEAD
```
gitsavvy version 2.22.0
Recently upgraded to Mac OS Catalina

Many thanks

kaste:
Updating git should do it. Which version have you installed?

Don't remember which version is the minimum required since it's been already a while since I coded this. Have to look it up again.

peteratemarsys:
Yes it did, thank you. I was on git `v2.10.x`. It works fine on `v2.26.0`.

kaste:
They added that in v 2.16 (Jan 2018). I misread that because I actually roughly wanted to be compatible with the last 4 years of git (circa v2.10). Don't know if I should revert this arg, since we already pushed, and updating git from time to time is not a bad thing.

Issue #1287, opened 2020-04-01
log view disappeared


dlnsk:
This error is in console when I go to Log view:
```
git log --graph --decorate --date=format:%Y-%m-%d %H:%M --pretty=format:%h%d %<|(80,trunc)%s | %ad, %an --decorate-refs-exclude=refs/remotes/origin/HEAD --exclude=refs/stash --all --simplify-by-decoration --sparse
fatal: unidentified argument: --decorate-refs-exclude=refs/remotes/origin/HEAD

```
Yes. `refs/remotes/origin/HEAD` doesn't exist in my repository. But I made `git remote set-head origin -a` and it has created but nothing to change.

kaste:
Has this been resolved since you opened the other issue? The error message looks like you're using an old version of git but I would have to look it up.

kaste:
Just checked again `--decorate-refs-exclude` can take any string t.i. it doesn't have to reference a valid, known `ref`.

dlnsk:
git version 2.7.4

dlnsk:
Sorry. Repository has old version of git.

PR #1291, opened 2020-04-01
Bump required git version to "2.16.0"

kaste:
#1287 and #1289 revealed that we require "2.16.0" since GitSavvy
"2.22.0". Make this hidden requirement explicit so that users know
what to do.

Note: The version jump was actually not intended, I misread the release notes. Then we could also revert some introduced changes. But just clearly bumping gives us the advantage of knowing a good minimum version as the new base. Clearly not all parts were 1.9.0 before. Having "2.16.0" makes it possible to use basically all known git features.

### DIFF
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -47,12 +47,8 @@ FALLBACK_PARSE_ERROR_MSG = (
     "operation has been aborted."
 )
 
+MIN_GIT_VERSION = (2, 16, 0)
 GIT_TOO_OLD_MSG = "Your Git version is too old. GitSavvy requires {:d}.{:d}.{:d} or above."
-
-# git minimum requirement
-GIT_REQUIRE_MAJOR = 1
-GIT_REQUIRE_MINOR = 9
-GIT_REQUIRE_PATCH = 0
 
 
 class LoggingProcessWrapper(object):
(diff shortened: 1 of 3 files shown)


kaste:
Lol, travis Linux has "2.15.x" installed, just so close; all other environments have recent enough git clients.

How to update git on travis?

core/git_command.py
```diff
@@ -346,16 +342,9 @@ def git_binary_path(self):
...
-                patch = int(match.group(3))
-                if major < GIT_REQUIRE_MAJOR \
-                        or (major == GIT_REQUIRE_MAJOR and minor < GIT_REQUIRE_MINOR) \
-                        or (major == GIT_REQUIRE_MAJOR and minor == GIT_REQUIRE_MINOR and patch < GIT_REQUIRE_PATCH):
-                    msg = GIT_TOO_OLD_MSG.format(
-                        GIT_REQUIRE_MAJOR,
-                        GIT_REQUIRE_MINOR,
-                        GIT_REQUIRE_PATCH)
+                version = tuple(map(int, match.groups()))
+                if version < MIN_GIT_VERSION:
```
asfaltboy, line 346: Could be useful to print out the found version to console here?

asfaltboy:
I think our Travis build uses Ubuntu trusty, which ships with an even older git version. We can add a step in our test to print out the git version available in the env (e.g `git --version`)

To use a later version of Git, we can [use the Bionic image][1] ([ships with 2.17.0][0]), but I have a recollection/feeling that would break availability of Python 3.3 (not sure if we care at this stage).

[0]: https://launchpad.net/ubuntu/bionic/+source/git
[1]: https://docs.travis-ci.com/user/reference/bionic/

core/git_command.py
kaste: Not enough fantasy right now for the correct wording, and how to ensure that this logging info is only printed once.

I think, it should show the the used `git_path` with its version. (Also maybe print on startup the used GitSavvy version. SublimeLinter does that.)

kaste:
Merge and release quickly because otherwise all users already updated.

PR #1292, opened 2020-04-02
Dynamic commit view

### LINKED ISSUE #980: Prevent opening commit window when there is nothing to commit

### LINKED ISSUE #1240: [commit view] Update diff when stage/index changes
Within the current system, this would typically run a refresh `on_activate` like all other views.

kaste:
Fixes #980 
Fixes #1240 

- Refresh diff on `activated`
- Show "Nothing to commit" if there is nothing to commit

### DIFF
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -137,28 +137,29 @@ class gs_prepare_commit_refresh_diff(TextCommand, GitCommand):
             show_commit_diff == "stat"
             or (show_commit_diff == "full" and self.savvy_settings.get("show_diffstat"))
         )
-        if shows_diff or shows_stat:
-            diff_text = "\n"
-            diff_text += self.git(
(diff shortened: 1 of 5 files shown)


kaste:
The plan is to ship this small iteration of the commit view (t.i. this PR and the bigger #1284) as a small patch (2.22.2) to the user rather quickly. If anybody is listening, please test shortly before we push. It just fixes too many small bugs.

PR #1297, opened 2020-04-11
Replace "replace view text" command

### LINKED ISSUE #1194: Viewport shifts to the right after replacing the whole view content
E.g.

SOTIDruhPy

Here I try the word diff cycling command. But since this is a side-effect of the `view.replace` function within `GsReplaceViewTextCommand` *all* views are actually affected. 

How to reproduce: difficult to tell exactly. 1. You need enough content so you have both scroll bars, 2. Viewport must be scrolled down t.i. not in `(0, 0)` position. 3. Cursor also down.

See https://github.com/SublimeTextIssues/Core/issues/2560 for the upstream issue description.

### LINKED ISSUE #1196: Replacing the view content shrinks selections
E.g.

IaoFNjL7ox

In the graph view or the diff view, the selection shrinks after the activate event.

kaste:
A rather boring refactoring PR. We apply the `replace_view_content` function all over the place. This function has been introduced by the rewrite (or rather the diff rendering implementation) of the `log_graph`. 

Some drive-by refactorings has been made in some modules.

Fixes #1194 
Fixes #1196

### DIFF
--- a/common/commands/debug.py
+++ b/common/commands/debug.py
@@ -9,9 +9,6 @@ from ...core.settings import GitSavvySettings
 from ...core.view import replace_view_content
 
 
-REPORT_URL_TEMPLATE = "https://github.com/timbrel/GitSavvy/issues/new?{q}"
-
-
 class GsReloadModulesDebug(WindowCommand):
 
     """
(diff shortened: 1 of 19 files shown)


kaste:
image 👀 Yeah, some shrinkage.

PR #1306, opened 2020-05-21
Iterate inline diff

kaste:
~~On top of #1305 for testing convenience.~~

It is recommended to bind `gs_inline_diff` to a key combo so it becomes a main entry point into the GitSavvy world. E.g.

```
  {
      "keys": ["ctrl+shift+["],
      "command": "gs_inline_diff",
  },
```

Now, this key acts like a toggle. You press it to switch to the inline view. Press it again to close it. Extra care has been taken that the viewport doesn't jump around while doing that. 

Being in that view, you can stage, undo staging, or discard changes. Use `[TAB]` to switch between the staged and unstaged mode. 

Use `[c]`, `[C]` or `[m]` to enter the commit message view.

Notable changes:

- Uses the builtin color scheme (just like the intra line colorization for the normal diff views).  You can compare toggling the built in "Show modifications" feat, (iirc `ctrl+shift+.` by default) with calling `gs_inline_diff` and notice that both look very similar, coherent now. 

- Fix: Show error for conflicting files
- Fix: Don't select a hunk if exactly on the line above a hunk
- Simplify how we find the right encoding being used. Remove "silent_fallback" setting by doing so.
- Fix: Respect original line endings. 
- Fix: Open file at cursor when in cached mode
- Fix: Don't throw when user "[h]unks" not being on a hunk
- Adds `[m]` to the normal diff view for consistency

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -41,6 +41,32 @@
         ]
     },
     {
+        "keys": ["c"],
+        "command": "gs_commit",
+        "context": [
+            { "key": "setting.command_mode", "operator": "equal", "operand": false },
+            { "key": "setting.git_savvy.inline_diff_view", "operator": "equal", "operand": true }
+        ]
+    },
+    {
+        "keys": ["C"],
+        "command": "gs_commit",
+        "args": {"include_unstaged": true},
+        "context": [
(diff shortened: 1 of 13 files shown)


kaste:
Lint errors are introduced by a new flake8 version. I did not touch these error locations. 🤷

PR #1307, opened 2020-06-02
set upstream after pulling a PR

randy3k:
It is useful to set the upstream to the repo of the PR author because github lets repo admin push to the PR branch.

### DIFF
--- a/github/commands/pull_request.py
+++ b/github/commands/pull_request.py
@@ -121,8 +121,6 @@ class GsGithubPullRequestCommand(WindowCommand, GitCommand, git_mixins.GithubRem
 
         url = self.best_remote_url_for_pr()
         ref = self.pr["head"]["ref"]
-        self.git("fetch", url, ref)
-        self.git("branch", branch_name, "FETCH_HEAD")
 
         owner = self.pr["head"]["repo"]["owner"]["login"]
         if owner == self.base_remote.owner:
@@ -136,7 +134,12 @@ class GsGithubPullRequestCommand(WindowCommand, GitCommand, git_mixins.GithubRem
         if set_upstream:
             if owner not in self.remotes.keys():
                 self.git("remote", "add", owner, url)
+            self.git("fetch", owner, ref)
+            self.git("branch", branch_name, "FETCH_HEAD")
             self.git("branch", "-u", remote_ref, branch_name)
+        else:
+            self.git("fetch", url, ref)
+            self.git("branch", branch_name, "FETCH_HEAD")
 
         if checkout:
             self.checkout_ref(branch_name)
--- a/github/commands/pull_request.py
+++ b/github/commands/pull_request.py
@@ -123,8 +123,6 @@ class GsGithubPullRequestCommand(WindowCommand, GitCommand, git_mixins.GithubRem
         ref = self.pr["head"]["ref"]
         self.git("fetch", url, ref)
         self.git("branch", branch_name, "FETCH_HEAD")
-        if checkout:
-            self.checkout_ref(branch_name)
 
         owner = self.pr["head"]["repo"]["owner"]["login"]
         if owner == self.base_remote.owner:
@@ -139,6 +137,9 @@ class GsGithubPullRequestCommand(WindowCommand, GitCommand, git_mixins.GithubRem
             if owner not in self.remotes.keys():
                 self.git("remote", "add", owner, url)
             self.git("branch", "-u", remote_ref, branch_name)
+
+        if checkout:
+            self.checkout_ref(branch_name)
 
         util.view.refresh_gitsavvy_interfaces(self.window, refresh_sidebar=True)
 
--- a/github/commands/pull_request.py
+++ b/github/commands/pull_request.py
@@ -31,7 +31,7 @@ class GsGithubPullRequestCommand(WindowCommand, GitCommand, git_mixins.GithubRem
 
     def run_async(self):
         self.remotes = self.get_remotes()
-        self.base_remote_name = self.get_integrated_remote_name()
+        self.base_remote_name = self.get_integrated_remote_name(self.remotes)
         self.base_remote_url = self.remotes[self.base_remote_name]
         self.base_remote = github.parse_remote(self.base_remote_url)
         self.pull_requests = github.get_pull_requests(self.base_remote)
--- a/github/commands/pull_request.py
+++ b/github/commands/pull_request.py
@@ -30,8 +30,11 @@ class GsGithubPullRequestCommand(WindowCommand, GitCommand, git_mixins.GithubRem
         sublime.set_timeout_async(self.run_async, 0)
 
     def run_async(self):
-        base_remote = github.parse_remote(self.get_integrated_remote_url())
-        self.pull_requests = github.get_pull_requests(base_remote)
+        self.remotes = self.get_remotes()
+        self.base_remote_name = self.get_integrated_remote_name()
+        self.base_remote_url = self.remotes[self.base_remote_name]
+        self.base_remote = github.parse_remote(self.base_remote_url)
+        self.pull_requests = github.get_pull_requests(self.base_remote)
 
         pp = show_paginated_panel(
             self.pull_requests,
@@ -64,7 +67,7 @@ class GsGithubPullRequestCommand(WindowCommand, GitCommand, git_mixins.GithubRem
         self.window.show_quick_panel(
             ["Checkout as detached HEAD.",
              "Checkout as local branch.",
-             "Create local branch, but do not checkout.",
+             "Create local branch without checking out.",
              "View diff.",
              "Open in browser."],
             self.on_select_action
@@ -74,59 +77,95 @@ class GsGithubPullRequestCommand(WindowCommand, GitCommand, git_mixins.GithubRem
         if idx == -1:
             return
 
+        owner = self.pr["head"]["repo"]["owner"]["login"]
+
+        if owner == self.base_remote.owner:
+            # don't add prefix for integrated remote
+            branch_name = self.pr["head"]["ref"]
+        else:
+            branch_name = "{}-{}".format(owner, self.pr["head"]["ref"])
+
         if idx == 0:
-            self.fetch_and_checkout_pr()
+            self.checkout_detached()
         elif idx == 1:
             show_single_line_input_panel(
                 "Enter branch name for PR {}:".format(self.pr["number"]),
-                "{}/{}".format(self.pr["user"]["login"], self.pr["head"]["ref"]),
-                self.fetch_and_checkout_pr
+                branch_name,
+                lambda branch_name: self.create_branch_for_pr(branch_name, checkout=True)
             )
         elif idx == 2:
             show_single_line_input_panel(
                 "Enter branch name for PR {}:".format(self.pr["number"]),
-                "{}/{}".format(self.pr["user"]["login"], self.pr["head"]["ref"]),
-                self.create_branch_for_pr
+                branch_name,
+                lambda branch_name: self.create_branch_for_pr(branch_name, checkout=False)
             )
         elif idx == 3:
             self.view_diff_for_pr()
         elif idx == 4:
             self.open_pr_in_browser()
 
-    def fetch_and_checkout_pr(self, branch_name=None):
-        self.window.status_message("Fetching PR commit...")
-        self.git(
-            "fetch",
-            self.pr["head"]["repo"]["clone_url"],
-            self.pr["head"]["ref"]
-        )
-
-        if branch_name:
-            self.window.status_message("Creating local branch for PR...")
-            self.git(
-                "branch",
-                branch_name,
-                self.pr["head"]["sha"]
-            )
-
+    def checkout_detached(self):
         self.window.status_message("Checking out PR...")
-        self.checkout_ref(branch_name or self.pr["head"]["sha"])
+
+        clone_url = self.pr["head"]["repo"]["clone_url"]
+        ssh_url = self.pr["head"]["repo"]["ssh_url"]
+        ref = self.pr["head"]["ref"]
+
+        url = ssh_url if self.base_remote_url.startswith("git@") else clone_url
+        self.git("fetch", url, ref)
+        self.checkout_ref("FETCH_HEAD")
         util.view.refresh_gitsavvy_interfaces(self.window, refresh_sidebar=True)
 
-    def create_branch_for_pr(self, branch_name):
-        self.window.status_message("Fetching PR commit...")
-        self.git(
-            "fetch",
-            self.pr["head"]["repo"]["clone_url"],
-            self.pr["head"]["ref"]
-        )
+    def create_branch_for_pr(self, branch_name, checkout=False, ask_set_upstream=True):
+        if not branch_name:
+            return
 
         self.window.status_message("Creating local branch for PR...")
-        self.git(
-            "branch",
-            branch_name,
-            self.pr["head"]["sha"]
-        )
+        set_upstream = False
+
+        owner = self.pr["head"]["repo"]["owner"]["login"]
+        if owner == self.base_remote.owner:
+            owner = self.base_remote_name
+            # set upstream automatically if the pr is from integrated remote
+            ask_set_upstream = False
+            set_upstream = True
+
+        clone_url = self.pr["head"]["repo"]["clone_url"]
+        ssh_url = self.pr["head"]["repo"]["ssh_url"]
+        url = ssh_url if self.base_remote_url.startswith("git@") else clone_url
+        ref = self.pr["head"]["ref"]
+
+        if ask_set_upstream:
+            set_upstream = sublime.ok_cancel_dialog(
+                "Set upstream to '{}/{}'?".format(owner, ref))
+
+        if set_upstream:
+            if owner not in self.remotes.keys():
+                self.git("remote", "add", owner, url)
+
+            self.create_branch_from_remote_for_pr(branch_name, owner, ref, checkout)
+        else:
+            self.create_branch_from_sha_for_pr(branch_name, url, checkout)
+
+    def create_branch_from_remote_for_pr(self, branch_name, remote, remote_branch, checkout):
+        self.git("fetch", remote, remote_branch)
+        ref = "{}/{}".format(remote, remote_branch)
+        self.git("branch", branch_name, ref)
+        self.git("branch", "-u", ref, branch_name)
+
+        if checkout:
+            self.checkout_ref(branch_name)
+
+        util.view.refresh_gitsavvy_interfaces(self.window, refresh_sidebar=True)
+
+    def create_branch_from_sha_for_pr(self, branch_name, remote_url, checkout):
+        self.git("fetch", remote_url, self.pr["head"]["ref"])
+        self.git("branch", branch_name, "FETCH_HEAD")
+
+        if checkout:
+            self.checkout_ref(branch_name)
+
+        util.view.refresh_gitsavvy_interfaces(self.window, refresh_sidebar=True)
 
     def view_diff_for_pr(self):
         response = interwebs.get_url(self.pr["diff_url"])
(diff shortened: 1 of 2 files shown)


github/commands/pull_request.py
```diff
@@ -79,13 +79,13 @@ def on_select_action(self, idx):
         elif idx == 1:
             show_single_line_input_panel(
                 "Enter branch name for PR {}:".format(self.pr["number"]),
-                "{}/{}".format(self.pr["user"]["login"], self.pr["head"]["ref"]),
```
randy3k, line 82: also change the default value of branch name from owner/repo to owner-repo to avoid ambiguity.

github/commands/pull_request.py
```diff
@@ -101,17 +101,20 @@ def fetch_and_checkout_pr(self, branch_name=None):
...
         )
 
-        if branch_name:
-            self.window.status_message("Creating local branch for PR...")
-            self.git(
-                "branch",
-                branch_name,
-                self.pr["head"]["sha"]
-            )
+        if not branch_name:
```
kaste, line 104: This change is not ok since it breaks the "Checkout detached" feature. I mean `branch_name` can be `None` intentionally.

github/commands/pull_request.py
```diff
@@ -101,17 +101,20 @@ def fetch_and_checkout_pr(self, branch_name=None):
...
+            "branch",
+            branch_name,
+            self.pr["head"]["sha"]
+        )
 
         self.window.status_message("Checking out PR...")
-        self.checkout_ref(branch_name or self.pr["head"]["sha"])
+        self.checkout_ref(branch_name)
         util.view.refresh_gitsavvy_interfaces(self.window, refresh_sidebar=True)
+        sublime.set_timeout(lambda: self.set_upstream_for_branch(branch_name))
```
kaste, line 117: Avoid `set_timeout` here. Why is that needed. Also call `refresh_gitsavvy` at the end.

github/commands/pull_request.py
```diff
@@ -121,12 +124,48 @@ def create_branch_for_pr(self, branch_name):
             self.pr["head"]["ref"]
         )
 
+        if not branch_name:
+            branch_name = self.pr["head"]["sha"]
+
```
kaste, line 129: We did not catch `not branch_name` before but here in that case the user did not enter a branch name, and we should probably return/abort the operation.

github/commands/pull_request.py
```diff
@@ -121,12 +124,48 @@ def create_branch_for_pr(self, branch_name):
...
         self.window.status_message("Creating local branch for PR...")
         self.git(
             "branch",
             branch_name,
             self.pr["head"]["sha"]
         )
+        sublime.set_timeout(lambda: self.set_upstream_for_branch(branch_name))
```
kaste, line 136: Avoid `set_timeout`.

github/commands/pull_request.py
```diff
@@ -121,12 +124,48 @@ def create_branch_for_pr(self, branch_name):
...
+                elif index == 0:
+                    url = clone_url
+                elif index == 1:
+                    url = ssh_url
+
+                self.git("remote", "add", remote, url)
+                self.git("fetch", remote, remote_branch)
+                self.git("branch", "-u", "{}/{}".format(remote, remote_branch), branch_name)
+
+            sublime.set_timeout(
```
kaste, line 162: Avoid `set_timeout`

github/commands/pull_request.py
```diff
@@ -121,12 +124,48 @@ def create_branch_for_pr(self, branch_name):
...
+                lambda: self.window.show_quick_panel(
+                    [clone_url, ssh_url], on_select_url)
```
kaste, line 164: I don't know what `clone_url` vs. `ssh_url` actually represent. I will check that on the functional tests later.

kaste: That being said, Github recommends `https` urls these days because these are accessible and just work in most environments.

github/commands/pull_request.py
```diff
@@ -121,12 +124,48 @@ def create_branch_for_pr(self, branch_name):
...
+            )
+        else:
+            self.git("fetch", remote, remote_branch)
```
kaste, line 167: Here and above L159: we actually already fetched at the beginning of this action. 

```
        self.git(
            "fetch",
            self.pr["head"]["repo"]["clone_url"],
            self.pr["head"]["ref"]
        )
```

Why is that needed? Is there any difference? Not if the user selects `clone_url`, if the user selects `ssh` it just uses a different transport protocol, the ref is the same, and we already have it. No?

github/commands/pull_request.py
randy3k: `clone_url` is like "https://github.com/octocat/Hello-World.git" and `ssh_url` is like "git@github.com:octocat/Hello-World.git"

github/commands/pull_request.py
randy3k: it is needed for "Checkout as detached HEAD." where we do need to set upstream. Perhaps we should prompt user to choose what protocol to fetch.

randy3k: Now we always prompt user to choose a protocol when checkout as detached head.

kaste: Why is the protocol a matter here? It never was before. As I said these days using https is just fine. Is there an issue open requesting and explaining the need? If this is a real issue we should deduce: if the user already used only non https for the remotes, we do it here as well. Otherwise we use https.

It confuses me to set an upstream if the user chooses detached head. That doesn't make any sense to me. We could ask to add the remote, but iirc the user can just checkout the same PR as a branch later if she needs to add commits. I mean, you only need the remote and upstream if you want to edit the PR, and that doesn't play well with a detached HEAD.

github/commands/pull_request.py
```diff
@@ -94,39 +94,84 @@ def on_select_action(self, idx):
...
+
+        remotes = list(self.get_remotes().keys())
+        remote = self.pr["user"]["login"]
+        remote_branch = self.pr["head"]["ref"]
+
+        clone_url = self.pr["head"]["repo"]["clone_url"]
+        ssh_url = self.pr["head"]["repo"]["ssh_url"]
+
+        if remote not in remotes:
+            if not sublime.ok_cancel_dialog("Add remote '{}'?".format(remote)):
```
kaste, line 144: Now you cannot checkout a PR without adding the remote. Is this intended?

github/commands/pull_request.py
kaste: Ah you already changed that and don't set the upstream in the detached case. So the only question is why the protocol matters now.

randy3k: In Linux, I think it is more common/easier to use ssh key rather than `credential.helper`. 

Also, the https protocol is particularly painful if two-factor authentication is used.

github/commands/pull_request.py
randy3k: I don't know if it is a good default. Do you think we should proceed without setting upstream?

randy3k: Or we modify the behavior, `Checkout as local branch` sets upstream and `Create local branch` does not.

github/commands/pull_request.py
asfaltboy: I feel users should be allowed to choose, regardless of github recommendation. Also, it's important that we don't break any behaviour for existing users

kaste: The user already made that choice. We can take whatever is set up before. Just look at the already configured remote urls and go with https if it's in there. I don't see why I should make this decision here over and over again. If you made the initial clone using https, just stick with it.

randy3k: If the remote exists, you won't be prompted to choose a URL. However, user will be asked if they are fetching the PR as detached head or the remote doesn't exist.

kaste: The remote of the main repo determines which protocol I want to use and have set up. Since we're checking out a fork here there must be at least the main repo already configured.

randy3k: Perhaps the easier solution is to make a setting `pull_request_protocol` with default value `"ask"` and possible values "ssh" and "https".

randy3k:
By the way, we also need to port the code to `GsGitlabMergeRequestCommand`. Or even better, make the code more CI agnostic.

github/commands/pull_request.py
kaste: No that's the most complicated solution. Doesn't `self.get_integrated_remote_url()` already gives us the current base remote url, we query for this at the beginning of the process, and so we just look if `base_remote.starts_with("https:")`.

randy3k: Alright. It makes sense. I believe that most people don’t need to switch between protocol. I have to go sleeping now, will do the changes tomorrow.

kaste:
Note that the flake8 errors are fixed in dev now, you should drop this commit here.

github/commands/pull_request.py
```diff
@@ -74,59 +76,113 @@ def on_select_action(self, idx):
...
+        owner = self.pr["head"]["repo"]["owner"]["login"]
+
+        if owner == self.base_remote.owner:
+            # don't add prefix for integrated remote
+            branch_name = self.pr["head"]["ref"]
+        else:
+            branch_name = "{}-{}".format(owner, self.pr["head"]["ref"])
+
         if idx == 0:
             self.fetch_and_checkout_pr()
```
kaste, line 88:
```suggestion
            self.checkout_detached()
```

github/commands/pull_request.py
```diff
@@ -74,59 +76,113 @@ def on_select_action(self, idx):
...
         elif idx == 1:
             show_single_line_input_panel(
                 "Enter branch name for PR {}:".format(self.pr["number"]),
-                "{}/{}".format(self.pr["user"]["login"], self.pr["head"]["ref"]),
+                branch_name,
                 self.fetch_and_checkout_pr
```
kaste, line 93:
```suggestion
                lambda branch_name: self.create_branch_for_pr(branch_name, checkout=True)
```

github/commands/pull_request.py
```diff
@@ -74,59 +76,113 @@ def on_select_action(self, idx):
...
             )
         elif idx == 2:
             show_single_line_input_panel(
                 "Enter branch name for PR {}:".format(self.pr["number"]),
-                "{}/{}".format(self.pr["user"]["login"], self.pr["head"]["ref"]),
-                self.create_branch_for_pr
+                branch_name,
+                lambda x: self.create_branch_for_pr(x, ask_set_upstream=True)
```
kaste, line 99:
```suggestion
                lambda branch_name: self.create_branch_for_pr(branch_name, checkout=True)
```

github/commands/pull_request.py
```diff
@@ -74,59 +76,113 @@ def on_select_action(self, idx):
...
         util.view.refresh_gitsavvy_interfaces(self.window, refresh_sidebar=True)
 
-    def create_branch_for_pr(self, branch_name):
-        self.window.status_message("Fetching PR commit...")
-        self.git(
-            "fetch",
-            self.pr["head"]["repo"]["clone_url"],
-            self.pr["head"]["ref"]
-        )
+    def create_branch_for_pr(self, branch_name, checkout=False, ask_set_upstream=False):
```
kaste, line 130:
```suggestion
    def create_branch_for_pr(self, branch_name, checkout=False, ask_set_upstream=True):
```

github/commands/pull_request.py
```diff
@@ -74,59 +76,113 @@ def on_select_action(self, idx):
...
         if branch_name:
-            self.window.status_message("Creating local branch for PR...")
-            self.git(
-                "branch",
-                branch_name,
-                self.pr["head"]["sha"]
-            )
+            self.create_branch_for_pr(branch_name, checkout=True, ask_set_upstream=True)
+        else:  # detached HEAD
+            self.checkout_detached()
```
kaste, line 110: Remove method; it also has a bug (or is it a feature?) that if the user did not enter a branch name we continue with `checkout_detached` instead of aborting. If it's a feature we should have it of course for case 1 *and* 2 t.i. on L132.

github/commands/pull_request.py
```diff
@@ -74,59 +76,113 @@ def on_select_action(self, idx):
...
+        clone_url = self.pr["head"]["repo"]["clone_url"]
+        ssh_url = self.pr["head"]["repo"]["ssh_url"]
+        ref = self.pr["head"]["ref"]
+
+        if self.base_remote_url.startswith("https://"):
+            url = clone_url
+        elif self.base_remote_url.startswith("git@"):
+            url = ssh_url
+        else:
+            raise RuntimeError("Cannot recognize url {}.".format(self.base_remote_url))
```
kaste, line 124: Simplify and be open, e.g. 

```
url = ssh_url if self.base_remote_url.startswith("git@") else clone_url
```

github/commands/pull_request.py
```diff
@@ -74,59 +76,113 @@ def on_select_action(self, idx):
...
+                "Set upstream to '{}/{}'?".format(owner, ref))
+
+        if set_upstream:
+            if owner not in remotes.keys():
+                if self.base_remote_url.startswith("https://"):
+                    url = clone_url
+                elif self.base_remote_url.startswith("git@"):
+                    url = ssh_url
+                else:
+                    raise RuntimeError("Cannot recognize url {}.".format(self.base_remote_url))
```
kaste, line 160: Same as above, usually be open

```
url = ssh_url if self.base_remote_url.startswith("git@") else clone_url
```

github/commands/pull_request.py
```diff
@@ -30,8 +30,10 @@ def run(self):
         sublime.set_timeout_async(self.run_async, 0)
 
     def run_async(self):
-        base_remote = github.parse_remote(self.get_integrated_remote_url())
-        self.pull_requests = github.get_pull_requests(base_remote)
+        self.base_remote_name = self.get_integrated_remote_name()
```
kaste, line 33: `self.get_integrated_remote_name()` is costly as it does at min 2 git calls, but the result is only used in a conditional once. Note that `get_integrated_remote_url` calls this internally.
```suggestion
```

github/commands/pull_request.py
```diff
@@ -74,59 +76,113 @@ def on_select_action(self, idx):
...
-            "branch",
-            branch_name,
-            self.pr["head"]["sha"]
-        )
+        set_upstream = False
+
+        remotes = self.get_remotes()
+        owner = self.pr["head"]["repo"]["owner"]["login"]
+        if owner == self.base_remote.owner:
+            owner = self.base_remote_name
```
kaste, line 140:
```suggestion
            owner = self.get_integrated_remote_name()
```

Can you explain what this does. First, owner is `self.pr["head"]["repo"]["owner"]["login"]` which is probably a GitHub username? Here we change the `owner` to `base_remote_name`, which is the local remote name(?), so we have for example `owner == origin` here?

github/commands/pull_request.py
```diff
@@ -74,59 +76,113 @@ def on_select_action(self, idx):
...
+        self.git("branch", branch_name, ref)
+        self.git("branch", "-u", ref, branch_name)
+
+        if checkout:
+            self.checkout_ref(branch_name)
+
+        util.view.refresh_gitsavvy_interfaces(self.window, refresh_sidebar=True)
+
+    def create_branch_from_sha_for_pr(self, branch_name, checkout):
+        self.git("branch", branch_name, self.pr["head"]["sha"])
```
kaste, line 180: I think you forgot the "fetch" here.

Generally I don't understand why `create_branch_from_remote_for_pr` and `create_branch_from_sha_for_pr` are here, and especially why they differ so much.  From the code flow you would say they both should do the same, except one variant *also* sets upstream which roughly are one or two git calls. 

So they should be the same code, except:

```
self.git("remote", "add", owner, url)  # if the remote is new
self.git("branch", "-u", ref, branch_name)  # set actually new side-effect, setting the upstream
```

Can't we do 

```
self.git("fetch", url, self.pr["head"]["ref"])
self.git("branch", branch_name, self.pr["head"]["sha"])
if set_upstream:
	self.git("remote", "add", remote, url)
	self.git("branch", "-u", "{}/{}".format(remote, remote_branch), branch_name)
if checkout:
	self.checkout_ref(branch_name)
```

github/commands/pull_request.py
randy3k: `self.pr["head"]["repo"]["owner"]["login"]` could be an github user/organization name. For example, it could be `timbrel` if a PR is coming from a branch of `GitSavvy`. It is checking if the `base_remote` owner is the same as the PR's owner. 

In most cases, `self.base_remote.owner` would be the username of the PR submitter and `self.base_remote_name` would be `"origin".`

github/commands/pull_request.py
randy3k: In order to set up stream
```
self.git("branch", "-u", "{}/{}".format(remote, remote_branch), branch_name)
```

We need to make sure that the remote branch is fetched from the remote.

```
        self.git("fetch", remote, remote_branch)
        ref = "{}/{}".format(remote, remote_branch)
        self.git("branch", branch_name, ref)
        self.git("branch", "-u", ref, branch_name)
```
and it cannot be simply
```
self.git("fetch", url, self.pr["head"]["ref"])
``

File: github/commands/pull_request.py [hunk 20]
```diff
@@ -74,59 +77,95 @@ def on_select_action(self, idx):
...
+        self.git("branch", branch_name, ref)
+        self.git("branch", "-u", ref, branch_name)
+
+        if checkout:
+            self.checkout_ref(branch_name)
+
+        util.view.refresh_gitsavvy_interfaces(self.window, refresh_sidebar=True)
+
+    def create_branch_from_sha_for_pr(self, branch_name, remote, checkout):
+        self.git("fetch", remote, self.pr["head"]["sha"])
```
kaste, line 162: Honestly, I feel stupid, but for this branch of the code, t.i. we don't set upstream, we can't assume we have a remote set. Can we? We need to fetch from the URL I would have thought.

Also why Sha and not ref?

github/commands/pull_request.py
kaste: So there is a difference between git fetch URL ref and git fetch remote which points to URL ref.

randy3k: Yes. 

```
self.git("fetch", url, self.pr["head"]["ref"])
```
would fetch the ref to `FETCH_HEAD` instead.

github/commands/pull_request.py
randy3k: Oh, you are right.

github/commands/pull_request.py
kaste: Haha, after 53 comments you introduce a new concept "FETCH_HEAD". 

> would fetch the ref to FETCH_HEAD instead

Yeah, okay, but then `git branch branch_name FETCH_HEAD`. I argue that you only need the `remote/remote_branch` for the `branch -u` call. It may not matter, I would like the least moving parts/variants as possible for readability. 

Otherwise, I think it's okay now. 

BTW. Have you seen "View diff" is broken. Github changed its server configuration and now sends lowercase headers.  Basically we look for "Location: ..." and they send "location: ...". Or "content-type"...

randy3k: > I argue that you only need the remote/remote_branch for the branch -u call

I am not sure if I understand.

> BTW. Have you seen "View diff" is broken. 

Maybe, I never used this option.

kaste:
Can you take a look on the refactoring.  Now a typical "checkout as branch" looks like this:

```
 (D) [ 94ms] $ git remote -v
 (D) [ 78ms] $ git config --local --get GitSavvy.ghRemote
 (M) [784ms] $ git fetch https://github.com/timbrel/GitSavvy.git review
 (M) [ 78ms] $ git branch review2 FETCH_HEAD
 (M) [ 94ms] $ git checkout review2
 (M) [ 78ms] $ git branch -u origin/review review2
```

I don't see why this shouldn't work. Is this stupid? Note that I reordered here, we fetch and checkout immediately, and ask to set the upstream after. T.i. we don't ask if the previous commands failed.

The second commit is to remove one further `get_remotes()` call.

randy3k:
We will need to fetch from remote explicitly after `remote add`.

kaste:
Oh, fuck, you're right 🙄  but really I only understand and see this now you wrote it and every git command is aligned, t.i. after the refactoring. Simple code line by line.

randy3k:
LGTM. Thank for reviewing it.

kaste:
🔥

kaste:
Love the feature

PR #1311, opened 2020-06-12
missing force argument

randy3k:
The `force` argument was missing.


kaste:
Yeah, fair. At least it asked you to delete anyway.  (I probably wanted to deprecate the `D` binding. But since then I always delete branches in the graph view.)

randy3k:
How did you delete a branch in graph view? I can't find such an option.

kaste:
`[enter]`

image

randy3k:
Oh, I must be blinded... 🤦

PR #1324, opened 2020-07-13
Implement smart next/prev hunk commands

kaste:
With the Sublime gutter markers, they also introduced the commands
"next_modification" and "prev_modification".  Unfortunately, they
jump really between single, recognized edits, leading to frustrating
fine-grained jumps.  On the other side, GitGutter always jumped from
hunk to hunk but is slightly discouraged becuase of the - as it seems -
unnecessary overhead.

We introduce the commands "gs_next_hunk" and "gs_prev_hunk" which jump
hunk per hunk *using* Sublime's basically instant, and computational
cheap markers.

They can be used from normal views; for this PR we bind it for standard 
hunk-per-hunk navigation to the "File Revision View" on `[.,]`.

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -787,6 +787,22 @@
     ////////////////
 
     {
+        "keys": ["."],
+        "command": "gs_next_hunk",
+        "context": [
+            { "key": "setting.command_mode", "operator": "equal", "operand": false },
+            { "key": "setting.git_savvy.show_file_at_commit_view", "operator": "equal", "operand": true }
+        ]
+    },
+    {
+        "keys": [","],
+        "command": "gs_prev_hunk",
+        "context": [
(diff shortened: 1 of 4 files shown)


kaste:
Either `gs_next_hunk` (and always the "prev" counterpart of course) itself or yet another command maybe `gs_next_modification` (based on Sublime's naming) could be expanded to a generalized nav command.  The use case would be that users could simply bind this one command to some global shortcut, for example `ctrl+.` which then could be used (unified) in normal views and our diff/inline diff views, and the commit and file revision views, and so on. 

As Sublime iirc also suggest something like `ctrl+shift+.` to see the hunk diff inline, we could suggest a setup like the following

```
ctrl+,  prev hunk
ctrl+.  next hunk
ctrl+shift+,  inline diff
ctrl+shift+.  normal diff
```

I use a similar key-workflow for a very long time (though on a different base key) to enter the GitSavvy world.

PR #1325, opened 2020-07-14
Small graph iteration

kaste:
- Add "pull"/"push"/"fetch" to the action menu only when on the HEAD

E.g.

image

I think this will be convenient.

- Tweak syntax and mark "fixup!" and "squash!" prefixes as `string.other.rebase-hint.git-savvy`.  Just being "string" probably makes it yellow by default but you can fancy style as well

E.g.

image

This is should make it easy to spot if you still need to rebase before pushing or merging.

- Fine-tune "diff against" action: When on the head, open the normal diff views where you can actually stage etc.  Otherwise usually open a diff "commit..HEAD" (t.i. **not** workdir) which is best suited for reviewing topic branches (PRs). 

- Implement a smart copy feature to copy the commit hash, commit message, or a combination of both to the clipboard. 

E.g.

mUIMP2V6Fr

### DIFF
--- a/core/commands/__init__.py
+++ b/core/commands/__init__.py
@@ -16,6 +16,7 @@ from .init import *
 from .inline_diff import *
 from .log import *
 from .log_graph import *
+from .log_graph_smart_copy import *
 from .merge import *
 from .mv import *
 from .navigate import *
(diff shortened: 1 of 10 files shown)


kaste:
I'm not completely confident about the special "fixup" marking as "string", and the smart copy thing (esp. how we should order the cycle so that the most useful copy comes first).  But that's what we have `dev` for.

PR #1329, opened 2020-07-18
Remove "word-diff" feature

### LINKED ISSUE #1322: Remove "word-diff" feature
Although I tried last year to make the "word-diff"s more (or a bit) useful, it was actually disappointing.  This feature has now been superseded by the intra-line colorization I implemented right after that. 

The fancy word-diff is only available in the "diff" views. The commit views actually have the word diff toggle but still show the raw unreadable output as ever. 

I intend to remove the feature completely.

In addition: for the diff view, move the remaining key binding to `[w]` (away from `[s]`), so that we can use `sS` for staging.

kaste:
Fixes #1322

Removes the "word-diff" feature completely. Remaps `s` to `w` in the diff and commit view. 

As of now, no deprecation policy, no warnings.

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -643,17 +643,9 @@
         ]
     },
     {
-        "keys": ["s"],
+        "keys": ["w"],
         "command": "gs_diff_toggle_setting",
         "args": { "setting": "ignore_whitespace" },
-        "context": [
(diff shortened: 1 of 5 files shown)


kaste:
Merge for testing into dev

PR #1330, opened 2020-07-18
Small refactor inside inline diff

kaste:
As a follow up from #1327.  Already being in that "inline_diff.py" apply some basic refactorings.  Among others, again make a better distinction between "line" and "row". 

User facing changes:

- Flash a status message if the undo stack is empty.
- Maybe slightly better cursor movement after staging a single line.


kaste:
Merge for testing into dev

PR #1331, opened 2020-07-18
Implement "Line History View"

kaste:
A "Line History" (aka "log wtf" or "log why") basically calls
"git log -L<some_file>,1:3".  Rarely used on the command line because of
its syntax, it is a super easy to write and then very useful tool in
the editor.  It is usually the faster (in terms of you get the info
you're looking for faster) "blame" and also the faster "File History"
because often you're only interested in some part of a file. Really
answering "Why is this line or section of code here?".

Comes with `[o]` to open the complete commit under the cursor;
`[O]` to open the file revision;
`[g]` to open the graph context.

### DIFF
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -89,9 +89,13 @@
         "args": { "all": false }
     },
     {
-        "caption": "git: File history",
+        "caption": "git: File History",
         "command": "gs_graph_current_file",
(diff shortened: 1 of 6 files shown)


kaste:
🔥

kaste:
U5lSPIitQh

PR #1359, opened 2020-09-02
Slightly modernize css and popup placement

kaste:
Always place the popup at the top-left of the view.  Slightly modernize the styling, grouping and wording but there is room for further improvement.  Actually, we should not write the html but rather generate it from a simpler markup structure.

image

Before: 

image


kaste:
(Esp. the list-item dots seem unnecessary.)

PR #1367, opened 2020-09-16
Speedup navigation esp. in the graph view

kaste:
Navigation using for example using `up|down` was a bit laggy esp. in the
graph view.  

1. We cache `get_available_regions` in `GsNavigate` and use `view.change_count()`as the
invalidator. 

2. We implement a binary search for these available regions.  This is needed esp.
for log graph view, and especially for going up.  For going up, we used a simple `reversed`linear search which is obviously slow if you have 5000 commits below the 
cursor.  For all other views there is no advantage, and a linear search would be maybe even faster but probably not measurable. 

3. In the log graph view, we implement a cached version of `view.find_by_selector` which speeds-up looking for fixup/squash commits.

4. We generally cache following a dot.

### DIFF
--- a/core/commands/log_graph.py
+++ b/core/commands/log_graph.py
@@ -1430,12 +1430,9 @@ def colorize_fixups(view):
 def _colorize_fixups(vid, dots):
     # type: (sublime.ViewId, Tuple[colorizer.Char]) -> None
     view = sublime.View(vid)
-    message_regions = find_by_selector(view, 'meta.graph.message.git-savvy')
-    extract_message = partial(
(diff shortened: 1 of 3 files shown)


kaste:
Closing because the scope of this change is too big.  We don't *need* to change anything in `GsNavigate` just because we can and it looks like it's working and makes everything faster for free. 

I would prefer to only change the `log_graph` navigation.  Holy moly, even a 5000 commits binary search is a waste (or *getting* the 5000 commits from Sublime to be clear, not the searching) when we mostly and usually just go one row up or down.  For the other views or usages, we usually would binary search over a list of ... say 5 items. Nah.

PR #1369, opened 2020-09-17
Do not cache reading the HEAD commit

### LINKED ISSUE #1368: gs_show_commit with `HEAD` opens up the wrong commit
It used to show the last commit in current branch (i.e HEAD), and now it seems stuck to some commit down the tree... perhaps there's some caching magic that was added that identifies HEAD as a unique commit :/ .. not sure if there's a use case here for typical GitSavvy usage; but, for me, it can be nice to see a `skip_cache` sort of argument to avoid this.

kaste:
Fixes #1368

Do not cache reading the HEAD commit. Everywhere else, we already checked if the `commit_hash` is `""` or `"HEAD"` to skip caching, so this one was missing here.

We go one step further, special case "HEAD", and dynamically update and thus follow HEAD.

The use case is a user binding to `gs_show_commit` with `commit_hash`
set to `HEAD`.  For example:

```
  {
      "caption": "git: show HEAD commit",
      "command": "gs_show_commit",
      "args": {
          "commit_hash": "HEAD"
      }
  },
```

However, arbitrary references, like for example tags or branch names, are not supported.

### DIFF
--- a/core/commands/show_commit.py
+++ b/core/commands/show_commit.py
@@ -41,6 +41,9 @@ class gs_show_commit(WindowCommand, GitCommand):
     def run(self, commit_hash):
         # need to get repo_path before the new view is created.
         repo_path = self.repo_path
+        if commit_hash in {"", "HEAD"}:
+            commit_hash = self.git("rev-parse", "--short", "HEAD").strip()
+
         this_id = (
             self.repo_path,
             commit_hash
--- a/core/git_mixins/history.py
+++ b/core/git_mixins/history.py
@@ -306,6 +306,15 @@ class HistoryMixin():
         ignore_whitespace=False
     ):
         # type: (str, Optional[str], bool, bool, bool) -> str
+        if not commit_hash or commit_hash == "HEAD":
+            return self._read_commit(
+                commit_hash,
+                file_path,
+                show_diffstat,
+                show_patch,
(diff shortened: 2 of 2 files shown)


asfaltboy:
Looking good, fix works except it doesn't refresh an existing open "show commit view"

kaste:
Yeah, I noticed that we actually fall down the rabbit hole rather immediately.  ("some" features fail, LOL)

a) We resolve HEAD to a commit hash; then everything works, and usually every time you invoke the command a new view gets created *except* HEAD didn't move.  (After your commit it *always* creates a new view.)  (Resolving would mean **1** extra git call for that special case.)

b) We just refresh, basically like all other views, for example `on_activated`, if and only if the commit is HEAD of course.  (For all other views the content is cacheable and static.)  Then we had a changing view for the HEAD case, basically if you don't close it, it would follow the current branch. 

🤔

asfaltboy (approved):
I'm not sure I understand how the listener works, but I tested it and it worked fine for me 👍

kaste:
I reverted the dynamic "follow HEAD" approach and just resolve "HEAD". T.i. I implemented a) from above comment. 

I don't feel that "Show commit" should be *not* immutable right now.  If we have previous/next commit navigation here, this feeling might change.

PR #1376, opened 2020-10-02
Improve dynamic filtering the graph view

kaste:
- For `f`, maintain a history of previous filters.  Use `up|down` while in the input panel to access them.  Pre-populate the history with some common good settings, for example "--reflog" is very useful.  Also compute  a "--author" arg based on the current commits author.  

- Add `l` (**ls**-tree), to add or remove paths to the command.  This is really a dynamic variant of our "File History".  You can also select paths, not just concrete files; and of course multiple ones.

- For `F`, document the binding in the `?` tooltip 🙄.  Make it a toggle which will toggle `filters` and the file paths.  (Also `filter_by_author`, which can probably be deprecated with the new helper for `f`.)  For example, looking at a "File History" `F` will show the commit with full context.  

image

image

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -1451,6 +1451,22 @@
         ]
     },
     {
+        "keys": ["up"],
+        "command": "gs_input_handler_go_history",
+        "args": { "forward": false },
+        "context": [
+            { "key": "setting.input_panel_with_history" }
+        ]
+    },
+    {
+        "keys": ["down"],
+        "command": "gs_input_handler_go_history",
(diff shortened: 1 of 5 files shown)


kaste:
This is very nice to play with.  I tend to get more graph views open and reusing these views will be a WIP, I think.  Otherwise, the "--reflog" option in concerto with the "rebase from graph" PR is 👍.  

I also esp. like that I can turn "File History" views into "normal" ones.  Basically, I find the commit I'm interested in, and then "F" to switch off the filter to see the context of that change.  😗

Issue #1355, opened 2020-08-27
Create fixup commit from graph view


FPtje:
## Proposal

In the GRAPH view, when pressing enter on a commit, add the option "Create fixup commit from stage", which is an alias for `git commit --fixup <commit>`. 

## Explanation

GitSavvy has two fixup related commands:

- git: fixup from stage
- git: quick stage current file and fixup

Both these options change the git history, by changing the commit to include the fixup. `git`, however has another option: `git commit --fixup <commit ref>`, the description of which is as follows:

```
--fixup=<commit>
Construct a commit message for use with rebase --autosquash. The commit message will be the subject line from the specified
commit with a prefix of "fixup! ". See git-rebase(1) for details.
```

I prefer this method when working in any non-master branch, precisely because it does not change history. This allows for a "soft" form of rebasing that allows other people to cooperate without force pulling. It also plays nicer with GitHub's review function. Right before merging into master, a `git rebase --autosquash` will do the actual rebasing.

## Why this is needed

My current way of creating fixup commits is as follows:

- Stage hunks/files in GitSavvy STATUS
- Copy commit ID from GitSavvy GRAPH
- In a terminal, enter `git commit --fixup <PASTE>`
- Push the branch from GitSavvy STATUS

This is cumbersome. Being able to fixup from GRAPH would be much nicer, since I use that to judge which commit I want to fixup anyway.

On another note, perhaps it would be nice to make the `git: fixup from stage` option available in GRAPH as well?

kaste:
Yeah, that is cumbersome, and I do it often.  I actually don't fully grasp "our" fixup commands; I really just make a fixup commit and go on, and do the squashing later.  It is okay for the happy path but when the fixup does not apply and sure it doesn't apply all the time; our fixup command just aborts.  And then I'm really out of my flow. 

What I wanted to do is write a helper in the actual commit message view.  Basically you stage in a diff or the status view, then `c` to enter the commit message.  Here type `fixup<TAB>` and it will open a menu where you can select a commit; upon selection it will complete the commit subject.  (Same for squash.)

In the next release will be a "copy" helper, to copy the hash or commit message easier.  After that, I probably add some rebasing functionality into the graph. 

Your idea is also good, what I don't like in the graph is that I can't tell if I have a dirty state, or already something staged, etc.  Don't know if showing the action all the time would be okay.  (The actions menu shouldn't have too many options to not scare people.)

FPtje:
[...]

PR #1377, opened 2020-10-02
Implement "fixup" helper in the commit message view

### LINKED ISSUE #487: GitSavvy commit: Pull from history (enhancement request)
I frequently pull bits of information from previous commit entries, especially the most recent one. Can we make it easier to do this from inside a GitSavvy commit session? Right now, my workaround is to navigate to the Rebase dashboard.

Possibilities:
- List prior commit messages in the comments.
- Make a keystroke to pull in the most recent commit message.
- Make a keystroke to open a separate window with commit messages listed last first.

kaste:
In addition to the `@<tab>` and `#<tab>` helper, 

1) add `fixup<tab>` and `squash<tab>` helpers.  When invoked a custom log panel opens.  On selecting a commit, we construct a fixup (or squash) commit message subject.  The user can add some further comments after that or `ctrl+enter` to actually commit.  (T.i. we don't commit automatically.)

2) Hitting `<tab>`exatly on BOF position, t.i. usually immediately after opening the view, will also show the recent log.  Either just for reading, or, on `enter`, to use the commit subject, usually as a template for further editing so we stay at BOF position after that.

Also some small fixes and glitches around that code. 

Fixes #487 
Ref #1355

### DIFF
--- a/Default (Linux).sublime-keymap
+++ b/Default (Linux).sublime-keymap
@@ -9,7 +9,7 @@
         "command": "gs_commit_view_do_commit",
         "context": [
             { "key": "setting.command_mode", "operator": "equal", "operand": false },
-            { "key": "setting.git_savvy.get_long_text_view", "operator": "equal", "operand": true },
+            { "key": "setting.git_savvy.commit_view", "operator": "equal", "operand": true },
             { "key": "setting.git_savvy.commit_on_close", "operator": "equal", "operand": false }
         ]
     },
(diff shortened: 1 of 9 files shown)


asfaltboy (approved):
Awesome work, I will use this a lot! Tested and it worked for me, at least the happy path :)

I didn't really understand why we remove `get_long_text_view` here?

kaste:
@asfaltboy `git_savvy.get_long_text_view` is used as a marker for the keybindings which is unusual.  All other views have a specific marker `git_savvy.<view_type>_view=True`.  This change thus just follows our rules.  (`get_long_text_view` reads as if it denotes a *capability*.  But we don't use such *capabilities* we use concrete tags (or classes) like `log_graph_view` to identify specific views.)

kaste:
Generalized the approach *a bit*.  The user can now hit `<tab>` at the very beginning of the commit subject, t.i. at BOF position, to open the recent log.  On `enter` will copy the selected commit subject.  That's only useful if the subject lines contain repetitive info one wants to copy over for further editing.

PR #1379, opened 2020-10-05
Fixes stash docs references

kinduff:
Hello :wave: thanks a lot for this awesome project, and keeping it alive.

This PR fixes some documentation references that had the wrong links, they should now point to the correct place.


kaste:
Oh, the help is not in a good condition.  Thanks for helping out.

PR #1385, opened 2020-10-09
Rebase actions in the graph view

kaste:
This adds rebase functionality to the graph view. 

image

- `[r]` to open above rebase menu 
- `[W]` to reword/edit the commit message of the selected commit
- `[E]` to edit the selected commit
- `[R]` to start automatic autosquashing beginning with the selected commit  (t.i. apply all fixup and squash commits without further asking)

If the cursor is on a fixup or squash commit, the menu has an additional item "Apply fix" which just applies the selected fixup. 

What's not in:

- [ ] Create fixup commit for the selected commit (as suggested per #1355)
- [ ] `--rebase-merges` [1]
- [ ] Anything with selections or multiple cursors [2]

[1] `rebase-merges` requires git v2.18; the `preserve-merges` we used to emulate is deprecated.  Sublime does not ship a syntax for the new todo format, so it looks very ugly.  But it is very powerful, for example "Extract selected commits to a new branch" is possible.  
[2] Examples: "Squash selected commits", "Flip commits" (if 2 are selected)

### DIFF
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -377,6 +377,18 @@
         "command": "gs_show_rebase"
     },
     {
+        "caption": "git: rebase --abort",
+        "command": "gs_rebase_abort"
+    },
+    {
+        "caption": "git: rebase --continue",
+        "command": "gs_rebase_continue"
+    },
+    {
+        "caption": "git: rebase --skip",
+        "command": "gs_rebase_skip"
+    },
+    {
         "caption": "git: reset",
         "command": "gs_reset"
     },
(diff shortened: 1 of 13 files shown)


kaste:
That turns out to be incredible useful and fast.  Even if I need to hack on it further, I probably merge rather soon because I miss it immediately if I change the branch. 

The only things so far: The `R` binding is a bit too magic maybe.  The same action from the menu is clear "Autosquash from X...HEAD" and very useful.  Maybe it's because the command is completely unattended whereby "W" popups the message to reword, and "E" opens the panel with further instruction and updates the graph itself which shows clearly a detached HEAD.

Otherwise minor: Probably renaming the commands which should be internal only right now.

PR #1387, opened 2020-10-13
Maybe offer `force-with-lease` if push fails

### LINKED ISSUE #1130: Status dashboard: Automatically ask for a 'push --force-with-lease'
If we're ahead and behind, the push command (which will fail) should automatically ask the user for a '--force-with-lease'.

kaste:
Fixes #1130 

This generally follows the "Delete Tag" philosophy.  For "Delete Tag" in the graph view, we automatically offer a forced deletion if the branch is not fully merged.  But the default action we select in the quick panel is "Nah, ok". 

We do this here for "pushing" as well.  If the push fails, we show a quick panel to either "abort" or "proceed", the second item offers a retry with `--force-with-lease` set.

Of course, ideally we wouldn't try a push (because it's very slow) *if* we could know that we're behind the upstream.  But we don't since the app state is still not shared.

### DIFF
--- a/core/commands/push.py
+++ b/core/commands/push.py
@@ -1,10 +1,11 @@
 import sublime
 from sublime_plugin import WindowCommand
 
-from ..git_command import GitCommand
+from ..git_command import GitCommand, GitSavvyError
 from ...common import util
 from ..ui_mixins.quick_panel import show_remote_panel, show_branch_panel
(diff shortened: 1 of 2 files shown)


kaste:
Nah, it's horrible to wait 5 seconds for the push to fail and *then* offer the force-with-lease and wait again.  Will re-open a better but larger 🤷 PR.

PR #1402, opened 2020-11-02
Enable GitHub test workflow

kaste:
As travis is becoming too slow.  Enable testing using Github Actions.


kaste:
We can make this a bit more eco-friendly, for example removing Appveyor and macos. Let's see if the Github workflow is reliable first.

PR #1409, opened 2020-11-03
Respect "prompt_for_tracking_branch" when creating pull request

kaste:
As pushing for the first time just before creating the PR is probably
the most common scenario, read "prompt_for_tracking_branch" just like
`gs_push` would do.

### DIFF
--- a/github/commands/pull_request.py
+++ b/github/commands/pull_request.py
@@ -183,10 +183,17 @@ class GsGithubCreatePullRequestCommand(WindowCommand, git_mixins.GithubRemotesMi
 
         if not current_branch.tracking:
             if sublime.ok_cancel_dialog(PUSH_PROMPT):
-                self.window.run_command("gs_github_push_and_create_pull_request", {
-                    "local_branch_name": current_branch.name,
-                    "set_upstream": True
-                })
+                if self.savvy_settings.get("prompt_for_tracking_branch"):
+                    self.window.run_command("gs_github_push_and_create_pull_request", {
+                        "local_branch_name": current_branch.name,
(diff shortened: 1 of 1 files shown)


kaste:
Not liking it.  It should behave like `gs_push` and for example read the same settings, like `prompt_for_tracking_branch`.  But overall the workflow is hitting to many times `<enter>`.  

- First, we ask basically "really?".  For example, you exec `gs_push`, then we ask "really? should we set an upstream?".  Yes of course, that's what I was asking for.  
- We ask for a remote even if there is only one.
- We ask for a remote branch name, although it is weird and usually confusing to *not* take the same name as the local one.  Why would you even do that?  If the name is taken on the remote?  Probably rename the local branch as well.  

All 3 questions can be aborted.  So the "really?" is really not necessary, is it?  I do understand that we want *some* pause before the actual action.  Basically if we were to take the first remote (if there is only one), and just take the local branch name, *then* we would just push it to the server without the "really" question.  Promptly which is maybe to fast.  

We could omit the "really" though, and make the branch name question mandatory.  So one `enter`, or `esc` to abort, if there is only one remote.  Otherwise two times `<enter>`.  We would go from 3 enter's, to either 2 or 1 enter.  (And remove one setting which nobody flipped anyway because the code on `false` was wrong (or annoying) in the it never set the upstream!)

For `gs_push` this would be relatively clear.  For "Create PR", it is a bit confusing because we ask of course for the remote we push the code to which is *not* necessarily the base remote we open the PR against.

kaste:
Close this and look at it in #1411

PR #1412, opened 2020-11-06
Refactor  `gs_delete_branch`

kaste:
- Extract the new `show_actions_panel` helper to `utils` and use it
- Extract the new input handler `ask_for_local_branch`, and make `gs_delete_branch` a `GsWindowCommand` using it.

### DIFF
--- a/common/util/actions.py
+++ b/common/util/actions.py
@@ -1,10 +1,13 @@
+from functools import wraps
+
 import sublime
+
 from ...core.settings import GitSavvySettings
 
 
 def destructive(description):
     def decorator(fn):
-
+        @wraps(fn)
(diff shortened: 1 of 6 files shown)


common/util/actions.py
```diff
@@ -1,10 +1,13 @@
...
+
 import sublime
+
 from ...core.settings import GitSavvySettings
 
 
 def destructive(description):
     def decorator(fn):
-
+        @wraps(fn)
```
kaste, line 10: Note that it is important that decorators use `wraps` to retain the function signature.  Without it decorating a main "run" function with `GsWindowCommand` wouldn't work.

PR #1414, opened 2020-11-06
Cleanup quick panels

kaste:
- For both, `show_remote_panel` and `show_branch_panel` implement a separate `on_cancel` callback.  We usually abort on cancel (value `-1`) which makes all our usages simpler. 

- We had a property `last_remote_used` very visible on `GitCommand` but its getter was never used actually because the caller didn't liked its side-effects.  Remove that and use the `store` for it. 

- Save another git call: `show_branch_panel` used a separate git call to get the current branch name.  Don't do that. Because `get_branches()` returns rich information we already have what we want *if* we don't reduce the info to `.name_with_remote` straight.

### DIFF
--- a/core/base_commands.py
+++ b/core/base_commands.py
@@ -148,8 +148,7 @@ if MYPY:
 def ask_for_local_branch(self, done):
     # type: (GsCommand, Kont) -> None
     def on_done(branch):
-        if branch:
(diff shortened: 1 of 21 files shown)


kaste:
Not that often, always good:
 
image

PR #1416, opened 2020-11-06
Streamline push

### LINKED ISSUE #1411: More streamlined "push" and "create PR" flow
"Creating a PR" should behave like `gs_push` and for example read the same settings, like `prompt_for_tracking_branch`.  But overall the workflow is hitting too many times `<enter>`.  

- First, we ask basically "really?".  For example, you exec `gs_push`, then we ask "really? should we set an upstream?".  Yes of course, that's what I was asking for.  
- We ask for a remote even if there is only one.
- We ask for a remote branch name, although it is weird and usually confusing to *not* take the same name as the local one.  Why would you even do that?  If the name is taken on the remote?  Probably rename the local branch as well.  

All 3 questions can be aborted.  So the "really?" is really not necessary, is it?  I do understand that we want *some* pause before the actual action.  Basically if we were to take the first remote (if there is only one), and just take the local branch name, *then* we would just push it to the server without the "really" question.  Promptly which is maybe to fast.  

We could omit the "really" though, and make the branch name question mandatory.  So one `enter`, or `esc` to abort, if there is only one remote.  Otherwise two times `<enter>`.  We would go from 3 enter's, to either 2 or 1 enter.  (And remove one setting which nobody flipped anyway because the code on `false` was wrong (or annoying) in that it never set the upstream!)

For `gs_push` this would be relatively clear.  For "Create PR", it is a bit confusing because we ask of course for the remote we push the code to which is *not* necessarily the base remote we open the PR against.

_Originally posted by @kaste in https://github.com/timbrel/GitSavvy/pull/1409#issuecomment-721364506_

kaste:
Fixes #1411 

- Remove setting "prompt_for_tracking_branch". 

a) If there is only one remote, take it.  Ask for the remote branch name, but the input handler is prefilled with the name of the current branch.  That essentially leads to one `<enter>` to push. 

b) With more remotes, we first ask for the remote, then for the branch name, again prefilled with the name of the current branch.  Ideally, coming down to two `<enter>`s.  

At any time `ESC` aborts the process.  `-u` is always set. 

If we'd just take the local branch name for granted, we could go to no question, t.i. an immediate push, if there is only one remote.  I don't think that this is a good idea. 

For b) we came down from two to one question because we would pause the process with the remote question.  But the flow would be a bit inconsistent, and it's harder to code.

See https://github.com/timbrel/GitSavvy/issues/1411#issuecomment-723199764 for a different workflow idea. 

~We could leave the setting and just flip its default for a trial.~ Angsthasensetting

- Further refactor `push.py` and use the new default handlers that comes with `GsWindowCommand`.

### DIFF
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -229,12 +229,6 @@
     "max_items_in_tags_dashboard": -1,
 
     /*
-        When set to `true`, GitSavvy will offer to set the upstream on `git: push`
-        when tracking branch is not configured.
-     */
-    "prompt_for_tracking_branch": true,
-
-    /*
         When set to `true`, GitSavvy will automatically display more info about the
         current commit in a output panel.
      */
--- a/core/base_commands.py
+++ b/core/base_commands.py
@@ -14,9 +14,9 @@ MYPY = False
 if MYPY:
     from typing import Any, Callable, Dict, Iterator, List, TypeVar
     CommandT = TypeVar("CommandT", bound=sublime_plugin.Command)
-    Kont = Callable[[object], None]
-    ArgProvider = Callable[[CommandT, Kont], None]
     Args = Dict[str, Any]
+    Kont = Callable[[object], None]
+    ArgProvider = Callable[[CommandT, Args, Kont], None]
 
 
 class WithProvideWindow:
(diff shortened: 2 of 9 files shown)


kaste:
Pull as it is an improvement in any way, has two ❤️s, a refactoring I really like, and I have another PR that relies on this refactor.

PR #1421, opened 2020-11-24
Add "Make fixup commit" to the rebase menu

### LINKED ISSUE #1355: Create fixup commit from graph view

kaste:
Fixes #1355

As suggested in #1355, add a menu entry to "create" a fixup commit from the Repo History.  

In reality, we don't actually "create" the commit but rather prepare the commit message and then open the commit view.  Another `ctrl+enter` will then create the commit. 

This need some discipline as a user might select the option here prior editing, saving and staging the wanted change.  It totally feels like a very useful, "correct" option though here in this menu.  Maybe further tweaking is necessary.

### DIFF
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -82,7 +82,7 @@ class gs_commit(WindowCommand, GitCommand):
     message area with the previous commit message.
     """
 
-    def run(self, repo_path=None, include_unstaged=False, amend=False):
+    def run(self, repo_path=None, include_unstaged=False, amend=False, initial_text=""):
         repo_path = repo_path or self.repo_path
 
         this_id = (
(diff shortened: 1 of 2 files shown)


kaste:
Since we don't "create", maybe `Make fixup commit for ...` instead of `Create fixup commit ...`.  Or: `Prepare fixup commit ...`

Issue #1435, opened 2020-12-03
Improve diff view


kaste:
Following ideas to improve the diff view

For consistency, bind mnemonic keys

- [x] `[sud]` to stage, unstage, or discard hunks
- [x] `[SUD]` to stage, unstage, or discard the whole file

- [x] Allow arbitrary line selections just like "git gui" to stage parts of a hunk or a combination of multiple hunks.  (Currently we abort if something is selected in the view.)

As hunks can get long due to the default context lines being shown -- let's call these modifications separated by context lines **chunks**:

- [ ] Maybe do *not* navigate from hunk to hunk, but from chunk to chunk
- [x] Keep the scrolling friendly. Try to show the whole jumped-to chunk/hunk completely; at the same time try to minimize the scrolling.  For example, do *not* just show the next cursor at center position.

- [ ] Maybe if the cursor is *on* a chunk, `sud` operates on this chunk (instead of the complete hunk).  With this, we would avoid selecting lines for the simple case, where the user usually wants to operate on complete modifications (chunks), seldom on single lines.  We also avoid that the user *zooms* in just to break the diff into more individual hunks.

- [x] Zooming must do better with cursor and viewport positioning or it's useless because too confusing.

PR #1441, opened 2021-01-13
Improve staging (et.al) in the diff view

kaste:
Ref #1435 

- For consistency with the status dashboard, bind `sud` to stage, unstage, or discard.
- Allow to act on arbitrary selections, t.i. allow partially staging a hunk or multiple hunks, for example a single line.
- Bind `SUD` to stage, unstage, discard the complete file under the cursor.

Staging single lines or generally parts of a hunk usually comes with some complications as to where to apply these lines, t.i. basically which starting lines we compute for them.  Both our inline-diff and for example `git gui` apply some heuristics and maybe artificially move a `+` to a (maybe, hopefully) desired line.  We don't do this here but instead expect that the user indicates making *multiple* selections which lines do belong together.  E.g.:

image

As there are two lines selected, we handle them as a "symmetric" modification, and it will just work as intended.  (In contrast to git gui, first staging the single "+  1environ" line and then the "-   environ" line will **not** work as expected.)   

Note that these heuristics used elsewhere, often don't work.  For example, "git gui" is optimized for 2/2 symmetric modifications.  It doesn't work for 3 lines/3 lines symmetric modifications.  They usually also fail for asymmetric changes, like 1-/2+.

In contrast to "git gui", we construct `-U0` patches and compute correct hunk headers (starting lines).

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -635,6 +635,37 @@
         ]
     },
     {
+        "keys": ["s"],
+        "command": "gs_diff_stage_or_reset_hunk",
+        "context": [
+            { "key": "setting.command_mode", "operator": "equal", "operand": false },
+            { "key": "setting.git_savvy.diff_view", "operator": "equal", "operand": true },
+            { "key": "setting.git_savvy.diff_view.in_cached_mode", "operator": "equal", "operand": false },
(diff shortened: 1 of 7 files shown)


braver:
@kaste man, I ❤️ you. Great job here, as always!

kaste:
Thanks man, ♥️ you too.

PR #1450, opened 2021-02-05
Mark merge commits with an `○`

### LINKED ISSUE #1447: Mark merge commits in the History view
In the repo history it is hard to see the merge commits quickly.  

image

image

Maybe the merge commits should have a different dot?  E.g. `◯`

image

image

kaste:
Fixes #1447

For orientation mark merge commits differently from normal commits.
Unfortunately, "git log" does not output all parents when using `%p`
but only the visual drawn parents.  T.i. when we tell git to simplify
the graph it only marks merge commits when the ascii art already
indicates a merge.

Since `%p` is not reliable, do a brute force search for typical merge
commit messages.

Use the scope `keyword.graph.commit.merge.git-savvy` for highlighting.

### DIFF
--- a/core/commands/log_graph.py
+++ b/core/commands/log_graph.py
@@ -71,7 +71,8 @@ if MYPY:
 
 
 COMMIT_NODE_CHAR = "●"
-COMMIT_NODE_CHAR_OPTIONS = "●*"
+MERGE_COMMIT_NODE_CHAR = "○"
+COMMIT_NODE_CHAR_OPTIONS = "●*○"
 GRAPH_CHAR_OPTIONS = r" /_\|\-\\."
 COMMIT_LINE = re.compile(
     r"^[{graph_chars}]*[{node_chars}][{graph_chars}]* "
@@ -551,9 +552,10 @@ class gs_log_graph_refresh(TextCommand, GitCommand):
         enqueue_on_worker(self.run_impl, should_abort, navigate_after_draw)
 
     def format_line(self, line):
+        is_merge_commit = re.search("Merge (pull request|branch)", line)
         return re.sub(
             r'(^[{}]*)\*'.format(GRAPH_CHAR_OPTIONS),
-            r'\1' + COMMIT_NODE_CHAR,
+            r'\1' + (MERGE_COMMIT_NODE_CHAR if is_merge_commit else COMMIT_NODE_CHAR),
             line,
             flags=re.MULTILINE
         )
@@ -1089,7 +1091,7 @@ def dots_after_dot(dot, forward=True):
     # type: (colorizer.Char, bool) -> Iterator[colorizer.Char]
     """Return exact next dots (commits) after `dot`."""
     fn = colorizer.follow_path_down if forward else colorizer.follow_path_up
-    return filter(lambda ch: ch == COMMIT_NODE_CHAR, fn(dot))
+    return filter(lambda ch: ch.char() in COMMIT_NODE_CHAR_OPTIONS, fn(dot))
 
 
 class gs_log_graph_navigate_to_head(TextCommand):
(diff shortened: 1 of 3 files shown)


kaste:
@braver can you take a look here, please.

The general idea is (in the log or graph view) that we have two kinds of commits, normal commits and merge commits.  If you have a very flat dev line it is hard to see the merge commits; but they still mark important points in a long streak.  

For example here I have a flat line:

image

but still with two merges; so we actually see the development of two or three features.  

Or here:

image

where you nicely see an octopus merging multiple separate things and above and below some straight lines, but again they also have merges in them and it's a lot less clear that we have separate things in here as well.

So the simple idea is to mark merge commits with a different dot.  

But now that I have them, I'm not completely convinced. 

For example, with more short branches we get also more merges and end with

image

Is that still informative or rather noisy with the different dot types?

And here a really flat development:

image

It seems still impossible to see the three merges in here.

So is the dot idea maybe too subtle?

Thanks for any feedback.

asfaltboy:
Thanks for taking the time to illustrate with different screenshots.

I personally find myself often starting to read the merge commit just to learn I have to skip it, so looking for "filled dots" will definitely make my life easier, so I think we ought to deploy this and see if we get differing feedback from users.

On the topic of multiple branches, I wonder.. did we ever consider coloring the different branches (and their corresponding "dots") in different colors, or is that rainbow/noisy as well?

braver:
Subtle is good. I guess I'm usually either looking for merge commits (to find where a feature got merged) or trying to ignore them (no sense reading the commit message). Depending on how a team works the merge commits can really indicate the rhythm of development, or maybe in most cases it's just really random. Since you can't really trust the commit message to start with "Merge" a visual indicator is nice. But I wouldn't want them to stand out too much. Already in your "short branches" scenario it gets really noisy. Because it's ascii art you need to be really careful, you can't balance out the noice with clean graphics.*

So, what I'm seeing is:

```
●
○
●
```
The case above is super useful, you wouldn't immediately recognise that as a merge commit otherwise.

```
●
○
 \
  ●  
```
Here it's a bit superfluous. The diagonal line into the dot cannot be anything other than a merge. 

```
●
●
 \
  ●
  ○
  ●
```
Maybe this would be better? I.e. open dots for merge commits on a straight line, but closed dots where the other branch already visually comes into the main line.

I was also considering maybe using italics for the commit id of merge commits? Not a lot of monospace fonts come with proper italics though. And your eye is on the graph itself usually to follow the paths. Some other tools I used to use (back to GitSavvy 100% of the time since a few months) represent the merge commits with a "fatter" dot: just a tiny fraction larger, maybe just one pixel. With ascii though the options are limited and we don't want to turn this into a Christmas tree.

*) Except of course in theory it doesn't need to be ascii art, you could use phantoms to render the graph. But that's a whole different kettle of fish :) I personally, unreasonably, love the ascii.

kaste:
Okay, thanks for the first feedback.  I think it's totally correct that there are superfluous marks when the graph shows that we're branching (by going diagonal) in an immediate way. 

Now, as the merge commits form the rhythm, there are also less merge commits than normal commits, what do we want emphasize or de-emphasize them?  I think we want to emphasize and that's not in form with the empty circle because then it's lighter than the normal dot.  (I think in the ascii graph @braver sent it is very clear that the solid circles are bold and the other light.) 

But we're also used to the bold, filled circles, so I cannot just switch them.  It's a too drastically change. 

For example, leave the dots and just colorize:

image



Or:

image

Or, ugly but did prove to me that we want the emphasis and rhythmic:

image

In the last picture, you almost want to collapse these merges, like click on the merge commit to hide and show the commits below it.

All these example in this comment are purely driven by the syntax, as such text based looking for "Merge branch" (et.al.) in the message.

braver:
Ugh, I hate the italics, I should’ve known better. I do really like the slight coloring of the dots, that’s a lot better than open/closed dots.
The last (ugly) one proves to me that you need subtlety here. If you have a lot of small merges close to each other it quickly becomes hard to visually “parse”. Maybe also color the commit id in addition to the dot, but that first screenshot here looks pretty good to me!

kaste:
I also use what I have in the first screenshot right now.  But the PR will be much different as all of these are just syntax driven and watch out for "Merge branch" on the same line.  We also don't control the colors, we just add a scope.  It's unlikely we could find a scope which looks reasonable/subtle for default themes.  We can probably only ship this as a "hidden" feature as it requires editing the color scheme.

It's okay because the graph view looks maybe very different from what I have when using default themes anyway.

kaste:
We could redraw the ascii art on a keystroke.  T.i. emphasize merge commits:

ORicYb5rGs

Or provide a dense, overview "style" where we basically zoom in.

kW06UC1Vyo

In all of these we would not rely on text processing/parsing, and we're dynamic: repaint on keystroke.

kaste:
Closing as per comments here.  Generally, I probably prefer the "repaint on keystroke" approach t.i. for example the first animated gif in my previous comment.

PR #1453, opened 2021-02-08
Only parse text between "@" markers in the hunk header

### LINKED ISSUE #1452: Files incorrectly flagged as merge conflicts during inline diff
For some files, when I try to start an inline diff it is flagged as containing a merge conflict and the diff does not open.

I tracked down the source of this and it was introduced in https://github.com/timbrel/GitSavvy/commit/8df9b46a4d34014949286aaecad3a6398fa1ee7b, specifically in `safely_parse_metadata`

The issue seems to be that the regex is picking up extra results based on the content of the diff, in addition to the metadata. As an example, for the input:

```
@@ -7,0 +8 @@ RCall = "6f49c342-dc21-5d91-9882-a32aef131414"
```

the output is `[(7, 0), (8, 1), (5, 1), (9882, 1)]`

Another example is 

```
@@ -295 +313 @@ docs-main:iai-2.0.0:
```

giving `[(295, 1), (313, 1), (2, 1)]`

It seems like the previous version was limiting the search to inside the `@@ ... @@` whereas now it is not, and it is flagging the diff a merge conflict because there are more than two elements in the returned metadata.

kaste:
Fixes #1452

We currently parse the whole line which can fail because the text after
the second "@" marker can contain arbitrary context information.

The fix is to parse only the text between the "@" markers.

### DIFF
--- a/core/parse_diff.py
+++ b/core/parse_diff.py
@@ -219,7 +219,9 @@ class HunkHeader(TextRange):
         """
         return [
             (int(start), int(length or "1"))
-            for start, length in SAFE_PARSE_HUNK_HEADER.findall(self.text)
+            for start, length in SAFE_PARSE_HUNK_HEADER.findall(
+                self.text.lstrip("@").split("@", 1)[0]
+            )
         ]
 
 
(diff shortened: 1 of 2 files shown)


JackDunnNZ:
Thanks so much for the quick fix @kaste!

kaste:
But you already did the bisect 🤹‍♀️ 🕺

PR #1455, opened 2021-03-07
Refactor getting the repo path

kaste:
This is mostly a boring refactoring around `self.repo_path`.  Some guidance:

* Make the dependencies more clear.  Do we need a *view*, or a *window*, or a *filename*?   
* Avoid repeated "realpath" calls.  On Windows we never had this functionality anyway, it was a noop.  But also `git --show-toplevel` already returns the resolved path.
* Implement `resolve_path` (as a primitive `realpath`) for all OSes.  (We still need this for `file_path` as `repo_path` is always resolved.)
* Do not cache the `file_path` nor the `repo_path` on the view.

### DIFF
--- a/core/commands/amend.py
+++ b/core/commands/amend.py
@@ -16,6 +16,11 @@ class gs_amend(WindowCommand, GitCommand):
 
 class gs_quick_stage_current_file_and_amend(gs_amend, GitCommand):
     def run(self):
-        self.git("add", "--", self.file_path)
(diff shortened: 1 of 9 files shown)


kaste:
Overall, this should not make a difference, but will fix a bug on Windows where the `repo_path` was already resolved (by git) but the `file_path` was not because we had no Windows implementation of `realpath`.

PR #1462, opened 2021-03-30
Suggest remote branch when creating a PR

kaste:
Instead of always offering the remote/branch-name wizard suggest
a `remote/local_branch_name`.  This keeps the eye at the top of the
view on the Palette Control for the normal case.

image


kaste:
The UX is correct but we can't use `get_integrated_remote_name` as it's designed to find the final upstream repository (where the issues and PR's land).  The default push is typically your fork, typically `origin` or maybe `fork` or in GitHub your username, e.g. `kaste`.  

We could also read out `remote.pushDefault` *but* as defining `pushDefault` changes everything `git push` unlikely *our* users do this.  (If you set `pushDefault`, just `git push` will push to that repo/remote and **not** to upstream anymore.) GitSavvy always pushes to the configured upstream.

kaste:
Closing.  Follow up is #1475

PR #1475, opened 2021-05-27
Suggest remote branch when creating a PR

kaste:
Instead of always offering the remote/branch-name wizard suggest
a `remote/local_branch_name`.  This keeps the eye at the top of the
view on the Command Palette for the normal case.

For the remote branch name, always suggest the local name.  For the
remote name, respect the values `gitsavvy.pushdefault` and
`remote.pushdefault` from the config, and fallback to "fork", origin",
and then any other known name.

Always use an in memory cache `last_remote_used_for_push` which stores
the last used value.  This key is now distinct from `last_remote_used`.
Automatically persist a used remote in `gitsavvy.pushdefault` if the
suggestion was overriden by the user so it will be suggested the next
time; even after restarts.  Persist after the comlete wizard has been
finished, just before pushing, t.i. cancelling via `ESC` has no further
side effects and will not change the suggestion.

Note that we use `.pushdefault` as key suffix to get all values in one
"git config" call.

Replaces #1462 

image

### DIFF
--- a/core/base_commands.py
+++ b/core/base_commands.py
@@ -12,10 +12,14 @@ from GitSavvy.core.ui_mixins.quick_panel import show_branch_panel
 
 MYPY = False
 if MYPY:
-    from typing import Any, Callable, Dict, Iterator, List, TypeVar
+    from typing import Any, Callable, Dict, Iterator, List, Protocol, TypeVar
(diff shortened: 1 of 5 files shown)


braver:
🚀

kaste:
🤗

PR #1487, opened 2021-06-04
Improve `gs_clone` and `gs_init`

### LINKED ISSUE #1469: For `GsClone`, read the clipboard for potential urls to clone from
As a quality for life minor improvement, read the clipboard contents, check them if they contain a valid git repo url, and offer them as a default for the user. 

Ideally, construct the right git url from a more complex or deeper url.  For example, if you're browsing a repo on GitHub, you might be looking at

- `https://github.com/timbrel/GitSavvy`  on the main page
- `https://github.com/timbrel/GitSavvy/blob/master/core/parse_diff.py` when looking at a specific file

etc.  

A good candidate is always `https://github.com/timbrel/GitSavvy.git`.  Basically, the first two parts of the path + `.git`

### LINKED ISSUE #1470: For `GsClone`, a better target dir candidate is the parent of the current workspace folder
For both `GsClone` and `GsInit`, we first offer `os.path.dirname(view.file_name())` as a candidate to clone to.  

Typically, that's not useful.  Instead assume the projects are opened at the root folder, t.i. `folders[0]` is right below the common project folder (often `~`), and therefore suggest `os.dirname(window.folders[0]) + project_name` as the clone target. 

Maybe introduce a setting, like `default_projects_folder` which might point to `~/my_projects` or `X:\` and just take that.

For `GsInit`, `window.folders[0]` almost always should be correct.

kaste:
Fixes #1469 
Fixes #1470 

- Read a possible url from the clipboard to use in the clone wizard
- Suggest a sibling parent of the current open folder as the target to clone to
- Use the output panel as a progress indicator.  
- Open a new window which will host the to be cloned project immediately.  This leads to a much nicer UX.  But reuse an empty window.
- If there is no open folder, suggest `~` or, on Windows, `~\Desktop`.

For `gs_init`, suggest to initialize the repository in the currently open folder.

We want to support two workflows:

- Cloning as a sibling project
  Invoke `gs_clone` from any, already used window
- Cloning possibly temporary projects to a default location
  Invoke `gs_clone` after (`ctrl+shift+n`) creating a new, empty window.

TBC:  It seems plausible to let the user define the default project root.

### DIFF
--- a/core/commands/init.py
+++ b/core/commands/init.py
@@ -172,10 +172,17 @@ class gs_clone(WindowCommand, GitCommand):
         enqueue_on_worker(self.do_clone)
 
     def do_clone(self):
-        window = open_folder_in_new_window(self.target_dir)
+        if not self.window.folders() and not self.window.views():
+            window = self.window
+        else:
+            window = open_new_window()
+            # HACK: `git()` uses `self.window`, we can't pass anything down
+            self.window = window
+
+        window.set_project_data({
+            "folders": [dict(follow_symlinks=True, path=self.target_dir)]
+        })
         window.status_message("Start cloning {}".format(self.git_url))
-        # HACK: `git()` uses `self.window`, we can't pass anything down
-        self.window = window
         self.git(
             "clone",
             "--recursive" if self.recursive else None,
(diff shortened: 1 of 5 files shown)


kaste:
The tests fail now bc Package Control is down

PR #1494, opened 2021-06-09
Fix: Extract filename from diff if it has spaces

kaste:
If the filename contains spaces, git adds a "\t" at EOL.  Strip this
tabulator sign as it's not part of the name.


kaste:
Failing tests bc PC dependencies are down

PR #1495, opened 2021-06-09
Test if the reported filename is accessible

kaste:
As an edge case, for example when viewing a file in a zip container,
Sublime Text might return a "correct" looking file path although the
later `resolve_path` will throw as the file does not exist.

That case was generally handled by a later call to `os.isdir`.  It is
only within this subroutine we might use the invalid filename.

Bug introduced by a5561c59 (Fallback to first folder only if necessary).


kaste:
Failing tests bc PC dependencies are down

PR #1506, opened 2021-11-14
fix rebase dashboard for git 2.34

### LINKED ISSUE #1505: rebase dashboard issue on git 2.34 rc1.
When using the rebase dashboard with "rebase_preserve_merges" setting on.

The panel shows
```
$ git rebase -p master

fatal: --preserve-merges was replaced by --rebase-merges

```

It seems that the setting name has changed from `--preserve-merges` to `--rebase-merges` in git 2.34.0.rc1, [link](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---rebase-mergesrebase-cousinsno-rebase-cousins).

It is a work computer and the git version is supplied by my company.
```
(=)-~$ git --version
git version 2.34.0.rc1
```

randy3k:
close #1505

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Sublime Text 3 plugin providing the following features:
 - `git diff` view, allowing user to (un)stage hunks across all files
 - status, branch, tag, and rebase dashboards
 
-**Note:** GitSavvy requires Git versions at or greater than 2.16.0.
+**Note:** GitSavvy requires Git versions at or greater than 2.18.0.
 
 **Note:** Sublime Text 2 is not supported.  Also, GitSavvy takes advantage of certain features of ST3 that have bugs in earlier ST3 releases.  For the best experience, use the latest ST3 dev build.
 
--- a/core/git_command.py
+++ b/core/git_command.py
@@ -46,7 +46,7 @@ encoded files.  In the latter case use the 'fallback_encoding' setting.
 -- Partially decoded output follows; � denotes decoding errors --
 """
 
-MIN_GIT_VERSION = (2, 16, 0)
+MIN_GIT_VERSION = (2, 18, 0)
 GIT_TOO_OLD_MSG = "Your Git version is too old. GitSavvy requires {:d}.{:d}.{:d} or above."
 
 NOT_SET = "<NOT_SET>"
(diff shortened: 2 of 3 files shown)


kaste:
Well, having a `self.git_version` property would be nice; but a property with a huge side-effect is not a fit.  Also: `git_path` is a hidden global here and it can be `None` which would break the code.  

It is not necessary to do this change so I would probably leave it out.

kaste:
Honestly, if we get a clean, cached `self.git_version` we don't need to bump the requirement and can just support both but it's more work.

core/git_command.py
```diff
@@ -46,7 +46,7 @@
 -- Partially decoded output follows; � denotes decoding errors --
 """
 
-MIN_GIT_VERSION = (2, 16, 0)
+MIN_GIT_VERSION = (2, 19, 0)
```
kaste, line 49:
```suggestion
MIN_GIT_VERSION = (2, 18, 0)
```
https://git-scm.com/docs/git-rebase/2.18.0#Documentation/git-rebase.txt---rebase-mergesrebase-cousinsno-rebase-cousins

README.md
```diff
@@ -16,7 +16,7 @@ Sublime Text 3 plugin providing the following features:
 - `git diff` view, allowing user to (un)stage hunks across all files
 - status, branch, tag, and rebase dashboards
 
-**Note:** GitSavvy requires Git versions at or greater than 2.16.0.
+**Note:** GitSavvy requires Git versions at or greater than 2.19.0.
```
kaste, line 19:
```suggestion
**Note:** GitSavvy requires Git versions at or greater than 2.18.0.
```

randy3k: Thanks.

kaste:
I can't take the last commit which changes the design of the rebase dashboard as it leads to e.g.

image

Note that `self.get_short_hash()` really just shortens a long sha.  (You can imagine it as a simple `return input[:6]`.)  The new introduced `MERGE BASE xxx` shows `{base_commit}` which is always present 3 lines below as `        ┬ ({base_commit})` so that's not more useful than before.

The current look is 

image

Note that the first line `        ┬  (xxx)` is also styled differently.  So the syntax change is probably not correct as well.

randy3k:
Ok, let me revert the last commit and work on it later.
Thank you for your comment.

PR #1513, opened 2021-12-06
Disable indentation guessing for historical file views

kaste:
Show the historical file revision as verbatim as possible as we're
comparing the file with the previous version read from git directly.
(Using `set_reference_document`.)


stoivo (approved):

stoivo:
Do you want me to merge if I don't have any comments?

If it was up to me I would squash and merge

kaste:
I dunno; if it gets too much I might rebase to make the git log look nicer.

PR #1564, opened 2022-08-05
Fix colorizing merge conflicted files

kaste:


kaste:
Unfortunately I have no gif for this patch.  This patch is from a very old stash.

PR #1611, opened 2022-09-24
GitLab: correctly fetch opened merge requests

loris:
`gitlab: review merge request` is using the `List merge requests` GitLab API endpoint: https://docs.gitlab.com/ee/api/merge_requests.html
The default behavior for the `state` query parameter is to `return all merge requests`.

In order to be consistent with the documentation (`Display open merge requests on the base repo`) + how the `github: review pull requests` work, I added a `state=opened` query parameter to the API call to display only opened MRs.

I also slightly refactored the helper methods to correctly differentiate the params from the path and the params from the query.

### DIFF
--- a/gitlab/commands/merge_request.py
+++ b/gitlab/commands/merge_request.py
@@ -34,7 +34,7 @@ class gs_gitlab_merge_request(WindowCommand, GitCommand, git_mixins.GitLabRemote
     def run_async(self):
         self.remote_url = self.get_integrated_remote_url()
         self.base_remote = gitlab.parse_remote(self.remote_url)
-        self.merge_requests = gitlab.get_merge_requests(self.base_remote)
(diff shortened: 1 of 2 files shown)


kaste:
Yes, that looks good to me.  Gitlab has not been updated for a long time here. So you when you spot errors let us know.

kaste:
Do you think I should I release this as a hotfix?  Was it unusable before?

loris:
Thanks! No need for a hotfix, current command is working, but just returning more items than necessary (all merge requests, even the ones already closed/merged).
And personally, I'm not in a hurry for the update as I already patched my extension locally

kaste:
I released because there is too much going on on master right now

kaste:
You're welcome to beta test master of course.  😁

PR #1619, opened 2022-09-26
Iterate on the interfaces (3/branches)

kaste:


core/interfaces/branch.py
```diff
@@ -219,7 +243,7 @@ class BranchInterfaceCommand(ui.InterfaceCommand):
     interface = None  # type: BranchInterface
 
 
-class GsBranchesCheckoutCommand(BranchInterfaceCommand):
+class gs_branches_checkout(BranchInterfaceCommand):
```
stoivo, line 263: Why do you want snake case classes?

kaste: It's so that find-in-files etc works.  For example, you're in the key bindings or commands and via find in files quickly switch to the implementation and vice versa.  The class name is now just the real command name, snake-cased and without the command suffix.

And hi!  Didn't expect you reading this anymore.

stoivo: hmm, what does flake8 thing of it? I think I looks strange but we never instansiate the classes so it not the worst. 
No. I don't do much related to gitsavvy but I try to follow and read a few PRs.

kaste: flake8 is okay with it.

PR #1637, opened 2022-10-09
Display worktree path in branch list

airtonix:
Works towards providing worktree support.

This PR just displays the worktree in the branch list.

- [ ] fix unit tests

---

No worktrees other than current branch.
image

after running `git worktree add ../clusterfiles-template template` : 
image

after running : `subl ../clusterfiles-template`
image

I noticed the main branch window shows : 

image

So perhaps we should be consistent and show the worktree like: 

image

---

I can follow up with creating and manipulating worktrees in some follow up PRs

### DIFF
--- a/core/interfaces/branch.py
+++ b/core/interfaces/branch.py
@@ -199,8 +199,9 @@ class BranchInterface(ui.ReactiveInterface, GitCommand):
     def _render_branch_list(self, remote_name, branches, descriptions):
         # type: (Optional[str], List[Branch], Dict[str, str]) -> str
         remote_name_length = len(remote_name + "/") if remote_name else 0
+        current_worktree = next((b.worktree_path for b in branches if b.active), None)
         return "\n".join(
-            "  {indicator} {hash:.7} {name}{tracking}{description}".format(
+            "  {indicator} {hash:.7} {name}{tracking}{description}{worktreepath}".format(
                 indicator="▸" if branch.active else " ",
                 hash=branch.commit_hash,
                 name=branch.canonical_name[remote_name_length:],
@@ -208,7 +209,12 @@ class BranchInterface(ui.ReactiveInterface, GitCommand):
                 tracking=(" ({branch}{status})".format(
                     branch=branch.upstream.canonical_name,
                     status=", " + branch.upstream.status if branch.upstream.status else ""
-                ) if branch.upstream else "")
+                ) if branch.upstream else ""),
+                worktreepath=(
+                    " ({})".format(branch.worktree_path)
+                    if branch.worktree_path and branch.worktree_path != current_worktree
+                    else ""
+                )
             ) for branch in branches
         )
 
--- a/core/git_mixins/branches.py
+++ b/core/git_mixins/branches.py
@@ -24,6 +24,7 @@ if MYPY:
         ("canonical_name", str),    # e.g. "origin/master"
         ("commit_hash", str),
         ("commit_msg", str),
+        ("worktree_path", str),
         ("active", bool),
         ("is_remote", bool),
         ("committerdate", int),
@@ -37,6 +38,7 @@ else:
         "canonical_name",
         "commit_hash",
         "commit_msg",
+        "worktree_path",
         "active",
         "is_remote",
         "committerdate",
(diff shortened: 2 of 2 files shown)


airtonix:
I noticed the main branch window shows : 

image

So perhaps we should be consistent and show the worktree like: 

image

kaste:
So we're in design phase already. Hm, I do prefer the `[ worktree: ... ]` thing over `root: ` but generally it is not tabular enough.  A column layout would make this more clear but maybe wastes too much space as the line lengths are going wild already. For example, I have usually much longer branch names, and then the tracking information comes along usually duplicating the branch name.  

image

I likely prefer one of those, esp. look at the greyed out on `less-status-polling` which really does not look good when compared to the other ones.  I really prefer `checked out at` t.i. something english and explicit over the "worktree" technical name.  Not sure if we really need the tag like thing around the commits, e.g. `[abcdef]` compared to `abcdef`, but if I remember correctly most actions are not valid for branches protected or checked out as worktrees or do something different.  (For example a `d` will *not* remove the branch but the worktree *maybe*.)  So the extra visual would indicate that these are not our normal branches.  

Then of course, we could just group the branches so all worktree branches comes after the "normal" ones.  But that's also not good because worktrees become detached too often.   (I actually find this potentially very confusing.  Can I check out a branch which is in rebasing/merge conflict mode in a worktree.  Can we stabilize this and always tell the user that you should not checkout this branch because it is handled in a worktree.  It just seems to me that the relation between worktree and branch is too much in flux.)

core/interfaces/branch.py
```diff
@@ -156,17 +157,20 @@ def render_branch_list(self, remote_name=None, branches=None):
         if not branches:
             branches = [branch for branch in self._branches if not branch.is_remote]
 
+        current_worktree = self.get_current_worktree()
```
kaste, line 160: This is a side-effect and as such not allowed in the render phase where you should just format the data you got.  It is also a redundant call because you have already all the branches in `branches`.

core/interfaces/branch.py
```diff
@@ -156,17 +157,20 @@ def render_branch_list(self, remote_name=None, branches=None):
...
                 indicator="▸" if branch.active else " ",
                 hash=branch.commit_hash,
                 name=branch.canonical_name[remote_name_l:],
                 description=" " + branch.description if branch.description else "",
                 tracking=(" ({branch}{status})".format(
                     branch=branch.upstream.canonical_name,
                     status=", " + branch.upstream.status if branch.upstream.status else ""
-                ) if branch.upstream else "")
+                ) if branch.upstream else ""),
+                worktreepath=" root: " + branch.worktree_path if branch.worktree_path and not remote_name and current_worktree != branch.worktree_path else ""
```
kaste, line 173: I doubt remote branches can have a `worktree_path`.

airtonix:
I'm happy to do whatever you want me to do, just give me direction 👍🏻.

I'll come back to this later in the week.

kaste:
I think we should go for 

image

then.  This way we also have a way to select or deselect these rows with the syntax selectors.  

For the scope, I went with `keyword` here.  

So the sha is probably 

```
meta.git-savvy.branches.branch.sha1 keyword.other.git-savvy.sha1
```

and the whole first line has `meta.git-savvy.branches.branch.as-worktree`, and the path on the second line is `keyword.other.git-savvy.path` or so.  The names are almost random as we don't have a typical syntax for a language here.

kaste:
Maybe the paths should be relative *iff* they would only have *one* `..` part (t.i. sibling paths).  Otherwise absolute.  For absolute Windows paths, `os.path.normpath` must be called to normalize the slashes. But not for relative paths.  T.i. on Windows a relative path is `foo/bar.py` but the absolute path is `C:\foo\bar.py`.

kaste:
Without further keybindings,  `rename` and `delete` already have a meaningful name for branches checked-out-as-worktree and could be implemented seemingly. 

`checkout` should probably *switch* (keeping the window).  Not sure how to do that.  To open a new folder in a new window is actually simple:

```
sublime.run_command("new_window")
window = sublime.active_window()
window.set_project_data({
    "folders": [dict(follow_symlinks=True, path=<some_path>)]
})
```

Don't know if there is a shorter path for the newer ST4 to do the same.  And if there is an elegant way to *just* switch.  If there is already an open Window it is important I believe that `window.project_file_name()` (and `window.workspace_file_name` ) also changes so that we don't override anything.  Ideally, we would somehow inherit the project and workspace.

kaste:
Closing because you probably won't come back anymore.

Issue #1502, opened 2021-08-26
OSError: resource 'auto' not found


cig0:
Hi,

I've got this error when `git: File history`, `git: Line history`, and `git: Repo history`:

```python
Package Control: Skipping automatic upgrade, last run at 2021-08-26 16:42:29, next run at 2021-08-26 17:42:29 or after
Exception in thread Thread-47:
Traceback (most recent call last):
  File "/Users/cig0/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/common/theme_generator.py", line 58, in __init__
    self.color_scheme_string = sublime.load_resource(original_color_scheme)
  File "/Applications/Sublime Text.app/Contents/MacOS/Lib/python33/sublime.py", line 435, in load_resource
    raise IOError('resource "%s" not found' % name)
OSError: resource "auto" not found

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./python3.3/threading.py", line 901, in _bootstrap_inner
  File "./python3.3/threading.py", line 858, in run
  File "/Users/cig0/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/commands/log_graph.py", line 237, in augment_color_scheme
    themeGenerator = ThemeGenerator.for_view(view)
  File "/Users/cig0/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/common/theme_generator.py", line 53, in for_view
    return JSONThemeGenerator(color_scheme)
  File "/Users/cig0/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/common/theme_generator.py", line 156, in __init__
    super().__init__(original_color_scheme)
  File "/Users/cig0/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/common/theme_generator.py", line 63, in __init__
    raise IOError("{} cannot be found".format(original_color_scheme))
OSError: auto cannot be found

```

The output seems to be fine, however.

Specs:
Sublime Text 4 Build 4113
GitSavvy: 2.31.0

cig0:
Hi @stoivo:

```json
"color_scheme": "auto",
"light_color_scheme": "Packages/Color Scheme - Default/Celeste.sublime-color-scheme",
"dark_color_scheme": "Packages/Theme - Monokai Pro/Monokai Pro.sublime-color-scheme",
```

stoivo:
[...]

PR #1645, opened 2022-10-17
support "color_scheme": "auto"

gwenzek:
Simple fix to #1502 

The main caveat is that if you switch the global color scheme light -> dark, the Gitsavvy view will stil use the old scheme that was generated from the light scheme, even though all other views are now using the dark one.
I'm not sure there is a better fix, since there is no event to detect this kind of changes. 

For me the "_right solution_" would be to replace the Git savvy scopes https://github.com/gwenzek/GitSavvy/blob/3b71e40b3f708e4163a2b0b636fca325f1b5a33b/core/commands/log_graph.py#L90-L94

by the "magic" colored scope (`region.redish`, etc): https://www.sublimetext.com/docs/api_reference.html#sublime.View.add_regions
This way you don't need to override the view scheme, and you let SublimeText deal with the light/dark scheme.
You can see what it would look like at: https://github.com/timbrel/GitSavvy/commit/10f08252a95e32e8ffa3e72ea5aa7e67750a996c.
This would require a bit more work to make the colors configurable, let me know if you're interested.

### DIFF
--- a/common/theme_generator.py
+++ b/common/theme_generator.py
@@ -46,7 +46,7 @@ class ThemeGenerator():
     @staticmethod
     def for_view(view):
         # type: (sublime.View) -> ThemeGenerator
-        color_scheme = view.settings().get('color_scheme')
+        color_scheme = sublime.ui_info()["color_scheme"]["resolved_value"]
         if color_scheme.endswith(".tmTheme"):
             return XMLThemeGenerator(color_scheme)
         else:


gwenzek:
Hmm, I didn't realized you were still supporting ST3. I'll update the PR.
For my the Mypy lint error is wrong. Where are the type stubs coming from?

PR #1658, opened 2022-11-03
Symptom: the innocuous looking

kaste:
Symptom: the innocuous looking

```
class gs_reset(ResetMixin, LogMixin):
    pass
```

did not work because it did not call `LogMixin.run()`.  Flipping there did not call `ResetMixin.do_action()` but the one from `LogMixin`.

The bases were defined as follows:

```
class ResetMixin(GsWindowCommand):
class LogMixin(GitCommand):
```

I can't calculate the MRO here so this is more intuition and hopefully luck

Introduced-by: 84c8c719 (Use `GsWindowCommand` in `reset.py`)

### DIFF
--- a/core/base_commands.py
+++ b/core/base_commands.py
@@ -133,19 +133,19 @@ def _get_run_command(cmd):
 
 
 class GsTextCommand(
+    GitCommand,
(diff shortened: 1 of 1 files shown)


kaste:
Shit, messed up the commit message during rebase 🤷

PR #1684, opened 2022-12-08
Change one type ignore line because mypy update

kaste:
🤷


kaste:
It auto-merged although mypy was red.  Great.

PR #1724, opened 2023-01-11
expand variables in env from setting by os.environ

zebyja:
As I'm primary python developer I'm using virtual environments per project. I'm using also pre-commit hooks for linting my code before commit. For fit version of liters I'm using local linters installed in my virtual environment, so I can easily fit version of my linters in requirements.txt. 

Unfortunately, this workflow require using git inside virtual environment (with activated VENV). Thats why my GitSavvy setting looks like that:
```
{
    "env": {
        "PATH":"venv/bin:${PATH}"
    },
}
```

This is the easiest way how I can activate my venv what I found.

This is my first edit of GitSavvy (any sublimeText plugin also), so I welcome suggestion to do that better.
I think that it could be better if I could expand project_path also, but I'm not sure how to do that and this is working for me, so I'm not much motivated to examine that.


core/git_command.py
```diff
@@ -279,9 +279,11 @@ def git(
                 raise GitSavvyError(str(e), show_panel=show_panel_on_error, window=window)
 
         stdout, stderr = None, None
+        savvy_env = self.savvy_settings.get("env")
+        savvy_env_expanded = {k: sublime.expand_variables(v, os.environ) for k, v in savvy_env.items()}
```
kaste, line 283: At this point `savvy_env` can be `None`.  That's why we do the `or {}` in the original code.

kaste (changes requested):
I do like it.  There is only the small `None` issue.  We should probably document this https://github.com/timbrel/GitSavvy/blob/e7bb9c1d5251448277961af6d72826a4d7b80b1f/GitSavvy.sublime-settings#L25-L29 in the settings file just with you example `"PATH":"venv/bin:${PATH}"` as it is compelling.

kaste:
We're also red here because of mypy.  The stubs for `expand_variables` are hand-written so changing them is ok.  `Dict[str, str]` would be correct as documented https://www.sublimetext.com/docs/api_reference.html#sublime.expand_variables  (Although I don't know if `_Environ[str]` is then acceptable FTOMH.)

kaste:
And when you want the `project_path`, this information is in `window.extract_variables()` (https://www.sublimetext.com/docs/api_reference.html#sublime.Window.extract_variables).  A `ChainMap(window.extract_variables(), os.environ)` could work.

core/git_command.py
zebyja: I document feature and fix the None issue

zebyja:
* I documented the feature as @kaste suggested, please check my english there. Sometime people say that I speak as Yoda.
* None problem mentioned by @kaste fixed
* mypy fixed
* project variables also can be used to replace.

kaste (changes requested):
Almost.

core/git_command.py
```diff
@@ -279,9 +279,19 @@ def git(
...
 
         stdout, stderr = None, None
+        vars_for_replace = ChainMap(
+            custom_environ or {},
+            window.extract_variables(),
+            os.environ
+        )
+        savvy_env = self.savvy_settings.get("env") or {}
+        savvy_env_expanded = {
+            k: sublime.expand_variables(v, dict(vars_for_replace))
```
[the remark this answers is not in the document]
zebyja: Now it should be correct. Sorry, I did not catch it at first.

kaste:
Excellent.

kaste:
Thanks for doing this.  This will be included in the next release.  Should not take too long to get that ready.

PR #1745, opened 2023-02-12
Iterate graph drawing

kaste:
Fixes https://github.com/timbrel/GitSavvy/issues/1364
Closes https://github.com/timbrel/GitSavvy/issues/494

Generally shorten the graphs to some "n" commits after the commit-of-interest (the `follow` setting).  

Also shorten the ASCII art ("the graph") if it gets too wide.

### DIFF
--- a/core/commands/log_graph.py
+++ b/core/commands/log_graph.py
@@ -8,6 +8,7 @@ import shlex
 import subprocess
 import time
 import threading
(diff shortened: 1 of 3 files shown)


kaste:
This looks in part a bit WIP but is good enough for a test in the wild.

PR #1748, opened 2023-03-10
Avoid active reloading after upgrade

deathaxe:
Actively unloading and reloading GitSavvy after it has just been installed/updated may fail with strange exceptions if multiple packages are being updated at that time.

Replacing default builtin importer may also cause other plugins to fail loading.

Everything needed to ensure all sub-modules being reloaded in correct order is to remove them from `sys.modules` before any local import appears.

As a result saving `git_savvy.py` is everything needed for every sub- module to be re-imported during normal plugin reloading.

The `common.utils.reload` module becomes somewhat obsolete then.

It may still be used for development/debugging purposes, but well ... .

Note: The choosen mechanism has been used by A File Icon, GitGutter, PackageDev and probably more packages for quite a while without issues. The only crucial part is that there must be a single plugin in a package's root directory only, which imports modules from a sub directory.

### DIFF
--- a/git_savvy.py
+++ b/git_savvy.py
@@ -1,3 +1,12 @@
+import sys
+
+prefix = __package__ + '.'  # don't clear the base package
+for module_name in [
+        module_name for module_name in sys.modules
+        if module_name.startswith(prefix) and module_name != __name__]:
+    del sys.modules[module_name]
+prefix = None
+
 import sublime
 
 from .common.commands import *
@@ -12,30 +21,7 @@ from .gitlab.commands import *
 
 
 def plugin_loaded():
-
-    try:
-        import package_control.events
-    except ImportError:
-        pass
-    else:
-        if (
-            package_control.events.install('GitSavvy') or
-            package_control.events.post_upgrade('GitSavvy')
-        ):
-            # The "event" (flag) is set for 5 seconds. To not get into a
-            # reloader excess we wait for that time, so that the next time
-            # this exact `plugin_loaded` handler runs, the flag is already
-            # unset.
-            sublime.set_timeout_async(reload_plugin, 5000)
-            return
-
     prepare_gitsavvy()
-
-
-def reload_plugin():
-    from .common import util
-    print("GitSavvy: Reloading plugin after install.")
-    util.reload.reload_plugin(verbose=False, then=prepare_gitsavvy)
 
 
 def prepare_gitsavvy():
--- a/git_savvy.py
+++ b/git_savvy.py
@@ -1,23 +1,23 @@
 import sys
-
 prefix = __package__ + '.'  # don't clear the base package
 for module_name in [
         module_name for module_name in sys.modules
         if module_name.startswith(prefix) and module_name != __name__]:
     del sys.modules[module_name]
-prefix = None
+del prefix
 
-import sublime
 
-from .common.commands import *
-from .common.ui import *
-from .common.global_events import *
-from .core.commands import *
-from .core.settings import *
-from .core.interfaces import *
-from .core.runtime import *
-from .github.commands import *
-from .gitlab.commands import *
+import sublime  # noqa: E402
+
+from .common.commands import *  # noqa: E402
+from .common.ui import *  # noqa: E402
+from .common.global_events import *  # noqa: E402
+from .core.commands import *  # noqa: E402
+from .core.settings import *  # noqa: E402
+from .core.interfaces import *  # noqa: E402
+from .core.runtime import *  # noqa: E402
+from .github.commands import *  # noqa: E402
+from .gitlab.commands import *  # noqa: E402
 
 
 def plugin_loaded():


kaste:
Actually difficult to test this PR and find unwanted, subtle differences.   One difference to the *traditional* hot-reloading is that we loose all in-memory state.  E.g. 

https://github.com/timbrel/GitSavvy/blob/8624166a4bcd98666c1a001aeac668f26c64956b/common/util/debug.py#L14-L18

will not work anymore.  (T.i. we don't "hot" reload but always "cold" load.)  

- How do you use this during development?  Do you have a script that triggers a save on the main file?

deathaxe:
Saving main plugin is basically, what is enough to trigger reloading. Have been too lazy to write scripts for that, however. Never got into trouble with that strategy, so far. Probably because none of my plugins requried global states to persist. Found it even useful that all globals are reinitialized.

Using that strategy in A File Icon, GitGutter, MarkdownEditing, ... .

The only packages, which this strategy doesn't work well, are those which are accessed by 3rd-party packages. LSP and probably SublimeLinter are such candidates.

The biggest issue with GitSavvy's current strategy is the whole package being disabled and re-enabled for reloading asynchronously, which might cause trouble when multiple packages are installed/upgraded via package control. I even saw exceptions causing reloading to fail when running GitSavvy on python 3.8.

Package Control "solves" the whole reloading stuff by a `1_reloader.py` which reloads all existing modules at import time, using a fixed loading order list (which needst to be maintained manually with care 😞 ).

So if you are in doupt with just removing modules from cache, you could call reload() synchronously at import time as well.

I just got into trouble with that strategy as soon as submodule structures changed by an update. That's why I dislike that approach a bit.

kaste:
I tried (even with the help of ChatGPT) to get *hot*-reloading to work while keeping it as simple as this PR; with no luck.  Generally I don't know enough about the domain here, hey I did not the reloader in the first place, but I do not want to loose the feature lightly.  Do you have traces of when the reloader fails and throws.  Can we fix just that without throwing the baby...?

deathaxe:
An upgrade of GitSavvy fires the following output to console

```
Package Control: Searching updates...
ignored packages updated to: ["GitSavvy", "Vintage"]
unloading plugin GitSavvy.__init__
unloading plugin GitSavvy.git_savvy
reloading settings Packages/GitSavvy/GitSavvy.sublime-settings
reloading settings Packages/GitSavvy/syntax/blame.sublime-settings
reloading settings Packages/GitSavvy/syntax/branch.sublime-settings
reloading settings Packages/GitSavvy/syntax/diff.sublime-settings
reloading settings Packages/GitSavvy/syntax/diff_view.sublime-settings
reloading settings Packages/GitSavvy/syntax/graph.sublime-settings
reloading settings Packages/GitSavvy/syntax/make_commit.sublime-settings
reloading settings Packages/GitSavvy/syntax/output_panel.sublime-settings
reloading settings Packages/GitSavvy/syntax/rebase.sublime-settings
reloading settings Packages/GitSavvy/syntax/show_commit.sublime-settings
reloading settings Packages/GitSavvy/syntax/status.sublime-settings
reloading settings Packages/GitSavvy/syntax/tags.sublime-settings
reloading settings Packages/User/Package Control.sublime-settings
reloading settings Packages/User/Preferences.sublime-settings
generating meta info summary
generating syntax summary
Package Control: Upgraded GitSavvy to 2.40.0
reloading settings Packages/User/ApplySyntax.ext-list
reloading settings Packages/User/Bash.sublime-settings
ignored packages updated to: ["Vintage"]
reloading settings Packages/User/Package Control.sublime-settings
reloading settings Packages/User/Preferences.sublime-settings
reloading settings Packages/GitSavvy/GitSavvy.sublime-settings
reloading settings Packages/GitSavvy/syntax/blame.sublime-settings
reloading settings Packages/GitSavvy/syntax/branch.sublime-settings
reloading settings Packages/GitSavvy/syntax/diff.sublime-settings
reloading settings Packages/GitSavvy/syntax/diff_view.sublime-settings
reloading settings Packages/GitSavvy/syntax/graph.sublime-settings
reloading settings Packages/GitSavvy/syntax/make_commit.sublime-settings
reloading settings Packages/GitSavvy/syntax/output_panel.sublime-settings
reloading settings Packages/GitSavvy/syntax/rebase.sublime-settings
reloading settings Packages/GitSavvy/syntax/show_commit.sublime-settings
reloading settings Packages/GitSavvy/syntax/status.sublime-settings
reloading settings Packages/GitSavvy/syntax/tags.sublime-settings
reloading plugin GitSavvy.__init__
  File "<path_to_sublime_text>\Lib\python38\sublime_plugin.py", line 325, in reload_plugin
    m = importlib.import_module(modulename)
  File "./python3.8/importlib/__init__.py", line 127, in import_module
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 655, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 618, in _load_backward_compatible
  File "<frozen zipimport>", line 241, in load_module
  File "<frozen zipimport>", line 709, in _get_module_code
  File "<frozen zipimport>", line 548, in _get_data
zipimport.ZipImportError: bad local file header: '<path_to_sublime_text>\\Data\\Installed Packages\\GitSavvy.sublime-package'
reloading plugin GitSavvy.git_savvy
  File "<path_to_sublime_text>\Lib\python38\sublime_plugin.py", line 432, in load_module
    m.plugin_loaded()
  File "<path_to_sublime_text>\Data\Installed Packages\GitSavvy.sublime-package\git_savvy.py", line 32, in plugin_loaded
    prepare_gitsavvy()
  File "<path_to_sublime_text>\Data\Installed Packages\GitSavvy.sublime-package\git_savvy.py", line 44, in prepare_gitsavvy
AttributeError: module 'GitSavvy.core.runtime' has no attribute 'determine_thread_names'
generating syntax summary
generating meta info summary
```

Why the hell does everyone trap into "ChatGPT being the solution for everything"? You'll probably have better luck asking your kids.

kaste:
Hm, do you think the old `ZipImportError` comes from the reloader.  In SublimeLinter land we have reports of zip-import-errors before we had a reloader.

PR #1750, opened 2023-04-21
`"log_follow_rename": true` setting works properly for `show_file_at_commit`, `checkout_file_at_commit`

kylebebak:
- `"log_follow_rename": true` setting works properly for `show_file_at_commit`, `checkout_file_at_commit`
- This uses existing `filename_at_commit` function, [in the same way it's used here](https://github.com/kylebebak/GitSavvy/blob/3266859a62c12bf16da02332106e5da5ef0bfc52/core/commands/blame.py#L220-L224) in `blame_file_atcommit`
- I've tested this locally to verify both `show_file_at_commit` and `checkout_file_at_commit` work as expected, for files that have been renamed and files that haven't been renamed

### DIFF
--- a/core/commands/log.py
+++ b/core/commands/log.py
@@ -99,7 +99,7 @@ class LogMixin(GitCommand):
         if window:
             window.run_command("gs_show_commit_info", {
                 "commit_hash": commit,
-                "file_path": file_path
+                "file_path": self.file_path_at(file_path, commit)
             })
 
     def do_action(self, commit_hash, **kwargs):
@@ -289,5 +289,5 @@ class gs_log_action(PanelActionMixin, WindowCommand):
         })
 
     def checkout_file_at_commit(self):
-        self.checkout_ref(self._commit_hash, fpath=self._file_path)
+        self.checkout_ref(self._commit_hash, fpath=self.file_path_at(self._file_path, self._commit_hash))
         util.view.refresh_gitsavvy_interfaces(self.window, refresh_sidebar=True)
--- a/core/commands/show_file_at_commit.py
+++ b/core/commands/show_file_at_commit.py
@@ -146,9 +146,16 @@ class _gs_show_file_at_commit_refresh_mixin(GsTextCommand):
         views_with_reference_document.add(self.view)
 
     def previous_file_version(self, current_commit: str, file_path: str) -> str:
-        previous_commit = self.previous_commit(current_commit, file_path)
-        if previous_commit:
-            return self.get_file_content_at_commit(file_path, previous_commit)
+        # `file_path` is the canonical HEAD-side name. With `log_follow_rename`
+        # on, we must walk renames using the file's name *at* `current_commit`
+        # (because `--follow` needs an existing pathspec at the range tip), and
+        # then read the previous version using the name at the previous commit.
+        follow = bool(self.savvy_settings.get("log_follow_rename"))
+        name_at_current = self.file_path_at(file_path, current_commit)
+        previous_commit = self.previous_commit(current_commit, name_at_current, follow=follow)
+        if previous_commit and previous_commit != current_commit:
+            name_at_previous = self.file_path_at(file_path, previous_commit)
+            return self.get_file_content_at_commit(name_at_previous, previous_commit)
         else:
             # For initial revisions of a file, everything is new/added, and we
             # just compare with the empty "".
@@ -171,13 +178,14 @@ class gs_show_file_at_commit_refresh(_gs_show_file_at_commit_refresh_mixin):
         settings = view.settings()
         file_path = settings.get("git_savvy.file_path")
         commit_hash = settings.get("git_savvy.show_file_at_commit_view.commit")
+        name_at_commit = self.file_path_at(file_path, commit_hash)
 
         def program():
-            text = self.get_file_content_at_commit(file_path, commit_hash)
+            text = self.get_file_content_at_commit(name_at_commit, commit_hash)
             render(view, text, position)
             view.reset_reference_document()
             commit_details = self.commit_subject_and_date(commit_hash)
-            self.update_title(commit_details, file_path)
+            self.update_title(commit_details, name_at_commit)
             self.update_status_bar(commit_details)
             enqueue_on_worker(self.update_reference_document, commit_hash, file_path)
 
@@ -307,7 +315,8 @@ def get_next_commit(
     if next_commit := recall_next_commit_for(view, commit_hash):
         return next_commit
 
-    next_commits = cmd.next_commits(commit_hash, file_path)
+    follow = bool(cmd.savvy_settings.get("log_follow_rename"))
+    next_commits = cmd.next_commits(commit_hash, file_path, follow=follow)
     remember_next_commit_for(view, next_commits)
     return next_commits.get(commit_hash)
 
@@ -322,7 +331,11 @@ def get_previous_commit(
     if previous := recall_previous_commit_for(view, commit_hash):
         return previous
 
-    if previous := cmd.previous_commit(commit_hash, file_path):
+    follow = bool(cmd.savvy_settings.get("log_follow_rename"))
+    # `--follow` needs the file's name *at* `commit_hash` (the range tip);
+    # the HEAD-side name may not exist there yet.
+    name_at_commit = cmd.file_path_at(file_path, commit_hash)
+    if previous := cmd.previous_commit(commit_hash, name_at_commit, follow=follow):
         remember_next_commit_for(view, {previous: commit_hash})
     return previous
 
--- a/core/commands/show_file_at_commit.py
+++ b/core/commands/show_file_at_commit.py
@@ -150,12 +150,12 @@ class _gs_show_file_at_commit_refresh_mixin(GsTextCommand):
         # on, we must walk renames using the file's name *at* `current_commit`
         # (because `--follow` needs an existing pathspec at the range tip), and
         # then read the previous version using the name at the previous commit.
-        follow = bool(self.savvy_settings.get("log_follow_rename"))
+        follow = bool(file_path and self.savvy_settings.get("log_follow_rename"))
         name_at_current = self.file_path_at(file_path, current_commit)
         previous_commit = self.previous_commit(current_commit, name_at_current, follow=follow)
         if previous_commit and previous_commit != current_commit:
             name_at_previous = self.file_path_at(file_path, previous_commit)
-            return self.get_file_content_at_commit(name_at_previous, previous_commit)
+            return self.get_file_content_at_commit(name_at_previous or file_path, previous_commit)
         else:
             # For initial revisions of a file, everything is new/added, and we
             # just compare with the empty "".
@@ -181,11 +181,11 @@ class gs_show_file_at_commit_refresh(_gs_show_file_at_commit_refresh_mixin):
         name_at_commit = self.file_path_at(file_path, commit_hash)
 
         def program():
-            text = self.get_file_content_at_commit(name_at_commit, commit_hash)
+            text = self.get_file_content_at_commit(name_at_commit or file_path, commit_hash)
             render(view, text, position)
             view.reset_reference_document()
             commit_details = self.commit_subject_and_date(commit_hash)
-            self.update_title(commit_details, name_at_commit)
+            self.update_title(commit_details, name_at_commit or file_path)
             self.update_status_bar(commit_details)
             enqueue_on_worker(self.update_reference_document, commit_hash, file_path)
 
@@ -315,7 +315,9 @@ def get_next_commit(
     if next_commit := recall_next_commit_for(view, commit_hash):
         return next_commit
 
-    follow = bool(cmd.savvy_settings.get("log_follow_rename"))
+    # `--follow` requires a pathspec; in views without a file (e.g. show-commit)
+    # `file_path` is None, so don't request follow there.
+    follow = bool(file_path and cmd.savvy_settings.get("log_follow_rename"))
     next_commits = cmd.next_commits(commit_hash, file_path, follow=follow)
     remember_next_commit_for(view, next_commits)
     return next_commits.get(commit_hash)
@@ -331,13 +333,18 @@ def get_previous_commit(
     if previous := recall_previous_commit_for(view, commit_hash):
         return previous
 
-    follow = bool(cmd.savvy_settings.get("log_follow_rename"))
+    follow = bool(file_path and cmd.savvy_settings.get("log_follow_rename"))
     # `--follow` needs the file's name *at* `commit_hash` (the range tip);
     # the HEAD-side name may not exist there yet.
     name_at_commit = cmd.file_path_at(file_path, commit_hash)
-    if previous := cmd.previous_commit(commit_hash, name_at_commit, follow=follow):
+    previous = cmd.previous_commit(commit_hash, name_at_commit, follow=follow)
+    # `previous_commit` returns `commit_hash` itself when there's no actual
+    # older commit (e.g., at the file's initial revision). Don't cache that
+    # self-referential entry, because it would later make `n` a NOOP.
+    if previous and previous != commit_hash:
         remember_next_commit_for(view, {previous: commit_hash})
-    return previous
+        return previous
+    return None
 
 
 def remember_next_commit_for(view: sublime.View, mapping: Dict[str, str]) -> None:
--- a/core/git_mixins/history.py
+++ b/core/git_mixins/history.py
@@ -1,15 +1,15 @@
 from __future__ import annotations
+
 import email.utils
 from itertools import chain, takewhile
+from typing import Iterator, List, NamedTuple, Optional
 
-from ..exceptions import GitSavvyError
-from ...common import util
 from GitSavvy.core.fns import last, pairwise
 from GitSavvy.core.git_command import mixin_base
 from GitSavvy.core.utils import cached
 
-
-from typing import Iterator, List, NamedTuple, Optional
+from ...common import util
+from ..exceptions import GitSavvyError
 
 
 class LogEntry(NamedTuple):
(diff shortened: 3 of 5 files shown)


core/commands/log.py
```diff
@@ -250,5 +250,13 @@ def blame_file_atcommit(self):
         })
 
     def checkout_file_at_commit(self):
-        self.checkout_ref(self._commit_hash, fpath=self._file_path)
+        commit_hash = self._commit_hash
+        file_path = self._file_path
+
+        if commit_hash and file_path and self.savvy_settings.get("log_follow_rename"):
```
kaste, line 256: At this point all three predicates should be `True` all the time if I'm reading this correctly.  

There is also https://github.com/timbrel/GitSavvy/blob/8afe2de4e8c9cb5407bd63fb649c9b604cbd5faa/core/commands/log_graph.py#L2702-L2704 which would need the same patch I think.

core/commands/show_file_at_commit.py
```diff
@@ -92,7 +92,12 @@ def run(self, edit, position=None):
...
         commit_hash = settings.get("git_savvy.show_file_at_commit_view.commit")
 
-        text = self.get_file_content_at_commit(file_path, commit_hash)
+        if commit_hash and self.savvy_settings.get("blame_follow_rename"):
+            filename_at_commit = self.filename_at_commit(file_path, commit_hash)
+        else:
+            filename_at_commit = file_path
+
+        text = self.get_file_content_at_commit(filename_at_commit, commit_hash)
         render(view, text, position)
```
kaste, line 101: I don't think the changes for "show_file_at_commit" do work.  Not sure I do get them though.  What confuses me is that we get the `filename_at_commit` here but don't use it on L103 and L105 too.  Is that intentional?

Also the whole `show_file_at_commit` view has `[p]revious` and `[n]ext` handlers, starting at L201 below.  Do they just work or did you miss them?

kaste (commented):
Generally very welcome change.  I think the `show_file_at_commit` view is more complicated to do right though.  Maybe you should split this PR in two; esp. as the `checkout` part looks almost trivial.

core/commands/log.py
kylebebak: Do you mean the three expressions in the `if` statement, `if commit_hash and file_path and self.savvy_settings.get("log_follow_rename"):`?

Type checker says `file_path` is `Unknown | None` in line 254 above, so I figured, and `self.filename_at_commit` takes `(str, str)`, so I figured I'd just make sure both were truthy

As for `self.savvy_settings.get("log_follow_rename")`, that's a user setting that's not guaranteed to be truthy in general, right?

---

Agreed on making same patch in `log_graph`.py` module, I'll do that

core/commands/show_file_at_commit.py
kylebebak: I realize there's an error in the code I wrote. It should be checking `self.savvy_settings.get("log_follow_rename"):`, not `self.savvy_settings.get("blame_follow_rename"):`

The code works for me in both cases, but that's because I have both of these enabled in my GitSavvy settings

```json
	"log_follow_rename": true,
	"blame_follow_rename": true,
```

I just fixed this

---

> What confuses me is that we get the filename_at_commit here but don't use it on L103 and L105 too

Thanks for catching this, I just fixed this as well

> Also the whole show_file_at_commit view has [p]revious and [n]ext handlers, starting at L201 below. Do they just work or did you miss them?

I tried making this code work and couldn't

kylebebak:
@kaste 

I think the bug you found in the `show_file_at_commit` code I wrote above is fixed now. The `[n]` and `[p]` hotkeys to page through commits are not working

I think having `show_file_at_commit` working for renamed files, even without the n/p hotkey support, is valuable and could go in with this PR. If you think that's essential to get these hotkeys right along with enabling `show_file_at_commit` for `log_follow_rename`, I'll make another PR

core/commands/log.py
kaste: > As for self.savvy_settings.get("log_follow_rename"), that's a user setting that's not guaranteed to be truthy in general, right?

Ah right, looks like `log_follow_rename` can be `False` still, but `self._commit_hash` is always truthy here and `self._file_path` too otherwise the menu item would not appear (see `def update_actions(self)`).

kaste:
I actually wonder why the defaults for `log_follow_rename` and `blame_follow_rename` are ... wrong?  At least they seem wrong.  Like you I have set them to `true`.

kaste:
Actually `n` and `p` should work because it also informs us if the implementation is correct here or hacky.  At least we must know why it doesn't work.)  Generally, with `show_file_at_commit`, your approach is surprising in that you patch the "refresh" command/function.  If _show_file_at_commit_ were a function I would say, shouldn't we call it with the resolved `(commit, file_name)` pair.  Isn't the resolving in the "refresh" instead of the constructor function too late and surprising?  Basically we call `show_file_at_commit(commit_hash, file_path)` but then show `commit_hash, different_file_path`.  Does this make everything easier or more complex?

kylebebak:
Hey @kaste , I finally got around to improving things in this PR. I think it's ready for another review. The [n]ext and [p]revious behavior is working in every view/workflow in which I've tested them:

- Blame previous/next
- `git: log`, in `SHOW-COMMIT: ...` views
- `git: log current file`, in `FILE: ...` views (this now works across file renames)

I wrote part of the code with claude: here's an LLM-generated summary of the work

## Follow renames in `git: log current file` and related views

Adds a `log_follow_rename` setting (default `false`, opt-in) that lets log/diff/show-file-at-commit views trace a file across past renames instead of stopping at the rename boundary.

### What this means for users

When `log_follow_rename: true` in GitSavvy settings:

- `git: log current file`: the quick panel now lists commits from before the file was renamed, and the bottom commit-info panel shows the file's diff at each highlighted commit (previously empty for pre-rename commits)
- `SHOW-FILE-AT-COMMIT` view: `n` / `p` navigate across the rename boundary. The gutter diff (reference document) renders correctly at pre-rename commits instead of being empty
- "Checkout file at commit": checks out the file under its name at the chosen commit, not the current name
- Graph view: same fix flows through the "Checkout file at commit" / "Show file at commit" actions there too

Default behavior is unchanged when the setting is off

### Implementation

- New helper `HistoryMixin.file_path_at(file_path, commit_hash)` gates `filename_at_commit` behind `log_follow_rename` in one place, so the setting check isn't sprinkled across callers
- Canonical-name convention: `git_savvy.file_path` always holds the HEAD-side name; historical names are computed on demand. Fixed `BlameMixin.show_file_at_commit` which previously violated this by pre-resolving to the historical name when launching the `SHOW-FILE-AT-COMMIT` view
- Rename-aware navigation: `gs_show_file_at_commit_refresh.run`, `previous_file_version`, `get_previous_commit`, `get_next_commit`, and `LogMixin.on_highlight` resolve the right name at each commit involved (the current commit for `--follow` to find a valid pathspec, and the previous commit for reading the older blob)
- Performance: a naive implementation called `git log --follow {commit}.. -- file_path` per navigation event, which made every `n` / `p` press and every quick-panel highlight noticeably slow even on unrenamed files. Replaced with a single cached `file_rename_history(file_path)` walk per file (one `git log --follow --name-status -- file_path`), reducing subsequent `filename_at_commit` calls to in-memory dict lookups. First navigation per file pays one git call; everything after is free until LRU eviction
- Robustness fallbacks:
  - Merge commits with default `--diff-merges=off` emit no `--name-status` line, so they're absent from the parsed history. When a commit isn't found in the cache, `filename_at_commit` falls back to the original targeted query (also cached). This also covers commits reachable only via `next_commits`'s `branch_hint` walk that aren't in HEAD's history.
  - `--follow` requires a pathspec, so all sites that derive `follow` from a setting now also gate on `bool(file_path and ...)`. This protects the `SHOW-COMMIT` view's `n` / `p` navigation (which has no `file_path`) from a `fatal: --follow requires exactly one pathspec` regression, and applies the same defensive pattern to `gs_blame_current_file.log`.

### Caveats

- Cache is keyed on `file_path` and lives in the global LRU (size 512). If you rename a file, you need to commit this change before you can walk the file's history. Same staleness model as the existing `read_commit` / `get_file_content_at_commit` caches
- `file_rename_history` walks from HEAD, so commits reachable only via side branches fall through to the slower fallback path (correct, just one git call per such commit)

kaste:
Oh hi!

A lot of time passed since then and I need a primer on what we're trying to solve here.  Esp. I don't remember why we have the setting `blame_follow_rename` (and if we need it why it is `False` by default.  Same for `log_follow_rename`.  I have both set to `True`.

Then I don't know and remember what actually didn't work in the first place.  

It is okay to let AI write the code, do you still have the conversation protocol and prompts.  These are more indicative to the work as the result (sometimes).   E.g. the first commit message is just not readable.  

I need to think about if we can cache this here; i.e. are these caches safe.

kylebebak:
Hey @kaste , hello again =)

We're trying to solve the problem of looking at log/history of a given a file, whether through `blame` or `log` commands, and having continuity across commits that rename a file

Without these changes, this simply doesn't work. Try doing `git: log current file`, scrolling down through commits in the quick panel to a commit where the file had a different name. The output panel is empty (no diff with changes for this file in this commit). Then try doing `Show file at commit`. The output panel shows something like this

```
$ git show 0da010251:file.py

fatal: path 'file.py' exists on disk, but not in '0da010251'
```

Hitting `p` from a commit with the previous file name is a NOOP, while hitting `n` immediately jumps to the oldest commit in which the file has its "current" name

---

Just copying from an earlier comment: with the changes in this PR, the `[n]ext` and `[p]revious` behavior is working in every view/workflow in which I've tested them:

- Blame previous/next
- `git: log`, in `SHOW-COMMIT: ...` views
- `git: log current file`, in `FILE: ...` views (this now works across file renames)

And, e.g. when doing `git: log current file`, we can see the individual file diff for every commit in the quick panel, across all the names the file has ever had

---

As far as the caching goes, it's necessary to ensure decent perf. I've done a fair bit of testing and I haven't gotten any stale cache reads. I can, e.g., rename a file that's already been committed, commit the rename, rename it again, commit the second rename, and still navigate over all the commits that touched that file, in all its previous names. So yes, I think it's safe

I've attached the full prompt logs using `claude-history`, but I don't think they're very useful (too verbose). 

[conversation-2026-04-29-001009.md](https://github.com/user-attachments/files/27191207/conversation-2026-04-29-001009.md)

kaste:
Thanks for the claude history; I think it is readable and gives a lot of information about how that came to be.  

Also thanks for the primer of course.  

I maybe annotate some interesting claude mistakes.  Based on your findings, I implemented the same thing from a different angle.  The protocol for #2059 is here: https://pi.dev/session/#45ec33a8e0b95abd7da563beb3639203.

core/commands/show_file_at_commit.py
```diff
@@ -171,13 +178,14 @@ def run(self, edit: sublime.Edit, position: Position = None, sync: bool = True)
         settings = view.settings()
         file_path = settings.get("git_savvy.file_path")
         commit_hash = settings.get("git_savvy.show_file_at_commit_view.commit")
+        name_at_commit = self.file_path_at(file_path, commit_hash)
```
kaste, line 181: This is likely the culprit for the slowness; claude calls this on the main/UI thread.  It should have put this into the `program()` which runs on a worker thread.

core/git_mixins/history.py
```diff
@@ -194,22 +194,69 @@ def get_short_hash(self, commit_hash):
...
+        # be the HEAD-side name, since `--follow` matches the pathspec at
+        # the tip of the range. Cached: lets callers do many cheap lookups
+        # against a single subprocess walk.
+        output = self.git(
+            "log",
+            "--follow",
+            "--name-status",
+            "--format=COMMIT %H",
+            "--", file_path
+        )
```
kaste, line 211: This is unbounded and walks the complete history which is really not ideal.

core/git_mixins/history.py
```diff
@@ -194,22 +194,69 @@ def get_short_hash(self, commit_hash):
     def resolve_commitish(self, ref: str) -> str:
         return self.git("rev-parse", "--short", ref).strip()
 
+    @cached(not_if={"file_path": lambda p: not p})
```
kaste, line 197: `p: not p` is likely bogus/slop.  The functions before it should guarantee that we have a filename.

core/git_mixins/history.py
```diff
@@ -194,22 +194,69 @@ def get_short_hash(self, commit_hash):
...
+                if len(cols) >= 2:
+                    history.append((current_commit, cols[-1]))
+                    current_commit = None
+        return history
+
     def filename_at_commit(self, filename, commit_hash):
+        # type: (str, str) -> str
+        for h, name in self.file_rename_history(filename):
+            if h.startswith(commit_hash):
+                return name
```
kaste, line 228: This is just strange because as far as I understand this, the cache is filled *once* per *filename* and then as soon as HEAD moves the chances are getting higher and higher to get a cache miss.  

You also need to guarantee that filename is a full path filename or you get clashes when working with different projects at the same time.

core/git_mixins/history.py
```diff
@@ -194,22 +194,69 @@ def get_short_hash(self, commit_hash):
...
+        # query so `get_file_content_at_commit` doesn't blow up.
+        return self._filename_at_commit_uncached(filename, commit_hash)
+
+    @cached(not_if={"commit_hash": is_dynamic_ref})
+    def _filename_at_commit_uncached(self, filename, commit_hash):
         # type: (str, str) -> str
         lines = self.git(
             "log",
-            "--format=",  # we don't need any commit info beside the name status
+            "--format=",
```
kaste, line 241: Surprising slop for 2026

core/commands/show_file_at_commit.py
kylebebak: `file_path_at` runs `git log --follow ...` in a subprocess under the hood. I'm pretty sure the `git ...` subprocess is the slow part of the `file_path_at`. Whether this subprocess is launched from the main thread or a worker thread won't make any difference to latency. The caching just ensures this doesn't have to be rerun every time e.g. the user scrolls to a new commit in the quick panel. The latency in this branch is already quite good for my taste, even for large repos (I've tried it on repos with ~20k commits)

core/git_mixins/history.py
kylebebak: For a given file, if you want to show its complete history, e.g. show all the commits that touched the file in the quick panel, you're going to have to walk the whole history at least once. There's no way around that. The caching ensures that for a call to e.g. `git: log current file`, we only walk this history **once**, so I actually think it's pretty close to optimal, but maybe I'm misunderstanding something

core/git_mixins/history.py
kylebebak: Ok, let's remove this check if you don't think it's necessary. Or leave it there -- it doesn't cost anything

core/git_mixins/history.py
kylebebak: Surprising that it removed a comment? Should I add it back in?

core/git_mixins/history.py
kylebebak: When you call something like `git: log current file`, the caching that prevents repeated full walks is only done once per call. Using this command as an example, it calls `filename_at_commit`, which calls `file_rename_history`, **regardless of caching**. This calls `git log --follow ...` and walks the complete history. When scrolling through these commits, for any commit that we haven't seen yet we launch a subprocess with e.g. `['/usr/bin/git', 'show', '--no-color', '--format=fuller', '--stat', '--patch', '26beee75', '--', 'core/git_mixins/history.py']`, but we're not walking the full graph (we're looking up the file at a single commit)

As for the stale cache reads problem, thank you for calling out this possibility. I tested it as follows: I created two repos with an `index.html` file relative to the root of each, made a few commits to each repo updating `index.html`, then did `git: log current file` on both. I was unable to get any stale cache reads. I don't think they are possible across projects, even for files that have the same paths relative to the root of different repos

kylebebak:
Thank you for reviewing @kaste . [I saw that you opened another PR with the same functionality](https://github.com/timbrel/GitSavvy/pull/2059). The implementation in this PR has worked well for the last couple of weeks for a couple of repos with ~20k commits. Calling it battle-tested would be a stretch (it's not even unit-tested), but I think I've run into and fixed most of the edge cases I could hit.

I don't feel qualified to review yours. I don't even have a truly deep understanding of mine. I think it would be fine to clean this one up and merge it, e.g. remove the `blame_follow_rename` setting if you think it's redundant with `log_follow_rename`, and have `log_follow_rename` default to true. Or, just merge yours. Either way I think this functionality is useful, because renaming files is fairly common, and I've been enjoying it a lot since I added it to my fork

kaste:
Thanks for the PR which basically showed all parts that need a change which made my PR easier to get done.  I just merged #2059 and will do some follow up PRs to get a deeper feature set.  Then release.

kylebebak:
Awesome, thank you @kaste =)

kylebebak:
@kaste , I actually just tried installing `GitSavvy` from https://github.com/timbrel/GitSavvy instead of from my fork (https://github.com/kylebebak/GitSavvy), and bugs with the history view persist. E.g. when I go back through the history of a file that's been renamed several times, things like this happen

```
During handling of the above exception, another exception occurred:

...

  File "./python3.8/contextlib.py", line 75, in inner
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/runtime.py", line 133, in task
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/runtime.py", line 359, in task
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/commands/show_commit_info.py", line 101, in run_impl
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/commands/show_commit_info.py", line 136, in _read_commit_for_file
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/fns.py", line 47, in unique
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/commands/show_commit_info.py", line 124, in <genexpr>
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/commands/show_commit_info.py", line 132, in <lambda>
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/git_mixins/history.py", line 252, in filename_at_commit
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/git_mixins/history.py", line 75, in __getitem__
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/caches.py", line 25, in __getitem__
KeyError: ('6b331c00a', '/Users/kyle/Elementary/cloud-django/sdk/inspections/views/ingest.py')
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/git_mixins/history.py", line 249, in filename_at_commit
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/git_mixins/history.py", line 75, in __getitem__
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/caches.py", line 25, in __getitem__
KeyError: ('b2464b302', '/Users/kyle/Elementary/cloud-django/sdk/inspections/views/ingest.py')

During handling of the above exception, another exception occurred:

...
KeyError: ('b2464b302', '/Users/kyle/Elementary/cloud-django/sdk/inspections/views/ingest.py')
...
KeyError: ('16ac23f31', '/Users/kyle/Elementary/cloud-django/sdk/inspections/views/ingest.py')

During handling of the above exception, another exception occurred:

...
KeyError: ('16ac23f31', '/Users/kyle/Elementary/cloud-django/sdk/inspections/views/ingest.py')
...
KeyError: ('6c188d15a1408d77713b9a1607545850bb6c7b05', '/Users/kyle/Elementary/cloud-django/sdk/inspections/views/ingest.py')

During handling of the above exception, another exception occurred:

  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/ui__quick_panel.py", line 47, in _on_done
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/ui_mixins/quick_panel.py", line 72, in on_action_selection
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/commands/log.py", line 280, in show_file_at_commit
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/git_mixins/history.py", line 252, in filename_at_commit
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/git_mixins/history.py", line 75, in __getitem__
  File "/Users/kyle/Library/Application Support/Sublime Text/Installed Packages/GitSavvy.sublime-package/core/caches.py", line 25, in __getitem__
KeyError: ('6c188d15a1408d77713b9a1607545850bb6c7b05', '/Users/kyle/Elementary/cloud-django/sdk/inspections/views/ingest.py')
```

I don't know what the cause is, but I figure the simplest fix would be to revert your changeset and merge this one instead (I tested this one pretty heavily, on repos with tens of thousands of commits). In any case, thank you for your help

kaste:
I forgot to push the latest code.  Just released 2.59.0 now.

Restart after update.  🤞

PR #1753, opened 2023-05-11
Move cursor to the next hunk after staging (et.al.)

kaste:
Implements precise cursor movement after staging/unstaging etc. in the diff.  (Surprisingly, we did not had any code here but relied more or less on luck.)


kaste:
The actions did not trigger here.  Just try to open a new PR.

PR #1760, opened 2023-05-27
Scroll to nearest hunk when switching from file to diff

### LINKED ISSUE #1040: Add jump from file to hunk in diff view

kaste:
Fixes #1040

We do this for inline diff for ages now:  when opening (or switching to) a diff view from the source file, try to jump to /or show) the position we're currently at.  

E.g. total seamless:

https://github.com/timbrel/GitSavvy/assets/8558/de9c07ab-087f-4dd5-a589-e0d58868f774

Without this patch the diff would land on the first hunk of the edited file.

### DIFF
--- a/core/commands/blame.py
+++ b/core/commands/blame.py
@@ -6,7 +6,7 @@ import unicodedata
 import sublime
 
 from .navigate import GsNavigate
-from ..view import Position
+from ..view import scroll_to_pt, y_offset, Position
 from ...common import util
 from .log import LogMixin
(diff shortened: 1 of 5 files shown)


kaste:
There are a couple of refactorings, but the last commit has the feature and could be undone easily enough.

PR #1797, opened 2023-10-08
fix(branch): pass name as initial prompt text

asfaltboy:
Unlike with remote branches, local branches that were the target of the [b] shortcut, were failing to pass the branch name as the initial text of the prompt for the new branch.

For context, this is a problem for me as I use this flow:
1. when I have a big change planned
2. I create a copy of the local branch by hitting [b]
3. I usually add a suffix in the end of the branch to indicate it's a copy

Tested on a local copy

### DIFF
--- a/core/commands/checkout.py
+++ b/core/commands/checkout.py
@@ -75,7 +75,9 @@ class gs_checkout_new_branch(GsWindowCommand):
     Prompt the user for a new branch name, create it, and check it out.
     """
     defaults = {
-        "branch_name": ask_for_name(),
+        "branch_name": ask_for_name(
+            initial_text=lambda start_point: start_point,
+        ),
     }
 
     def run(self, branch_name, start_point=None, force=False, merge=False):


kaste:
I had to revert and make a hotfix release 2.42.1 because it threw on the standard `git: checkout new branch` command.  

```
  File "C:\Users\c-flo\AppData\Roaming\Sublime Text\Packages\GitSavvy\core\base_commands.py", line 75, in run_
    self.defaults[name](self, args, done)
  File "C:\Users\c-flo\AppData\Roaming\Sublime Text\Packages\GitSavvy\core\commands\branch.py", line 71, in handler
    initial_text_ or call_with_wanted_args(initial_text, args),
  File "C:\Users\c-flo\AppData\Roaming\Sublime Text\Packages\GitSavvy\core\commands\branch.py", line 48, in call_with_wanted_args
    return fn(**{k: args[k] for k in fs.parameters.keys() if k in args})
TypeError: <lambda>() missing 1 required positional argument: 'start_point'
```

See: #1800

PR #1855, opened 2024-01-17
feature: inline diff select commit/branch to diff

ta946:
Inline diff currently compares the file's current state vs the last commit, then you can cycle through previous states using n/p.
My usecase is usually current state vs last commit or completely different branch/commit so I've come up with a workaround.

image

Manual commit id entry
image

Branches
fetches the branch list then it will compare against the lastest commit in that branch
image

Commits
fetches the last N commits (default is 20) along with the commit message
image

Theres probably a better way of integrating the idea with the graph view, but this was good enough for me to whip up quickly. Thought id share incase it can be adapted and be a useful idea/addition

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -743,6 +743,15 @@
             { "key": "selector", "operand": "git-savvy.make-commit meta.dropped.git.commit" }
         ]
     },
+    {
+        "keys": ["s"],
+        "command": "gs_commit_view_open_transient_view",
+        "context": [
+            { "key": "setting.command_mode", "operator": "equal", "operand": false },
+            { "key": "setting.git_savvy.commit_view", "operator": "equal", "operand": true },
+            { "key": "selector", "operand": "git-savvy.make-commit meta.dropped.git.commit" }
+        ]
+    },
 
 
     ///////////////
--- a/core/commands/commit.py
+++ b/core/commands/commit.py
@@ -24,6 +24,7 @@ __all__ = (
     "gs_commit_view_do_commit",
     "gs_commit_view_sign",
     "gs_commit_view_close",
+    "gs_commit_view_open_transient_view",
     "gs_commit_log_helper",
     "GsPrepareCommitFocusEventListener",
     "GsPedanticEnforceEventListener",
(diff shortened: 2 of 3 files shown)


kaste:
Not sure what to do with this one.  Did you even pushed the correct version?  I don't see how having a normal `WindowCommand` in `class SelectInlineDiffCommitCommand(WindowCommand):` can call `git` e.g. `commit_text_list = self.git('log',...`.   The code is not very aligned with what we have, the checks are also red here.  

I do like the idea to some extend.  T.i. I can imagine being in an inline-diff view to hit maybe `[t]` which will open a quick panel to set a different `target`.  Or is it `base`, I never remember these correctly.  

You're correct that there aren't any obvious other ways to open a "historical" inline-diff.  I think you can add normal diffs between branches and/or commits, e.g. from the graph or branches views.  And then when you're seeing such a diff you can open the inline diff which will copy the target and base correctly.  That's a rather indirect feature.  

We could have it in the menu of a File History.  Inline diffs show only one file at a time, that's why we don't have links for them in say the branches view or the standard graph view. It just doesn't fit as long as you're not in a context of a file.  From that perspective one could add a command to the file revision views to then compare that file revision with the current HEAD/working dir state.

ta946:
you're right, i forgot to push the commit which allows calling git, I'm using different code on my machine that doesn't modify gitsavvy so it can update without me losing this functionality. pushed now. 

Tbh, this was a low quality PR that doesn't make proper use of gitsavvy's codebase and features, i haven't delved into it enough to contribute correctly. Just wanted to share the idea and if it was something you guys liked, then I could try spending time on fixing it, with some direction

yea wrt using the graph, as there is no context of a file I couldn't think of an easy way to get this functionality without having to go through a few steps that might seem abit convoluted. Maybe a command that when selecting to compare with another commit, a quick panel can open to select a file?

As for while in File History, I assume it only follows the current branch? That covers my use case 90% of the time, but sometimes I want to compare against a different branch that is newer

So yea, please let me know if this idea seems useful, maybe abit more brainstorming is needed for workflow, and I can give it more attention

kaste:
I don't forget you, I'm just a bit busy, okay?

ta946:
no problem at all, there's no rush

PR #1936, opened 2024-06-17
Tune guessing the next commit

kaste:
Given a random commit we always looked from "HEAD" up to that commit to find (or rather guess) the next commit.  This fails if the user is looking at a commit not contained in the current line.

Tune this and check first for a branch that contains the commit.  If there are many, take the one that has the most recent changes.  Then graph the line from the commit to the tip of the branch.

Since guessing a next commit is not cacheable, this new approach would require two git calls per user action.  To avoid that take the whole log line and store it on the view so that subsequent `[n]ext`-actions are free.  T.i. cache for the lifetime of the view.


kaste:
Hopefully, that gets a follow-up as there is some potential in it.

PR #1943, opened 2024-08-16
Exclude development files from sublime-package

deathaxe:
This PR adds various development related resources to .gitattributes in order to exclude them from GitSavvy.sublime-package installed via Package Control.

Notably `__init__.py` in root directory is excluded to avoid exceptions, when reloading package after upgrade without impact on current development or test workflows. The file is considered a dev-only resource.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,16 +2,16 @@
 # archive files. See http://git-scm.com/docs/gitattributes for details.
 
 # git
-.github/        	export-ignore
-*.git           	export-ignore
-*.gitignore     	export-ignore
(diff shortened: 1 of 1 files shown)


kaste:
Had the same idea wrt to "hiding" *init.py*; never was brave enough to actually do it.

kaste (changes requested):
No mixed tabs and spaces, I would say.  

And leave out "/scripts"; these are unused and in fact I was unaware of them.  I rather delete that directory.  

Other than that LGTM

kaste:
Let's try this.

PR #1944, opened 2024-08-29
feat: adds branch command to unset upstream

cdloh:


kaste:
Ah, that's useful.  

We actually have a command in the Command Palette to unset the tracking information for the checked out branch, that is `gs_unset_tracking_information`  https://github.com/timbrel/GitSavvy/blob/5aa2dfa11713421e2a08706841097b56253e3f4d/core/commands/branch.py#L145-L154

Let's just delegate to that command.  We do this quite often in this interface. Something among `self.window.run_command...`.  Also probably check if the selected branch is a remote branch, e.g. https://github.com/timbrel/GitSavvy/blob/ca8605ad651389a77f3991f9267be5465e9fff5b/core/interfaces/branch.py#L481-L485

and if it actually has upstream/tracking set.

cdloh:
Thanks @kaste. I think I've implemented all of those changes. 

I couldn't find any docs or relevant tests but happy to if I've missed something.

kaste:
Excellent.     

I usually don't have tests for these ... basically simple wrappers.  Most logic is in the parent classes e.g. `CommandForSingleBranch`.  

Actually, happy that this is such a simple diff/patch as well.

(I still need to fix these mypy errors; the latest mypy version got smarter and I did not pin the version here. 😬)

Let's roll.

kaste:
🥁

cdloh:
many thanks for maintaining the plugin @kaste! Not sure how much time I save with it weekly. All muscle memory now.

How long does package control take to pickup the new version?

kaste:
It's a crawler... checks every 3 hours

PR #1945, opened 2024-08-30
Add "github: create repo"

kaste:
Note that this initial version only supports public, user repos.  That is the minimum valuable enhancement.

### DIFF
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -137,6 +137,10 @@
         "command": "gs_github_add_fork_as_remote"
     },
     {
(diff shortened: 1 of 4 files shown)


deathaxe:
Wouldn't it be easier to provide a front-end for gh?

kaste:
Not sure tbh.  Typically the ux/ui thing is the difficult part.  The lowest level API call in this case is easy for example but to support all the possible flags and knobs of how to create a repo is difficult to *design*.  We probably need a new special view/dashboard just for github.

PR #1982, opened 2025-03-27
Teach diff view to handle untracked files

kaste:
Just having "UNTRACKED" in the header is lame, temporarily run
`intent_to_add` on that file, and read the diff again. That is mildly easy.

Staging (and undoing that) actually is more code.  Ensure that the status 
view does not intent-to-add untracked files too (which was esp. annoying.)


kaste:
It would be worth it now to add proper types around undo/the history.

PR #1983, opened 2025-03-27
Redesign the files panel in the diff view

kaste:
The goal is esp. to optimize `[N]` and `[P]` after staging:

When a user staged a file, GitSavvy will switch to "staged" mode as soon
as the diff gets clean.  This is implemented, a) to actually see what's
gonna be committed and b) e.g. to issue "Line History" directly to
select a commit to fixup.

Unfortunately this broke `gs_diff_switch_files` intuitive UX.  As we
switched to the staged mode, a subsequent `N` could not select a useful
*next* file to continue staging.

This commit fixes that.  The intended UX is, e.g.
    `[S]` to stage a file, followed by
    `[N]` to switch to the next _unstaged_ file.

### DIFF
--- a/core/commands/diff.py
+++ b/core/commands/diff.py
@@ -548,13 +548,15 @@ class gs_diff_switch_files(TextCommand, GitCommand):
         # type: (sublime.Edit, bool, bool, Optional[bool]) -> None
         view = self.view
         window = view.window()
-        if not window:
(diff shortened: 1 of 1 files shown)


kaste:
Tbh, this seems convoluted but maybe I was just clear minded and it works.

PR #1985, opened 2025-04-24
Add setting `fetch_new_remotes: true`

kaste:
When adding remotes, we always asked if we should now also fetch this remote.  Add "fetch_new_remotes" and just fetch without asking if set to `True`.

This is also the default because frankly you likely want that.  (Except for huge repos, but then set this in the project settings, no?)

### DIFF
--- a/GitSavvy.sublime-settings
+++ b/GitSavvy.sublime-settings
@@ -291,6 +291,11 @@
     "confirm_force_push": true,
 
     /*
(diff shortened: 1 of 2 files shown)


kaste:
Ideally, we had an UI where we could select which refs we're interested in.  E.g. a preview of what the remote provides -- esp. highlighting the stuff we don't have already.  Also very likely we don't need the tags, as tags tend to be shared ("centralized") even in a distributed git system.

PR #2058, opened 2026-04-12
Resolve SSH host aliases for GitHub remotes

### LINKED ISSUE #2057: Support ssh aliases when resolving repo's URL
It's possible to set up an alias for a host in `/~/.ssh/config` and then use that aliased hostname in git ssh URL.

For example in `~/.ssh/config`:

```
Host my-github
    HostName github.com
    User git
    AddKeysToAgent yes
    UseKeychain yes
    IdentityFile ~/.ssh/rchl-id_ed25519
```

and then define origin in `./repository/.git/config` like:

```
[remote "origin"]
	url = git@my-github:me/my-project.git
	fetch = +refs/heads/*:refs/remotes/origin/*
```

(The reason for doing it that way is because I want to use non-default ssh key for that repo.)

Git is able to resolve the hostname to github.com in this case and everything work correctly.

This also works with Github's GH CLI but with GitSavvy all commands use the alias which doesn't work (for example when opening repo's URL in a browser).

kaste:
Fixes #2057

Teach GitHub remote parsing to resolve SSH aliases from user ssh
config before constructing HTTPS URLs.

For SSH remotes, we now run `ssh -G <host>` and read the final
`hostname` value. This allows remotes like
`git@my-github:owner/repo.git` to open as
`https://github.com/owner/repo` when `my-github` is an alias.

### DIFF
--- a/github/github.py
+++ b/github/github.py
@@ -5,13 +5,15 @@ GitHub methods that are functionally separate from anything Sublime-related.
 from __future__ import annotations
 import os
 import re
-from webbrowser import open as open_in_browser
-from functools import partial
+import subprocess
+from functools import lru_cache, partial
 from typing import NamedTuple
+from webbrowser import open as open_in_browser
 
 from ..common import interwebs
 from ..core.exceptions import FailedGithubRequest
 from ..core.settings import GitSavvySettings
+from ..core.utils import STARTUPINFO
 
 
 GITHUB_PER_PAGE_MAX = 100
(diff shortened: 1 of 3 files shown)


michaelblyons:
Oh neat, you mocked the responses. I had to test the same thing for GitLink and made a dummy SSH config file to run real `ssh -G` lookups on.

https://github.com/rscherf/GitLink/pull/61/changes

kaste:
Yes, and I'm also surprised how good they actually read when using `mockito`.

